PR #24, opened 2022-07-20
add unit test case: test lower_bounds/upper_bounds of pos-delete file

wangtaohz:


core/src/test/java/com/netease/arctic/io/TaskWriterTest.java
```diff
@@ -62,6 +75,39 @@ public void testBasePosDeleteWriter() throws IOException {
     Assert.assertEquals(1, result.size());
   }
 
+  @Test
+  public void testBasePosDeleteWriterMetrics() throws IOException {
```
zhoujinsong, line 79: It seems new test case cover old `testBasePosDeleteWriter` one, so we can replace it ?

zhoujinsong: and change this test case name?

wangtaohz: `testBasePosDeleteWriterMetrics` is removed, and `testBasePosDeleteWriter` now check the metric

zhoujinsong (approved):

PR #30, opened 2022-07-21
position-delete optimize

shidayang:
reuse position delete for node


trino/src/test/java/com/netease/arctic/trino/iceberg/ArcticQueryRunnerForClient.java
```diff
@@ -61,7 +61,8 @@ public static DistributedQueryRunner createIcebergQueryRunner(
 
         queryRunner.installPlugin(new ArcticPlugin());
         Map<String, String> icebergProperties = ImmutableMap.<String, String>builder()
-            .put("arctic.url", "thrift://localhost:8080/" + ARCTIC_CATALOG)
+            // .put("arctic.url", "thrift://localhost:8080/" + ARCTIC_CATALOG)
+            .put("arctic.url", "thrift://10.196.98.23:18111/trino_online_env")
```
hameizi, line 65: private ip

shidayang: 。。。。

core/src/main/java/com/netease/arctic/io/reader/ArcticDeleteFilter.java
```diff
@@ -58,11 +64,22 @@
 
 /**
  * Abstract implementation of filtering equality delete files with insert files and base files.
```
zhoujinsong, line 66: It seems the class description should be updated

shidayang: ok

core/src/main/java/com/netease/arctic/io/reader/BaseIcebergDataReader.java
```diff
@@ -107,6 +107,19 @@ public CloseableIterable<T> readData(ArcticFileScanTask task) {
     return iterable;
   }
 
+  public CloseableIterable<T> readDataWithoutApplyPosDelete(ArcticFileScanTask task) {
```
zhoujinsong, line 110: the new method makes this class very strange. functions use the old `readData` method still have bad performace like flink reader?

shidayang: The readData method is used by original iceberg read

zhoujinsong (approved):

PR #36, opened 2022-07-21
clear optimize task history when all tasks are removed

wangtaohz:


hzluting (approved):

zhoujinsong:
when should we clear optimize task history?

wangtaohz:
> 
two cases:
case1: after table was dropped
case2: after tasks (which were not committed) were cleared (now tasks will be cleared only when table's 'ResourceGroup' is changed)

this PR fix case2

PR #62, opened 2022-07-25
[WIP] arctic table used as temporal table join.

YesOrNo828:
It is a demo showing how to lookup join with the arctic table by temporal table join.
`SELECT [column_list]
FROM table1 [AS <alias1>]
[LEFT] JOIN table2 FOR SYSTEM_TIME AS OF table1.{ proctime | rowtime } [AS <alias2>]
ON table1.column-name1 = table2.column-name1`

Two pain spots：
1. Left table joined with the arctic table would be missing join when initializing the arctic table.

2. How to emit immediately to the next operator, when the right table(arctic) can't push forward its' watermark.


zhoujinsong:
@YesOrNo828  Dose this pr is usefual now ? If not, plese close it.

YesOrNo828:
I close it because this patch is finished. https://github.com/NetEase/arctic/pull/127

PR #64, opened 2022-07-25
remove uselss code

wangtaohz:


hzluting (approved):

ams/ams-server/pom.xml
```diff
@@ -201,13 +205,6 @@
             <version>2.9.0</version>
         </dependency>
 
-        <!--   TODO wangtao3 temp, remove when api change thrift version to 0.13.0 -->
-        <dependency>
```
zhoujinsong, line 205: dose `mvn thrift:compile` depend on this ?

wangtaohz: No.

AFAIK, we use `mvn thrift:compile` to regenerate java code for thrift api in module `ams-api`, but in this PR, only the module `ams-server` is affected.

zhoujinsong: got

zhoujinsong (approved):

PR #68, opened 2022-07-25
remove dependency arctic-flink-1.12 from module optimizer

wangtaohz:


wangtaohz:
Arctic-flink is going to support flink 1.12 and 1.14, and set the version of flink by `mvn profile`, this brings error if optimizer have the dependency of arctic-flink-1.12. 

And also,  optimizer doesnt' need to have the dependency of arctic-flink.

zhoujinsong (approved):

PR #92, opened 2022-08-01
[prepare for conference]File cache check when optimize

hzluting:
If file cache isn't latest snapshot of base table when minor optimize, will cause inaccurate data. 
Optimize will use old data files rather than correct data files in base table.


zhoujinsong (changes requested):
Check the base store's current snapshot is the snapshot witch optimize plan for.

wangtaohz (approved):

zhoujinsong (approved):

PR #97, opened 2022-08-01
Fix file cache service

hameizi:
1.change `replace into` to `insert into`
2.add `primary_key_md5` as primary key
3.table `file_cache_info` add `parent_snapshot_id`
4.move `getCurrentSnapInfo` from `snapshot_info_cache` to `file_cache_info`  avoid there is one snapshot info in `snapshot_info_cache` but not  `file_cache_info`
5.add logic `lastSnapInfo` to get current snap when there is many snap are commit in one TX use same commit time


ams/ams-server/src/main/java/com/netease/arctic/ams/server/config/ArcticMetaStoreConf.java
```diff
@@ -138,7 +138,7 @@ public class ArcticMetaStoreConf {
   public static final ConfigOption<Long> TABLE_FILE_INFO_CACHE_INTERVAL =
       ConfigOptions.key("table.file.info.cache.interval")
           .longType()
-          .defaultValue(6 * 60 * 60 * 1000L)
+          .defaultValue(5 * 60 * 1000L)
```
zhoujinsong, line 141: why do us change this default value from 6 hours to 5 minutes?

hameizi: For a more timely update to file info. Because minor optimize is frequent.

zhoujinsong (approved):

PR #146, opened 2022-08-09
feactor BaseKeyedTable, move operations to keyedTable

baiyangtx:


baiyangtx:
seems not a good refactor

PR #153, opened 2022-08-10
[Spark] Support get table from catalog

zhaomin1423:
Support get table from catalog.


baiyangtx (commented):
ArcticSource has implement SupportCatalogOptions interface.  Spark use extractIdentifier method to extract table identifier and load table from CatalogManager. so this pr is not necessary

PR #170, opened 2022-08-15
[bugfix] Using Math.abs() make the hashcode positive.

YesOrNo828:
Fixed the log/file shuffle OutOfIndexExp which is caused by the hashcode of the record is not positive.

**the main change log**:
**ShuffleHelper.hashPartitionValue(RowData)** and **ShuffleHelper.hashPartitionValue(RowData)**


zhoujinsong:
I suggest add the abs logic to `com.netease.arctic.data.PrimaryKeyData.hashCode`.
`Math.abs(Integer.MIN_AVLUE)` is illegal, need specially processing.

zhoujinsong (approved):

PR #178, opened 2022-08-16
for #175 modify Optimize commit thread pool

wangtaohz:


core/src/main/java/com/netease/arctic/table/TableProperties.java
```diff
@@ -41,7 +41,9 @@ private TableProperties() {
   public static final String ENABLE_OPTIMIZE = "optimize.enable";
   public static final String ENABLE_OPTIMIZE_DEFAULT = "true";
 
+  @Deprecated
```
zhoujinsong, line 44: remove it from site too

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
```diff
@@ -71,9 +73,10 @@ public class OptimizeService extends IJDBCService implements IOptimizeService {
   private static final Logger LOG = LoggerFactory.getLogger(OptimizeService.class);
   private static final long DEFAULT_CHECK_INTERVAL = 60000;
 
-  private ScheduledTasks<TableIdentifier, OptimizeCommitTask> commitTasks;
   private ScheduledTasks<TableIdentifier, OptimizeCheckTask> checkTasks;
 
+  private final BlockingQueue<TableIdentifier> toCommitTables = new ArrayBlockingQueue<>(1000);
```
zhoujinsong, line 78: what if blocking queue is full?

wangtaohz: If the queue is full, when try putting TableIdentifier into queue, it will fail and return false immediately, and it will periodically retry putting this TableIdentifier later.

core/src/main/java/com/netease/arctic/table/TableProperties.java
wangtaohz: Removed.

zhoujinsong (approved):

PR #193, opened 2022-08-18
Check input is validate at OverwriteFiles  && ReplacePartitions

baiyangtx:
check input is validate at OverwriteFiles  && ReplacePartitions


hive/src/main/java/com/netease/arctic/hive/exceptions/CannotAlterHiveLocationException.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netease.arctic.hive.exceptions;
+
+/**
+ * Exception will be thrown when alter hive location check failed
+ */
+public class CannotAlterHiveLocationException extends IllegalStateException {
```
zhoujinsong, line 24: Does IllegalStateExcpetion with message is enough?

baiyangtx: may conflict with hadoop client.

zhoujinsong (approved):

PR #223, opened 2022-08-24
fix duplicate commit properties logic

hameizi:
There commit properties multipal in old logic, because the keyedTable and unkeyedTable both commit properties to ams. But keyedTable's base and change table are also unkeyedTable, so there may commit properties multipal.

### DIFF
--- a/core/src/main/java/com/netease/arctic/trace/AmsTableTracer.java
+++ b/core/src/main/java/com/netease/arctic/trace/AmsTableTracer.java
@@ -171,7 +171,7 @@ public class AmsTableTracer implements TableTracer {
       commitMeta.setSchemaUpdateMeta(ddlCommitMeta);
       update = true;
     }
-    if (this.properties != null) {
+    if (this.properties != null && Constants.INNER_TABLE_BASE.equals(innerTable)) {
       commitMeta.setProperties(this.properties);
       update = true;
     }
--- a/core/src/main/java/com/netease/arctic/table/BaseKeyedTable.java
+++ b/core/src/main/java/com/netease/arctic/table/BaseKeyedTable.java
@@ -160,13 +160,7 @@ public class BaseKeyedTable implements KeyedTable {
 
   @Override
   public UpdateProperties updateProperties() {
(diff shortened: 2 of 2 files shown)


core/src/main/java/com/netease/arctic/trace/AmsTableTracer.java
```diff
@@ -171,7 +171,7 @@ public void commit() {
       commitMeta.setSchemaUpdateMeta(ddlCommitMeta);
       update = true;
     }
-    if (this.properties != null) {
+    if (this.properties != null && Constants.INNER_TABLE_BASE.equals(innerTable)) {
```
SteNicholas, line 174: Could you please combine the `Constants.INNER_TABLE_BASE.equals(innerTable)` into one condition with 163 line?
```
if (Constants.INNER_TABLE_BASE.equals(innerTable)) {
  update = true;
}
```

SteNicholas (commented):
@hameizi, thanks for the contribution. I left the minor comments for this pull request. PTAL.

core/src/main/java/com/netease/arctic/trace/AmsTableTracer.java
hameizi: When there is datafile commited both baseTable and changeTable should commit tableChange, so the we can't set `update = true`  when Constants.INNER_TABLE_BASE.equals(innerTable)

SteNicholas: What is the condition that there is datafile commited both baseTable and changeTable should commit tableChange? IMO, whether the properties are null couldn't prove that there is datafile commited both baseTable and changeTable.

hameizi: yes you are right, the condition of tableChange is  datafile commited both baseTable and changeTable should commit tableChange.  the condition  properties are null is just for table just alter properties but not commit datafile. AmsTableTracer can commit many kinds of tracer data like alter properties, alter schema, commit data.

SteNicholas (approved):
LGTM.

zhoujinsong (approved):

Issue #173, opened 2022-08-16
[spark] Support Batch Insert/Delete/Update/MergeInto


baiyangtx:
Support Insert/Delete/Update/MergeInto SQL in Spark 

TaskList:

* Design for Batch  Insert/Delete/Update/MergeInto

Implement insert command at  top priority.
* Support insert for iceberg base store keyed table 
* Support insert for hive base store keyed table 
* Support insert for hive base store unkeyed table (high priority)

* Support delete for iceberg base store keyed table
* Support delete for hive base store keyed table 
* Support delete for hive base store unkeyed table

* Support update for iceberg base store keyed table
* Support update for hive base store keyed table 
* Support update for hive base store unkeyed table

* Support merge-into for iceberg base store keyed table 
* Support merge-into for hive base store keyed table
* Support merge-into for hive base store unkeyed table

Finally, rewrite code of **iceberg base store unkeyed table**  to keep logical consistency of all table types.
* Rewrite code for iceberg base store unkeyed table.

baiyangtx:
merge into is not supportted in  v0.3.2

PR #230, opened 2022-08-25
[WIP][ARCTIC #173] Feature spark update code framework

baiyangtx:
code framework of spark3.1 update/delete/upsert


spark/v3.1/spark/src/main/scala/org/apache/spark/sql/catalyst/parser/ArcticExtendSparkSqlAstBuilder.scala
```diff
@@ -125,7 +125,6 @@ class ArcticExtendSparkSqlAstBuilder(delegate: ParserInterface)
       case Some(query) =>
         val propertiesMap = buildProperties(primaryForCtas, properties)
         var writeOptions: Map[String, String] = Map.empty
-        writeOptions += ("overwrite-mode" -> "dynamic")
```
hellojinsilei, line 128: this option is used in data-frame api(use data-frame api to create keyed table), will it be affected if you delete it ?

hellojinsilei (approved):

hellojinsilei (approved):

zhoujinsong (approved):

PR #236, opened 2022-08-29
[ARCTIC-228] add Optimize integration test and fix some bugs

wangtaohz:

Continues Optimizing is one of the key feature of arctic, the changes of serval module (include ams, optimizer, core...) may affect it's stability, and unit test are hard to find these implicit effects. So it is necessary to add integration tests for continues optimizing to reduce it's fail risk when changes happens in these moudules.

## Brief change log

  - Add OptimizeIntegrationTest for continues optimizing of keyed/unkeyed, partitioned/unpartitioned table
  - Before integration test start, load an AMS on derby and a Local Optimizer.
  - After integration test end, stop and release AMS and Optimizer.
  - Fix some bugs the integration test find: 
    - fix optimize_task_history lost primary key, quota occupation is wrong
    - fix minor optimize commit an empty snapshot
    - fix derby not support null default value when insert into optimize_history for unkeyed table
    - fix first snapshot of table can't be write into file info cache
    - fix derby not support insert ignore when insert optimize_task_history
    - fix optimize not commit immediately when all tasks are prepared after added
    - fix optimize worker pool wait too long

## How was this patch tested?

## Documentation


ams/ams-server/src/test/resources/config/conf/config.yaml
```diff
@@ -0,0 +1,86 @@
...
+      storage.type: hdfs
+      core-site:
+      hdfs-site:
+      hive-site:
+    # auth config.sh now support SIMPLE and KERBEROS
+    auth_config:
+      type: SIMPLE
+      hadoop_username: root
+    properties:
+      warehouse.dir: /tmp/arctic_integration/warehouse
```
zhoujinsong, line 41: Does this test case depend on a cerntain folder?
I recommand create table on a temp folder generated by junit TemporaryFolder, so table folder would be removed automatically.

core/src/main/java/com/netease/arctic/op/OverwriteBaseFiles.java
```diff
@@ -109,29 +109,31 @@ protected StructLikeMap<Long> apply(Transaction transaction, StructLikeMap<Long>
...
+      OverwriteFiles overwriteFiles = transaction.newOverwrite();
+      if (baseTable.currentSnapshot() != null) {
+        overwriteFiles.validateFromSnapshot(baseTable.currentSnapshot().snapshotId());
+      }
+      for (DataFile d : this.addFiles) {
+        overwriteFiles.addFile(d);
+        partitionData = keyedTable.spec().isUnpartitioned() ? null : d.partition();
+        partitionMaxTxId.put(partitionData, getPartitionMaxTxId(partitionData));
+      }
+      // 为什么？
```
zhoujinsong, line 122: remove chinese comments.

ams/ams-server/src/test/resources/config/conf/config.yaml
wangtaohz: Changed to use temp folder generated by junit.

core/src/main/java/com/netease/arctic/op/OverwriteBaseFiles.java
wangtaohz: Removed

zhoujinsong (approved):

PR #252, opened 2022-08-30
[ARCTIC-34][Flink] Support create and get timestamp(3) timestamp(0) type in the arctic table

### LINKED ISSUE #34: [Flink] Support create and get timestamp(3) timestamp(0) type in the arctic table
Scenario: 
Create an iceberg table that has a timestamp-type column with a precision equal to 0 by Flink SQL.

Using the below DDL to create a watermark definition table like the iceberg table would be failed.
`
create table hadoop_catalog.db.dim (id int, name string, opt timestamp(0), primary key (id) not enforced) ;
`

`create table arctic_dim ( watermark for opt as opt - interval '5' second) like hadoop_catalog.db.dim;
`

org.apache.flink.table.api.ValidationException: Invalid data type of time field for watermark definition. The field must be of type TIMESTAMP(p) or TIMESTAMP_LTZ(p), the supported precision 'p' is from 0 to 3, but the time field type is TIMESTAMP(6)

The [TypeToFlinkType](https://github.com/apache/iceberg/blob/master/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/TypeToFlinkType.java#L111) default converts it to flink timestamp type with precision 6.

SteNicholas:

Create an Iceberg table that has a timestamp-type column with a precision equal to 0 by Flink SQL.

Using the below DDL to create a watermark definition table like the iceberg table would be failed.
```create table hadoop_catalog.db.dim (id int, name string, opt timestamp(0), primary key (id) not enforced) ;```

```create table arctic_dim ( watermark for opt as opt - interval '5' second) like hadoop_catalog.db.dim;```

org.apache.flink.table.api.ValidationException: Invalid data type of time field for watermark definition. The field must be of type TIMESTAMP(p) or TIMESTAMP_LTZ(p), the supported precision 'p' is from 0 to 3, but the time field type is TIMESTAMP(6)

The [TypeToFlinkType](https://github.com/apache/iceberg/blob/master/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/TypeToFlinkType.java#L111) default converts it to Flink timestamp type with precision 6.

Fix #34 .

## Brief change log

- Overwrite the `Types`, `FlinkTypeToType` and `TypeToFlinkType` class in Iceberg to covert the timestamp type with precision.

## How was this patch tested?

## Documentation

### DIFF
--- a/flink/v1.12/flink/src/main/java/org/apache/iceberg/flink/FlinkTypeToType.java
+++ b/flink/v1.12/flink/src/main/java/org/apache/iceberg/flink/FlinkTypeToType.java
@@ -1,0 +1,203 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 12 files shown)


zhoujinsong:
@SteNicholas This change will cause format conflict with iceberg leading Iceberg lib cannot read table created by Arctic.

And this break the core rules of arctic.

PR #271, opened 2022-08-31
stop thread while cache InterruptedException

hameizi:
improved https://github.com/NetEase/arctic/issues/270
some thread will not stop while cache InterruptedException

## Brief change log
stop thread while cache InterruptedException

## How was this patch tested?
none

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
@@ -243,6 +243,7 @@ public class ArcticMetaStore {
           Thread.sleep(1000);
         } catch (InterruptedException e) {
           LOG.warn("Signalling thread was interrupted: " + e.getMessage());
+          return;
         }
       } while (!server.isServing());
       startLock.lock();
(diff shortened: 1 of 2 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
```diff
@@ -369,6 +374,7 @@ private static void checkLeader() {
           Thread.sleep(checkLeaderInterval);
         } catch (InterruptedException e) {
           LOG.warn("notLeader thread was interrupted: " + e.getMessage());
+          return;
```
wangtaohz, line 377: Should the checkLeader Thread be interrupted and exit after ams stopped?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
```diff
@@ -321,12 +322,14 @@ private static void syncAndExpiredFileInfoCache(final TServer server) {
...
           } catch (InterruptedException e) {
             LOG.warn("sync and expired file info cache thread was interrupted: " + e.getMessage());
+            return;
           }
         }
         try {
           Thread.sleep(60 * 1000);
         } catch (InterruptedException e) {
           LOG.warn("sync and expired file info cache thread was interrupted: " + e.getMessage());
+          return;
```
wangtaohz, line 332: while(true) is risky, if the thread is doing `task.doTask()` when interrupted, it can't exit.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
hameizi: add while (isLeader.get())

ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
hameizi: drop while(true), only retain while(server.isServering)

wangtaohz (approved):

PR #127, opened 2022-08-04
[ARCTIC-94][Flink] Arctic table supports dimension table join

zstraw:
Resolve #94

PR #277, opened 2022-09-01
[ARCTIC-272][Flink] Sync Flink temporal join Arctic table to Flink 1.15

SteNicholas:

There are some requirements for real-time data widening. Now hive supports lookup join, but this solution is not available for production, and the hive table needs to be loaded into memory. Large tables are prone to oom problems. Besides, neither Iceberg nor Hudi support lookup joins.

Here is a summary proposal:
Flink affords the event time temporal join. The right table will be used as a version table, and its data can be managed in rocksdb instead of memory.

```
-- create a left table, using localtimestamp as event time.
create table source (
  ...,
  arcitc_process_time AS LOCALTIMESTAMP,
  WATERMARK FOR arcitc_process_time AS arcitc_process_time,
) with (...);

create table arctic_dim (...) with ('connector'='arctic', 'dim-table.enabled'='true');
```

select * from source as O left join arctic_dim FOR SYSTEM_TIME AS OF O.arcitc_process_time as P on O.id = P.id;
The arctic source will automatically create a custom watermark strategy if `dim-table.enabled` equals true.

## Brief change log

  - Sync #127 to module arctic-flink-1.15

## How was this patch tested?

## Documentation

### DIFF
--- a/flink/v1.15/flink/pom.xml
+++ b/flink/v1.15/flink/pom.xml
@@ -334,6 +334,21 @@
             <scope>test</scope>
         </dependency>
 
+        <!--   for values test connector     -->
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-planner_${scala.binary.version}</artifactId>
+            <version>${flink.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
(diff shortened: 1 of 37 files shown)


zstraw (approved):
LGTM

YesOrNo828 (approved):

YesOrNo828:
@SteNicholas  Thanks for your contribution. I'll merge to master.

PR #280, opened 2022-09-01
Fix table runtime current snapshot

hzluting:
1. when unKeyed table append between optimize plan and optimize commit, will can't optimize the new append until new append after optimize commit.

## Brief change log

*(For example:)*
  - *use current snapshot id when optimize plan rather than optimize commit *
  - *check whether have new snapshot that not produced by optimize between last plan snapshot and current snapshot*

## How was this patch tested?
- [ ] UnKeyedTable continous optimize


ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BaseOptimizeCommit.java
```diff
@@ -258,15 +258,6 @@ public long commit(TableOptimizeRuntime tableOptimizeRuntime) throws Exception {
                 "add {} new files [{} posDelete files]",
             arcticTable.id(), majorDeleteFiles.size(), deleteDeleteFiles.size(), majorAddFiles.size(),
             addDeleteFiles.size());
-
-        // update table runtime base snapshotId, avoid repeat Major plan
```
wangtaohz, line 262: What about the risk of not update base snapshotId in table runtime after commit?

hzluting: no risk.
Because check base table is changed accord to this base snapshotId is equal to current snapshotId before.
Now accord to whether have new append not produced by optimize between  this base snapshotId and current snapshotId

wangtaohz (approved):

PR #292, opened 2022-09-01
[AMS][HIVE]fix hive schema update and sync ddl  

hameizi:
1.add function moveColBeforePar, and add this logic in every `addColumn` function
2.while before schema is null not sync ddl
3.add DEFAULT_NAME_MAPPING to PROTECTED_PROPERTIES


ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
```diff
@@ -204,6 +209,13 @@ public static boolean isStarted() {
   public static void failover() {
     stopMetaStore();
     AmsRestServer.stopRestServer();
+    isLeader.set(false);
+  }
+
+  public static void shutDown() {
```
wangtaohz, line 215: please change `com.netease.arctic.ams.server.AmsEnvironment#stopAms` to use `ArcticMetaStore.shutDown()` in this PR.

pom.xml
```diff
@@ -29,13 +29,13 @@
...
+        <module>hive</module>
         <module>optimizer</module>
         <module>flink</module>
         <module>spark</module>
-        <module>dist</module>
         <module>trino</module>
-        <module>hive</module>
+        <module>ams</module>
+        <module>dist</module>
     </modules>
```
wangtaohz, line 39: why change the order of modules?

hameizi: because package ams and dist depend on other module, the old order of modules maybe cause ams and dist package with old dependency

ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
hameizi: will change in new commit

wangtaohz (approved):

PR #302, opened 2022-09-02
fix create table cant set colmun struct type

hellojinsilei:
fix create table cant set colmun struct type and timestamp type

## Brief change log

*(For example:)*
  - *Add the 'scan.startup.mode' configuration to the ArcticScanContext in the flink modules*
  - *The flink 1.12 and 1.14 versions are affected.*
  - *Add the documentation in the flink chapter*

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

PR #330, opened 2022-09-06
[Arctic-329] fix mysql/ams-init.sql for optimize_task_history

### LINKED ISSUE #329: [Bug]: mysql init sql error
### What happened?

when create table optimize_task_history
```
ERROR 1171 (42000): All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead
```

### Affects Versions

master/0.3.1

AMS

### How to reproduce

in mysql client
```
source src/main/resources/sql/mysql/ams-init.sql
```

### Relevant log output

```shell
ERROR 1171 (42000): All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead
```

### Anything else

_No response_

### Code of Conduct

wangtaohz:

fix #329

## Brief change log

  - move optimize_task_history upgrade sql to src/main/resources/sql/mysql/upgrade-0.3.0-to-0.3.1.sql

## How was this patch tested?

image
image

## Documentation

### DIFF
--- a/ams/ams-server/src/main/resources/sql/mysql/ams-init.sql
+++ b/ams/ams-server/src/main/resources/sql/mysql/ams-init.sql
@@ -237,7 +237,6 @@ CREATE TABLE `optimize_task_history`
     `cost_time`       bigint(20) DEFAULT NULL COMMENT 'Task cost time',
     `queue_id`        int(11) DEFAULT NULL COMMENT 'Task queue id',
     `task_group_id`   varchar(40) DEFAULT NULL COMMENT 'Task group id',
-    PRIMARY KEY (`task_history_id`,`task_group_id`),
     KEY               `task_group_id_index` (`task_history_id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT 'History of each optimize task';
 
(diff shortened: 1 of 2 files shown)


hameizi:
I think we only need to update ams-init.sql because this exception was introduced in v0.3.1. And ams-init.sql is not executable.

ams/ams-server/src/main/resources/sql/mysql/ams-init.sql
```diff
@@ -237,7 +237,6 @@ CREATE TABLE `optimize_task_history`
     `cost_time`       bigint(20) DEFAULT NULL COMMENT 'Task cost time',
     `queue_id`        int(11) DEFAULT NULL COMMENT 'Task queue id',
     `task_group_id`   varchar(40) DEFAULT NULL COMMENT 'Task group id',
-    PRIMARY KEY (`task_history_id`,`task_group_id`),
```
zhoujinsong, line 240: Why remove this primary key?

wangtaohz: NOT NULL column is required for primary key, but `task_history_id` and `task_group_id` are DEFAULT NULL.

wangtaohz:
>

hameizi:
> > I think we only need to update ams-init.sql because this exception was introduced in v0.3.1. And ams-init.sql is not executable.
> 
> Shall we just change `task_history_id` and `task_group_id` from DEFAULT NULL to NOT NULL in ams-init.sql directly?
> 
> I would have thought adding into upgrade-0.3.0-to-0.3.1.sql was more appropriate.

it's ok

zhoujinsong (approved):

PR #338, opened 2022-09-06
move base.table.max-transaction-id from table property to partitionProperties

hzluting:
  1. make partition properties to unified saved.

## Brief change log
  - *move base.table.max-transaction-id from table property to partitionProperties*


hive/src/main/java/com/netease/arctic/hive/op/HiveOperationTransaction.java
```diff
@@ -173,182 +158,4 @@ public void commit() {
       }
     }
   }
-
-  private class TransactionalTable implements Table {
```
zhoujinsong, line 177: Why remove this `TransactionalTable` implementation? I think we need overwride transactional table `newOverwrite` and `newReplacePartitions` method?

zhoujinsong (changes requested):
@hzluting I left some comment. PTAL.

hive/src/main/java/com/netease/arctic/hive/op/HiveOperationTransaction.java
hzluting: First, I find no use the TransactionalTable anywhere before.
I think use the transactionalTable in HiveOperationTransaction.wrapped is enough?

zhoujinsong (approved):
LGTM.

PR #340, opened 2022-09-07
[ARCTIC-291] [Improvement]: support access Tables tab only for display table info only

Aireed:
fix https://github.com/NetEase/arctic/issues/291
people can use  ams table tab page only to display arctic table detail;

## Brief change log

First, i will support an api to generate a token with catalog,db,table.

second , all tabls tab request url will be /ams/v1/tables/catalogs/{catalog}/dbs/{db}/tables/{table}, i will extract catalog,db,table from url, and check the token.

if token mismatch , return signatureExecption.

if match, return the result of query.

## How was this patch tested?

## Documentation


hameizi (approved):

hameizi:
LGTM

wangtaohz (approved):

PR #350, opened 2022-09-09
[ARCTIC-173][Core] fix unkey table file name parse

### LINKED ISSUE #381: [Bug][core]: The files produced by full optimize output to max txid directory(existed), it will cause commit failed
### What happened?

Now, the hive location rule is table_location/hive/partition_name/txid, Full Optimize must produce the maximum  in already exists txid.
The OverwriteHiveFiles will check all files in old location should deleted.Now, the new files also in old location, so optimize commit will failed.
Optimize should output to a new location for base table.

### Affects Versions

0.3.1

_No response_

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

wangtaohz:

The format of `UnkeyedTable` file name and `KeyedTable` file name are different, now we use the `KeyedTable` file format to parse `UnkeyedTable` file name, and it always return file type BASE, which happens to be correct if `UnkeyedTable`  only have BASE files.

However, after spark support DELETE/UPDATE, `UnkeyedTable` have both POS_DELETE and BASE files, and it brings problem.

Also, fix #381.

## Brief change log

  - add file name parse method in FileUtils for `KeyedTable` and `UnkeyedTable` separately

## How was this patch tested?

## Documentation


zhoujinsong:
@wangtaohz Could you please rename pr title to match parttern: `[ARCTIC-[Issue number]][[Module name]] PR description`

core/src/main/java/com/netease/arctic/data/DefaultKeyedFile.java
```diff
@@ -215,6 +215,7 @@ public int hashCode() {
     return Objects.hash(internalFile.path());
   }
 
+  @Deprecated
```
zhoujinsong, line 218: How about remove this method and replace it with new method?
If not, could you please add some comment to identify which version this method was deprecated and point to the new method ?

core/src/main/java/com/netease/arctic/utils/ConvertStructUtil.java
```diff
@@ -70,14 +73,76 @@ public static com.netease.arctic.ams.api.DataFile convertToAmsDatafile(
...
+        amsDataFile.setFileType(fileMeta.type().name());
+        amsDataFile.setIndex(fileMeta.node().index());
+        amsDataFile.setMask(fileMeta.node().mask());
+      } else if (content == FileContent.POSITION_DELETES) {
+        DefaultKeyedFile.FileMeta fileMeta = FileUtil.parseKeyedFileMetaFromFileName(dataFile.path().toString());
+        amsDataFile.setFileType(DataFileType.POS_DELETE_FILE.name());
+        if (fileMeta.type() == DataFileType.POS_DELETE_FILE) {
+          amsDataFile.setIndex(fileMeta.node().index());
+          amsDataFile.setMask(fileMeta.node().mask());
+        } else if (fileMeta.type() == DataFileType.BASE_FILE) {
```
zhoujinsong, line 100: It seems `FileUtil.parseKeyedFileMetaFromFileName` use node(0, 0) as its default value, so maybe you can use this default value here instead of setting specific node(0, 0)

zhoujinsong (changes requested):
@wangtaohz Hi, I left some comments and PTAL.

core/src/main/java/com/netease/arctic/data/DefaultKeyedFile.java
wangtaohz: deprecated method removed

core/src/main/java/com/netease/arctic/utils/ConvertStructUtil.java
wangtaohz: Done. Now both POS_DELETE_FILE and BASE_FILE use the default (0,0) from fileMeta.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/FullOptimizePlan.java
```diff
@@ -145,8 +150,9 @@ protected boolean checkFullOptimizeInterval(long current, String partitionToPath
 
   /**
    * check whether node task need to build
+   *
    * @param posDeleteFiles pos-delete files in node
-   * @param baseFiles base files in node
+   * @param baseFiles      base files in node
```
zhoujinsong, line 155: It seems add some extra white spaces.

wangtaohz: I tried Auto Format in IDEA, it added the white spaces automatically. It aligns with the`pos-delete` on the previous line.

zhoujinsong: okay

zhoujinsong (approved):
LGTM.

PR #359, opened 2022-09-14
[Spark]fix update bug

hellojinsilei:
Fix the following bugs
1.Updating column value with decimal、double and float type will fail in Spark
2.Updating data across multiple partitions will fail in Spark

## Brief change log

  - *Fix bugs related to Spark SQL Update*

## How was this patch tested?

### DIFF
--- a/core/src/main/java/com/netease/arctic/io/writer/SortedPosDeleteWriter.java
+++ b/core/src/main/java/com/netease/arctic/io/writer/SortedPosDeleteWriter.java
@@ -80,6 +80,17 @@ public class SortedPosDeleteWriter<T> implements Closeable {
     this(appenderFactory, fileFactory, format, mask, index, partitionKey, DEFAULT_RECORDS_NUM_THRESHOLD);
   }
 
+  public SortedPosDeleteWriter(FileAppenderFactory<T> appenderFactory,
+                               OutputFileFactory fileFactory,
+                               FileFormat format,
+                               StructLike partitionKey) {
+    this.appenderFactory = appenderFactory;
+    this.fileFactory = fileFactory;
+    this.format = format;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/netease/arctic/io/writer/SortedPosDeleteWriter.java
```diff
@@ -80,6 +80,17 @@ public SortedPosDeleteWriter(FileAppenderFactory<T> appenderFactory,
     this(appenderFactory, fileFactory, format, mask, index, partitionKey, DEFAULT_RECORDS_NUM_THRESHOLD);
   }
 
+  public SortedPosDeleteWriter(FileAppenderFactory<T> appenderFactory,
+                               OutputFileFactory fileFactory,
+                               FileFormat format,
+                               StructLike partitionKey) {
+    this.appenderFactory = appenderFactory;
```
zhoujinsong, line 87: How about use `this(appenderFactory, fileFactory, format, 0, 0, partitionKey, DEFAULT_RECORDS_NUM_THRESHOLD);` instead.

zhoujinsong (commented):
@zhoujinsong I left some comment. PTAL.

core/src/main/java/com/netease/arctic/io/writer/SortedPosDeleteWriter.java
hellojinsilei: Using this constructor, the generated POS-delete file has the wrong name

zhoujinsong: Okay, but I donnot think using a `NULL` data node is a good policy. We could improve this in another PR.

zhoujinsong (approved):
LGTM.

PR #373, opened 2022-09-16
[SPARK] Update Primary Key Will Throw Exception

hellojinsilei:

*No errors will be reported when updating the primary key. An exception will now be thrown*

## How was this patch tested?


zhoujinsong (commented):
@hellojinsilei Is column name case sensitive in Spark SQL?

zhoujinsong (approved):
LGTM.

PR #383, opened 2022-09-19
[ARCTIC-351][OPTIMIZE]:Supports external system registration and management of Optimizer

### LINKED ISSUE #351: [Improvement][OPTIMIZE]:Supports external system registration and management of Optimizer
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?page=4&q=is%3Aissue) and found no similar issues.

now external system can't  registration and management of Optimizer

### How should we improve?

1.Provides interfaces for registering and unregistering Optimizers
2.Provides an interface to get the optimize Group

### Code of Conduct

hameizi:
resolve #351 

## Brief change log

register optimizer when there is not optimizerId in start optimizer config.

## How was this patch tested?
start optimizer with -a {ams-url} -qn {optimizerGroupName} -p {executor-parallel} -m {executor-memory-size}

## Documentation


Aireed (approved):
LGTM

optimizer/src/main/java/com/netease/arctic/optimizer/local/LocalOptimizer.java
```diff
@@ -150,6 +151,10 @@ public void stop() throws IOException {
 
   public static void main(String[] args) throws CmdLineException {
     OptimizerConfig optimizerConfig = new OptimizerConfig(args);
+    if (optimizerConfig.getOptimizerId() == null || optimizerConfig.getOptimizerId().isEmpty() ||
```
wangtaohz, line 154: I think not only external optimizer but also "internal" optimizer can `register` to ams, and then the start script of external/internal optimizer could be unified.

wangtaohz: It confusing that the start script of internal optimizer use `--queue-id`, but external optimizer use `--queue-name`

hameizi: > It confusing that the start script of internal optimizer use `--queue-id`, but external optimizer use `--queue-name`

the start script of `internal` optimize is generate with ams , so the `queue-id` is easy to get. but external optimizer is started by user who get `queue-id` diffcult

wangtaohz: what if user starts external optimizer but set optimizerId?

hameizi: > what if user starts external optimizer but set optimizerId?

it also work well.

PR #424, opened 2022-09-28
[ARCTIC-393][AMS]Improve the high concurrency processing capability of the allocateTransactionId interface

### LINKED ISSUE #393: [Improvement][AMS]: Improve the high concurrency processing capability of the allocateTransactionId interface
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?page=4&q=is%3Aissue) and found no similar issues.

When multiple tasks(like upsert partial fields of arctic table with Flink) write data to the same table, apply for TransActionID at the same time. So we should improve the high concurrency processing capability of the allocateTransactionId interface. 

### How should we improve?

we can compare the old and new transactionid when update row in mysql, like `update table_metadata where tablename={} and current_tx_id={old_tx_id}`. And we get the update return value(Number of rows affected), if the value greater than 0 means update success else retry more(just like cas).

### Code of Conduct

hameizi:
resolve #393
When multiple tasks(like upsert partial fields of arctic table with Flink) write data to the same table, apply for TransActionID at the same time. So we should improve the high concurrency processing capability of the allocateTransactionId interface.

## Brief change log

add table lock in ArcticTransactionService

## How was this patch tested?

## Documentation


Aireed (approved):
LGTM

PR #480, opened 2022-10-17
[ARCTIC-128] [Optimize] limit the file-count of minor optimize plan

### LINKED ISSUE #128: [optimize] risk of ams memory usage and task execute timeout, when process a large number of small files

wangtaohz:

to fix #128, when a large number of smalll files in change table, minor optimize may plan tasks with too much files, which costs a long time to execute, and also brings risk of out of memory for ams/optimizer.

## Brief change log

  - add table properties "optimize.file-count.max" to limit max file count of one Optimize plan (only Minor Optimize now)
  - default value of "optimize.file-count.max" is 100,000
  - before Minor Optimize plan, find all changes files which are not optimized before
  - then find up to 100,000 files with smallest transactionId from these unoptimized change files, and plan

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/config/ServerTableProperties.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/config/ServerTableProperties.java
@@ -27,9 +27,6 @@ import java.util.Set;
 public class ServerTableProperties {
 
   public static final Set<String> HIDDEN_EXPOSED = new HashSet<>();
-  
-  public static final String OPTIMIZE_FILE_COUNT_MAX = "optimize.file-count.max";
-  public static final int OPTIMIZE_FILE_COUNT_MAX_DEFAULT = 100000;
 
   static {
     HIDDEN_EXPOSED.add(TableProperties.BASE_TABLE_MAX_TRANSACTION_ID);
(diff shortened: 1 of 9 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/config/ServerTableProperties.java
```diff
@@ -27,6 +27,9 @@
 public class ServerTableProperties {
 
   public static final Set<String> HIDDEN_EXPOSED = new HashSet<>();
+  
+  public static final String OPTIMIZE_FILE_COUNT_MAX = "optimize.file-count.max";
```
zhoujinsong, line 31: what about  `optimize.max-file-count` ?

zhoujinsong (commented):
@wangtaohz  I left some comments.

zhoujinsong:
@wangtaohz 

How about put `optimize.max-file-count` property into `TableProperties` and add this new property to user documention?

zhoujinsong (approved):
LGTM.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/config/ServerTableProperties.java
wangtaohz: changed to `optimize.max-file-count`

wangtaohz:
> @wangtaohz
> 
> How about put `optimize.max-file-count` property into `TableProperties` and add this new property to user documention?

Added `optimize.max-file-count` to `TableProperties`, and alse documention.

PR #505, opened 2022-10-23
[ARCTIC-504]When the catalog name contains line-through, terminal executes sql with an error

### LINKED ISSUE #504: [Bug]: When the catalog name contains line-through, terminal executes sql with an error
### What happened?

image
When the catalog name contains line-through, terminal executes sql with an error

### Affects Versions

Versions earlier than 0.3.2

AMS

### How to reproduce

Terminal executes sql, when the catalog name contains line-through

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

shendanfengg:
Resolve #504 

* Resolve catalog names with special characters may lead terminal execute sql failed

## Brief change log
  - *Add back-quotes when execute "use catalog" 

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/TerminalService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/TerminalService.java
@@ -61,6 +61,7 @@ public class TerminalService {
   private static int sessionId = 1;
   private static final Map<Integer, SqlRunningInfo> sqlSessionInfoCache = new HashMap<>();
   private static final SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss:SSS");
+  private static final String QUOTE  = "`";
 
   public static LogInfo getLogs(int sessionId) {
     return new LogInfo(sqlSessionInfoCache.get(sessionId).getLogStatus(), sqlSessionInfoCache.get(sessionId).getLogs());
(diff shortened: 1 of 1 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/TerminalService.java
```diff
@@ -149,7 +149,7 @@ public static SesssionInfo executeSql(String catalog, String sqls) {
         }
         try {
           sqlRunningInfo.getLogs().add(df.format(new Date()) + " use " + catalog);
-          spark.sql("use " + catalog);
+          spark.sql("use " + "`" + catalog + "`");
```
YesOrNo828, line 152: PLS uses the `static final` keywords to define the String parameter: "`".

e.g. private static final String QUOTE = "`";

YesOrNo828 (commented):
@shendanfengg, is there any other places should also quote the catalog?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/TerminalService.java
shendanfengg: > @shendanfengg, is there any other places should also quote the catalog?

No need to quote other place,only ams-server terminal needs this change

shendanfengg: > PLS uses the `static final` keywords to define the String parameter: "`".
> 
> e.g. private static final String QUOTE = "`";

It has been fixed

YesOrNo828 (approved):
LGTM

PR #523, opened 2022-10-27
[ARCTIC-486][FLINK]: Fix 'Watermark may loss if split read too fast'

### LINKED ISSUE #486: [Bug][Flink]:  Watermark may loss if split read too fast.
### What happened?

If there is an eventTime field defined in arctic table by 'create table t (wm as cast(op_time as timestamp(3)), watermark for wm as wm) like XXX' , watermark may not emitted.

### Affects Versions

<=0.3.2

Flink

### How to reproduce

1. Define a watermark by sql 'create table t (wm as cast(op_time as timestamp(3)), watermark for wm as wm) like XXX'.
2. Then group by tumble window, or other sql need watermark.
3. Insert into source table with one record.

You may find that there is no watermark in Flink web ui.

### Relevant log output

_No response_

### Anything else

The cause may be the case like:
The watermark in {@link WatermarkOutputMultiplexer.OutputState} which is associated with a split has been updated, but watermark may not been emitted for that when {@link WatermarkOutputMultiplexer#onPeriodicEmit} called, the outputState has been removed by {@link WatermarkOutputMultiplexer#unregisterOutput(String)} after split finished.

### Code of Conduct

zstraw:
FIX #486 

If there is an eventTime field defined in arctic table by 'create table t (wm as cast(op_time as timestamp(3)), watermark for wm as wm) like XXX' , watermark may not emitted.

Fix this issue.

## Brief change log
  - *The flink 1.12 and 1.14, 1.15 versions are affected.*

## How was this patch tested?

## Documentation

### DIFF
--- a/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/util/FlinkClassUtil.java
+++ b/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/util/FlinkClassUtil.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netease.arctic.flink.util;
+
+import org.apache.flink.api.connector.source.ReaderOutput;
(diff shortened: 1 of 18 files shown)


flink/v1.12/flink/src/main/java/com/netease/arctic/flink/util/FlinkClassUtil.java
```diff
@@ -0,0 +1,64 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Nullable;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+import static com.netease.arctic.flink.util.ReflectionUtil.getField;
+
+public class FlinkClassUtil {
```
YesOrNo828, line 32: rename to FlinkClassReflectUtil? 
could you add some annotation about this class or methods?

flink/v1.12/flink/src/test/java/com/netease/arctic/flink/table/TestUnkeyedOverwrite.java
```diff
@@ -99,7 +106,7 @@ public void after() {
     }
   }
 
-  @Test
+  @Test(timeout = 3000)
```
YesOrNo828, line 109: 3 seconds?

YesOrNo828 (commented):
@zstraw  Thanks for fixing it. 
And I left some comments.

YesOrNo828 (approved):
LGTM

PR #533, opened 2022-10-28
Fix flink write change of compatible hive table

shidayang:


hellojinsilei (approved):

YesOrNo828:
@shidayang  Could you describe Why are the changes needed?
Is there any impact on the Optimization module?

shidayang:
Because the parquet type of adapted hive table is different. optimize module had adapted

YesOrNo828 (approved):
@shidayang  It makes sense to me.

PR #559, opened 2022-11-03
[ARCTIC-558] Drop table operation will leave an empty table directory in hdfs

### LINKED ISSUE #558: [Bug]: Drop table operation will leave an empty table directory in hdfs
### What happened?

Drop table operation will leave an empty table directory on hdfs.
When you create a new table with the same name as the deleted table, the ams will throw an exception that the path already exists.

### Affects Versions

0.3.x

Core

### How to reproduce

Drop an arctic table

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

shendanfengg:
Fix #558 

* Now drop arctic table will leave an empty directory in hdfs

## Brief change log

  -After drop table, delete table directory 

## How was this patch tested?

## Documentation

### DIFF
--- a/hive/src/main/java/com/netease/arctic/hive/catalog/ArcticHiveCatalog.java
+++ b/hive/src/main/java/com/netease/arctic/hive/catalog/ArcticHiveCatalog.java
@@ -125,6 +125,8 @@ public class ArcticHiveCatalog extends BaseArcticCatalog {
 
   @Override
   protected void doDropTable(TableMeta meta, boolean purge) {
+    // drop hive table only drop hive table metadata
+    // delete data files will use BaseArcticCatalog
     try {
       hiveClientPool.run(client -> {
         client.dropTable(meta.getTableIdentifier().getDatabase(),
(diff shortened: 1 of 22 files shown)


hive/src/main/java/com/netease/arctic/hive/catalog/ArcticHiveCatalog.java
```diff
@@ -137,6 +136,7 @@ protected void doDropTable(TableMeta meta, boolean purge) {
     } catch (TException | InterruptedException e) {
       throw new RuntimeException("Failed to drop table:" + meta.getTableIdentifier(), e);
     }
+    super.doDropTable(meta, purge);
```
zhoujinsong, line 139: Why moving droping hive table first?

zhoujinsong (commented):
@shendanfengg  I left some comments.

hive/src/main/java/com/netease/arctic/hive/catalog/ArcticHiveCatalog.java
shendanfengg: When super.doDropTable is called and purge == true, all directories of the table will be remove. And then when try to drop hive table, the path will not be found

zhoujinsong (approved):
LGTM.

PR #568, opened 2022-11-08
[ARCTIC-561] Fix 561. When flink job fails over, flink job will commit a wrong transaction id

### LINKED ISSUE #561: [Bug]: When flink job fails over, flink job will commit a wrong transaction id
### What happened?

When flink job fail over, flink job will commit a wrong transaction id. 
The wrong transaction id is 
image
And the right transaction id is 
image
The data associated with the wrong transaction id will be lost.
The problem can only be resolved by rebuilding the table and rewriting the data.

### Affects Versions

0.3.x

Flink

### How to reproduce

Insert data through flink and then make flink job failover

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

zstraw:
FIX #561 

## Brief change log

  - *The flink 1.12 and 1.14 , 1.15versions are affected.*

## How was this patch tested?

## Documentation

### DIFF
--- a/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
+++ b/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
@@ -26,6 +26,11 @@ import com.netease.arctic.flink.util.ArcticUtils;
 import com.netease.arctic.table.ArcticTable;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.api.common.state.ListState;
+import org.apache.flink.api.common.state.ListStateDescriptor;
+import org.apache.flink.api.common.typeutils.base.LongSerializer;
+import org.apache.flink.runtime.state.StateInitializationContext;
+import org.apache.flink.runtime.state.StateSnapshotContext;
 import org.apache.flink.streaming.api.operators.AbstractStreamOperator;
 import org.apache.flink.streaming.api.operators.BoundedOneInput;
 import org.apache.flink.streaming.api.operators.OneInputStreamOperator;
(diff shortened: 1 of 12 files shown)


zhoujinsong:
@zstraw Hi, Could you please add some description about the cause and solution of this bug.

zstraw:
> @zstraw Hi, Could you please add some description about the cause and solution of this bug.

The cause of this bug is that, transactionId is applied by jobId+checkpointId in flink file writer , and it would be the same if jobId+checkpointId is identical. When flink job failovers, the checkpointId would be initilized by value 0, which makes the applied transactionId is identical with the first transactionId in the job runs for the first time.

TransactionId can be applied by UUID+jobId+ckpId, so that transactionId can be identical once the job failovers or restarts.

YesOrNo828:
Why could the client apply the same transaction twice if this transaction has been committed?
This question came to my mind. Could anyone have a reasonable explanation?

zhoujinsong:
> Why could the client apply the same transaction twice if this transaction has been committed? This question came to my mind. Do I miss the context? Could anyone have a reasonable explanation?

Row sequence number is composed of transaction id and row offset, wich is used to compare bettwen different rows when optimizing and MOR processing, So the transaction id is required to be incresement.

zhoujinsong (approved):
LGTM.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
```diff
@@ -126,6 +157,11 @@ private Long getTransactionId() {
     return transaction;
```
YesOrNo828, line 157: Maybe adding a log message to show how to produce the signature is better. So it is clear what combines the signature.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
```diff
@@ -104,6 +109,32 @@ public void open() {
...
+    checkpointState =
+        context.getOperatorStateStore()
+            .getListState(
+                new ListStateDescriptor<>(
+                    subTaskId + "-task-file-writer-state",
+                    LongSerializer.INSTANCE));
+    
+    if (context.isRestored()) {
+      // get last ckp num from state when failover continuously
+      checkpointId = checkpointState.get().iterator().next();
```
YesOrNo828, line 126: Should the value of `checkpointId` plus one?

YesOrNo828 (approved):

PR #593, opened 2022-11-11
[Feature][AMS]: support show table info for native iceberg in ams dashboard

### LINKED ISSUE #531: [Feature][AMS]: support show table info for native iceberg in ams dashboard
### Description

support show table schema and fileinfo, snpashotinfo of native iceberg table. 
Attention: when get native iceberg table fileinfo by iceberg interface we should support get partial file info from iceberg manifest file to avoid oom and and show pages

### Use case/motivation

_No response_

### Describe the solution

_No response_

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

hameizi:
resolve #531 

## Brief change log

add logic: get native iceberg table snapshot info

## How was this patch tested?

## Documentation


core/src/main/java/com/netease/arctic/utils/SnapshotFileUtil.java
```diff
@@ -126,4 +130,55 @@ private static void getDeleteFiles(
       LOG.error("close manifest file error", exception);
     }
   }
+
+  public static void getNativeIcebergDeleteFiles(
```
zhoujinsong, line 134: How about rename method name to `getIcebergDeleteFiles` and make `getDeleteFiles` depende on this one?

zhoujinsong (commented):
@hameizi  I left some comments. PTAL.

zhoujinsong:
@hameizi Should the WEB UI hidden `Files` tabls for iceberg tables?

hameizi:
> @hameizi Should the WEB UI hidden `Files` tabls for iceberg tables?

This PR not support native iceberg file info in web ui

core/src/main/java/com/netease/arctic/utils/SnapshotFileUtil.java
hameizi: There is some difference between arctic and iceberg delete file. Arctic eq-delete file is datafile and all delete file is pos-delete file, but in iceberg there are both eq-delete and pos-delete type exist in table.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/model/AMSDataFileInfo.java
```diff
@@ -18,10 +18,9 @@
 
 package com.netease.arctic.ams.server.model;
 
-import com.netease.arctic.ams.api.DataFileInfo;
+import com.netease.arctic.ams.api.DataFile;
```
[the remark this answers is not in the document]
hameizi: remove in new commit.

core/src/main/java/com/netease/arctic/utils/SnapshotFileUtil.java
zhoujinsong: Yes, but Getting all changed delete files from iceberg's entries table is also the same for Arctic and Iceberg Table as I  know, and they share many deplicate code now.

zhoujinsong (approved):
LGTM.

PR #594, opened 2022-11-14
[Flink][bugfix] Fixed the problem of duplicate primary key data in Arctic table with upsert feature enabled

### LINKED ISSUE #592: [Bug][Flink]: Read duplicated data from keyed table, even though the upsert mode is enabled.
### What happened?

Read duplicated data from keyed table, even though the upsert mode is enabled.
Write to keyed table by flink.

### Affects Versions

all

Flink

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

The cause of this bug is that the shuffle policy is none before arctic writer operator. It leads to sending the records with same key to different subtask. Then the reader will read data in random order.

### Code of Conduct

lklhdu:
fixed #592 

*When an Arctic table with the write.upsert feature enabled is read, data with the same primary key will exist, which should be fixed*

## Brief change log

  - *replace the shuffle policy involved in write operations*

## How was this patch tested?

### DIFF
--- a/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/FlinkSink.java
+++ b/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/FlinkSink.java
@@ -91,7 +91,7 @@ public class FlinkSink {
     private Properties producerConfig;
     private String topic;
     private boolean overwrite = false;
-    private DistributionHashMode distributionMode = DistributionHashMode.NONE;
+    private DistributionHashMode distributionMode = null;
 
     private Builder() {
     }
--- a/flink/v1.12/flink/src/test/java/com/netease/arctic/flink/table/TestKeyed.java
+++ b/flink/v1.12/flink/src/test/java/com/netease/arctic/flink/table/TestKeyed.java
@@ -380,6 +380,68 @@ public class TestKeyed extends FlinkTestBase {
   }
 
   @Test
+  public void testFileUpsertWithSamePrimaryKey() throws Exception {
+    List<Object[]> data = new LinkedList<>();
+    data.add(new Object[]{RowKind.INSERT, 1000004, "a", LocalDateTime.parse("2022-06-17T10:10:11.0")});
+    data.add(new Object[]{RowKind.INSERT, 1000004, "b", LocalDateTime.parse("2022-06-17T10:10:11.0")});
+    data.add(new Object[]{RowKind.INSERT, 1000011, "e", LocalDateTime.parse("2022-06-17T10:10:11.0")});
+    data.add(new Object[]{RowKind.INSERT, 1000011, "f", LocalDateTime.parse("2022-06-17T10:10:11.0")});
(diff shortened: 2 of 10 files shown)


zhoujinsong:
@lklhdu Could you please add some description of the case and solution of this bug.
And I'm very curious why only Flink 1.14/1.15 version is affected.

lklhdu:
> @lklhdu Could you please add some description of the case and solution of this bug. And I'm very curious why only Flink 1.14/1.15 version is affected.

lklhdu:
> @lklhdu Could you please add some description of the case and solution of this bug. And I'm very curious why only Flink 1.14/1.15 version is affected.

Actually, I also found the same problem with upsert on Flink 1.12 later. The reason for this problem is that when the write distribution mode is null, no shuffle is performed on the data, whereas for data with the same primary key, a shuffle should be performed by the primary key to keep the data in order.

lklhdu:
> > @lklhdu Could you please add some description of the case and solution of this bug. And I'm very curious why only Flink 1.14/1.15 version is affected.
> 
> Actually, I also found the same problem with upsert on Flink 1.12 later. The reason for this problem is that when the write distribution mode is null, no shuffle is performed on the data, whereas for data with the same primary key, a shuffle should be performed by the primary key to keep the data in order.

When I use Flink SQL client to add two data with the same primary key to an Arctic table, in DistributionHashMode.NONE mode, the cdc data read may be out of order, causing the delete data to not work, like the following:
-D 1,eric,2022-07-01T12:32
-D 1,eric,2022-07-01T12:32
+I 1,eric,2022-07-01T12:32
+I 1,eric,2022-07-01T12:32

zhoujinsong:
> > @lklhdu Could you please add some description of the case and solution of this bug. And I'm very curious why only Flink 1.14/1.15 version is affected.
> 
> Actually, I also found the same problem with upsert on Flink 1.12 later. The reason for this problem is that when the write distribution mode is null, no shuffle is performed on the data, whereas for data with the same primary key, a shuffle should be performed by the primary key to keep the data in order.

So the default `None` distribution mode in `FlinkSink.Builder` makes the table property `write.distribution.hash-mode` is not used, and all Flink version is affected. Is my understanding correct？

Can we add some unite test to validate this case?

zhoujinsong (approved):
LGTM.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/FlinkSink.java
```diff
@@ -91,7 +91,7 @@ public static class Builder {
     private Properties producerConfig;
     private String topic;
     private boolean overwrite = false;
-    private DistributionHashMode distributionMode = DistributionHashMode.NONE;
+    private DistributionHashMode distributionMode = null;
```
YesOrNo828, line 94: Why does `distributionMode` change to null?

YesOrNo828 (approved):

PR #597, opened 2022-11-14
[Spark]: fix un-keyed table transaction id not change for each execute

baiyangtx:
fixbug;  for unkeyed hive table.  file transaction id not changed for each spark sql executed

## Brief change log

- change unkeyed table transactioId field from static to private

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

PR #623, opened 2022-11-16
[ARCTIC-622][Improvement]: fix file upload when use derby as storage

Aireed:
resolve https://github.com/NetEase/arctic/issues/622

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/mapper/derby/DerbyPlatformFileInfoMapper.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/mapper/derby/DerbyPlatformFileInfoMapper.java
@@ -22,5 +22,5 @@ public interface DerbyPlatformFileInfoMapper extends PlatformFileInfoMapper {
 
   // get file content encoded by base64 by fileId
   @Select("select id from " + TABLE_NAME + " where file_content_b64=#{content} limit 1")
-  String getFileId(@Param("content") String content);
+  Integer derbyGetFileId(@Param("content") String content);
 }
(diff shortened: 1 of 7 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/mapper/derby/DerbyPlatformFileInfoMapper.java
```diff
@@ -0,0 +1,26 @@
...
+  String TABLE_NAME = "platform_file_info";
+  /**
+   * add a file with content encoded by base64
+   */
+  @Insert("insert into " + TABLE_NAME + "(file_name,file_content_b64)" +
+          "values(#{fileInfo.fileName},#{fileInfo.fileContent})")
+  void addFile(@Param("fileInfo") PlatformFileInfo platformFileInfo);
+
+  // get file content encoded by base64 by fileId
+  @Select("select id from " + TABLE_NAME + " where file_content_b64=#{content}")
```
zhoujinsong, line 24: When there are multiple files with the same content，multiple `id` will be obtained.

zhoujinsong (commented):
@Aireed I left some comments.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/mapper/derby/DerbyPlatformFileInfoMapper.java
Aireed: done， add limit 1 to keep only one result

zhoujinsong (approved):
LGTM.

PR #625, opened 2022-11-17
[ARCTIC-629][SPARK]Fix spark update has no filter will throw exception

### LINKED ISSUE #629: [Bug]: Spark sql update/delete if has no filter will throw exception
### What happened?

If the filter is not added to the update/delete sql, it will throw exception

### Affects Versions

0.3.2

Spark

### How to reproduce

these sqls will throw exception:
update table set col = 'xxx'
delete from table

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hellojinsilei:

fix #629 
when update or delete has no filter, will throw exception

## How was this patch tested?

### DIFF
--- a/spark/v3.1/spark/src/main/scala/com/netease/arctic/spark/sql/optimize/RewriteDeleteFromArcticTable.scala
+++ b/spark/v3.1/spark/src/main/scala/com/netease/arctic/spark/sql/optimize/RewriteDeleteFromArcticTable.scala
@@ -24,7 +24,7 @@ import com.netease.arctic.spark.sql.utils.ArcticRewriteHelper
 import com.netease.arctic.spark.table.{ArcticSparkTable, SupportsExtendIdentColumns, SupportsUpsert}
 import com.netease.arctic.spark.writer.WriteMode
 import org.apache.spark.sql.SparkSession
-import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, Expression, Literal}
+import org.apache.spark.sql.catalyst.expressions.{Alias, AttributeReference, EqualTo, Expression, Literal}
 import org.apache.spark.sql.catalyst.plans.logical._
 import org.apache.spark.sql.catalyst.rules.Rule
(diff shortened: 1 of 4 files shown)


baiyangtx (approved):
LGTM

zhoujinsong:
@hellojinsilei Is there a issue about this bug? If not you better be able create one so user can find that when meeting this bug.

spark/v3.1/spark/src/main/scala/com/netease/arctic/spark/sql/optimize/RewriteDeleteFromArcticTable.scala
```diff
@@ -40,7 +40,12 @@ case class RewriteDeleteFromArcticTable(spark: SparkSession) extends Rule[Logica
       val r = asTableRelation(table)
       val upsertWrite = r.table.asUpsertWrite
       val scanBuilder = upsertWrite.newUpsertScanBuilder(r.options)
-      pushFilter(scanBuilder, condition.get, r.output)
+      if (condition.isEmpty) {
+        val cond = EqualTo(Literal(1), Literal(1))
```
pan3793, line 44: `Literal.TrueLiteral` is more straightforward

spark/v3.1/spark/src/main/scala/com/netease/arctic/spark/sql/optimize/RewriteDeleteFromArcticTable.scala
```diff
@@ -40,7 +40,12 @@ case class RewriteDeleteFromArcticTable(spark: SparkSession) extends Rule[Logica
...
+        pushFilter(scanBuilder, cond, r.output)
+      } else {
+        pushFilter(scanBuilder, condition.get, r.output)
+      }
```
pan3793, line 48:
```suggestion
      condition match {
        case Some(cond) => pushFilter(scanBuilder, condition.get, r.output)
        case None => pushFilter(scanBuilder, Literal.TrueLiteral, r.output)
      }
```

PR #654, opened 2022-11-21
[Arctic-479][MOR]: fix getting null sequence from manifest entry when plan change table

### LINKED ISSUE #479: [Bug]: For Arctic keyd table, the out-of-order TransactionId lead to inconsistent data
### What happened?

Arctic keyed table's TransactionId  may be out-of order, if spark and flink write into change table at the same time.

### Affects Versions

master/0.3.1

Core, AMS

### How to reproduce

1.Spark write into change, begin transaction1, get tid = 1
2.Flink write into change,  begin transaction2, get tid = 2
3.Fllink commit transaction2 with tid = 2
4.Minor Optimize transfer change files to base store with tid = 2
5.Spark commit transaction1 with tid = 1
6.Optimize/MOR will ignore change files with tid = 1, and lead to data loss

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

wangtaohz:

To fix #479 , Iceberg EntriesTables return Entry from manifest with null sequence, the right sequence should be inherited from manifest metadata (in manifest list).

## Brief change log

  - use the `rows` API of `org.apache.iceberg.DataTask` which uses the inherited sequence number

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/scan/BaseChangeTableIncrementalScan.java
+++ b/core/src/main/java/com/netease/arctic/scan/BaseChangeTableIncrementalScan.java
@@ -19,30 +19,33 @@
 package com.netease.arctic.scan;
 
 import com.netease.arctic.data.DefaultKeyedFile;
-import com.netease.arctic.iceberg.optimize.InternalRecordWrapper;
 import com.netease.arctic.io.ArcticHadoopFileIO;
 import com.netease.arctic.table.ChangeTable;
 import com.netease.arctic.table.TableProperties;
 import com.netease.arctic.utils.FileUtil;
-import com.netease.arctic.utils.ManifestEntryFields;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.iceberg.DataFile;
 import org.apache.iceberg.DataFiles;
+import org.apache.iceberg.DataTask;
+import org.apache.iceberg.FileScanTask;
+import org.apache.iceberg.MetadataTableType;
 import org.apache.iceberg.Metrics;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netease/arctic/scan/BaseChangeTableIncrementalScan.java
```diff
@@ -208,25 +213,32 @@ private DataFile buildDataFile(GenericRecord dataFile, String filePath, StructLi
...
-  @SuppressWarnings("unchecked")
-  public Metrics buildMetrics(GenericRecord dataFile) {
-    return new Metrics((Long) dataFile.getField(DataFile.RECORD_COUNT.name()),
-        (Map<Integer, Long>) dataFile.getField(DataFile.COLUMN_SIZES.name()),
-        (Map<Integer, Long>) dataFile.getField(DataFile.VALUE_COUNTS.name()),
-        (Map<Integer, Long>) dataFile.getField(DataFile.NULL_VALUE_COUNTS.name()),
-        (Map<Integer, Long>) dataFile.getField(DataFile.NAN_VALUE_COUNTS.name()),
-        (Map<Integer, ByteBuffer>) dataFile.getField(DataFile.LOWER_BOUNDS.name()),
-        (Map<Integer, ByteBuffer>) dataFile.getField(DataFile.UPPER_BOUNDS.name()));
+  private int dataFileFieldIndex(String fieldName) {
```
wangtaohz, line 227: It's better to use entriesTable's schema.

zhoujinsong (approved):
LGTM.

PR #687, opened 2022-11-22
[Flink][bugfix] fixed the problem that passing a null value to some fields would be given a default value

### LINKED ISSUE #683: [Bug]: When writing data with a null value to the logstore of an arctic table, a non-null default value is assigned to the field
### What happened?

When writing data with a null value to the log-store of an arctic table, a non-null default value is assigned to the field. For example: in the case of a table with schema (int, int, double, float, string), inserting a value of (1, null, null, null, null, 'a') and then reading this data from the log-store of the arctic table will return (1, 0, 0.0, 0.0, 'a')

### Affects Versions

0.3.2

Flink

### How to reproduce

First create a test table with schema (int, int, double, float, string), then write data to the logstore of the arctic table by reading the contents of the kafka topic,such as (1, null, null, null, null, 'a') and finally read the logstore of the arctic table to get the data with the default values (1, 0, 0.0, 0.0, 'a')

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

lklhdu:
fix #683 

When users insert data into the log-store of arctic table via Flink, some fields of the entire data have a null value, but when the data is read from the log-store these fields with a null value are given a default value, which should be fixed

## Brief change log

  - *Added logic for null value determination*
  - *The flink 1.12, 1.14, 1.15 versions are affected.*

## How was this patch tested?

### DIFF
--- a/flink/v1.12/flink-format/src/main/java/com/netease/arctic/flink/shuffle/LogRecordV1.java
+++ b/flink/v1.12/flink-format/src/main/java/com/netease/arctic/flink/shuffle/LogRecordV1.java
@@ -255,19 +255,25 @@ public class LogRecordV1 implements LogData<RowData>, Serializable {
 
         @Override
         public FieldGetter<RowData> createFieldGetter(Type type, int pos) {
+          final FieldGetter<RowData> fieldGetter;
           switch (type.typeId()) {
             case BOOLEAN:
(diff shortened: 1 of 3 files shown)


zhoujinsong (approved):
LGTM.

PR #706, opened 2022-11-22
[Arctic][AMS] admin-configured default table properties with adding into catalog #171

### LINKED ISSUE #171: [Improvement]Support admin-configured default table properties
Currently, to set user-configued properties of a table, it can be achieved by 1. adding table properties in create table DDL or 2.  add dynamic table properties in a DML. Whereas for a multi-user platform, it is common to share options such as `write.target-file-size-bytes` within a group of users. Thus, administrators should be capable to set service-wide options with default vaules instead of setting shared options by users in their SQLs repeatly.

nicochen:
Resolve #171

## Brief change log
  - *Add logical to filter  default properties for a catalog with specific prefix in `TablePropertiesUtil`.*
  - *Add logical to  set default properties into table metadata when creating tablemeta in `TableMetaBuilder`*

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/catalog/BaseArcticCatalog.java
+++ b/core/src/main/java/com/netease/arctic/catalog/BaseArcticCatalog.java
@@ -44,6 +44,7 @@ import com.netease.arctic.table.UnkeyedTable;
 import com.netease.arctic.trace.CreateTableTransaction;
 import com.netease.arctic.utils.CatalogUtil;
 import com.netease.arctic.utils.ConvertStructUtil;
+import com.netease.arctic.utils.TablePropertyUtil;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hadoop.fs.Path;
 import org.apache.iceberg.PartitionSpec;
@@ -433,8 +434,8 @@ public class BaseArcticCatalog implements ArcticCatalog {
 
     @Override
     public ArcticTable create() {
-      doCreateCheck();
       ConvertStructUtil.TableMetaBuilder builder = createTableMataBuilder();
+      doCreateCheck();
       TableMeta meta = builder.build();
       ArcticTable table = doCreateTable(meta);
       createTableMeta(meta);
@@ -556,6 +557,10 @@ public class BaseArcticCatalog implements ArcticCatalog {
       ConvertStructUtil.TableMetaBuilder builder = ConvertStructUtil.newTableMetaBuilder(
           this.identifier, this.schema);
       String tableLocation = getTableLocationForCreate();
+
+      // default properties would not override user-defined properties
+      this.properties = CatalogUtil.mergeCatalogPropertiesToTable(this.properties, catalogMeta.getCatalogProperties());
+
       builder.withTableLocation(tableLocation)
           .withProperties(this.properties)
           .withPrimaryKeySpec(this.primaryKeySpec);
--- a/core/src/main/java/com/netease/arctic/utils/CatalogUtil.java
+++ b/core/src/main/java/com/netease/arctic/utils/CatalogUtil.java
@@ -27,6 +27,7 @@ import com.netease.arctic.io.ArcticFileIO;
 import com.netease.arctic.op.ArcticHadoopTableOperations;
 import com.netease.arctic.op.ArcticTableOperations;
 import com.netease.arctic.table.TableMetaStore;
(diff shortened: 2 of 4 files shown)


core/src/main/java/com/netease/arctic/catalog/BaseArcticCatalog.java
```diff
@@ -547,6 +548,14 @@ protected ConvertStructUtil.TableMetaBuilder createTableMataBuilder() {
       ConvertStructUtil.TableMetaBuilder builder = ConvertStructUtil.newTableMetaBuilder(
           this.identifier, this.schema);
       String tableLocation = getTableLocationForCreate();
+      try {
+        Map<String, String> catalogProperties = client.getCatalog(this.identifier.getCatalog()).getCatalogProperties();
```
zhoujinsong, line 552: Can we reuse the properties in  `catalogMeta` ?

core/src/main/java/com/netease/arctic/utils/TablePropertyUtil.java
```diff
@@ -151,4 +153,41 @@ public static long getTableWatermark(Map<String, String> properties) {
...
+  /**
+   * Merge default properties into table properties.
+   * Filter out unnecessary properties from catalog. once 'log-store-enable' or
+   * 'optimize-enable' flag is off, filter out related default properties.
+   *
+   * @param target user-defined table properties
+   * @param catalogProperties catalog properties
+   * @return merged properties
+   */
+  public static Map<String, String> mergeTableProperties(Map<String, String> target,
```
zhoujinsong, line 166: Does `CatalogUtil#mergeCatalogPropertiesToTable` method match your requirement yet?

zhoujinsong (commented):
@nicochen  Thanks a lot for your work! I left some comments, please take another look.

nicochen:
@zhoujinsong As we talked before, merge filter prefix code into `CatalogUtil` to merge properties  and  call it in `BaseArcticCatalog`

core/src/main/java/com/netease/arctic/utils/CatalogUtil.java
```diff
@@ -185,12 +186,25 @@ public static boolean isIcebergCatalog(ArcticCatalog arcticCatalog) {
...
+    Map<String, String> mergedProperties = catalogProperties.entrySet().stream()
+            .filter(e -> e.getKey().startsWith(CatalogMetaProperties.TABLE_PROPERTIES_PREFIX))
+            .collect(Collectors.toMap(e ->
+                    e.getKey().substring(CatalogMetaProperties.TABLE_PROPERTIES_PREFIX.length()), Map.Entry::getValue));
+
+    if (!PropertyUtil.propertyAsBoolean(tableProperties, TableProperties.ENABLE_LOG_STORE,
+            TableProperties.ENABLE_LOG_STORE_DEFAULT)) {
+      mergedProperties = mergedProperties.entrySet().stream()
+              .filter(e -> !e.getKey().startsWith(CatalogMetaProperties.LOG_STORE_PROPERTIES_PREFIX) &&
+                      !e.getKey().equals("log.consistency.guarantee.enable"))
```
zhoujinsong, line 198: Can we rename the property name `log.consistency.guarantee.enable` to `log-store.consistency-guarantee.enable`?  @YesOrNo828  @zstraw 
We can do this in another PR

core/src/main/java/com/netease/arctic/utils/CatalogUtil.java
```diff
@@ -185,12 +186,25 @@ public static boolean isIcebergCatalog(ArcticCatalog arcticCatalog) {
...
+              .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+    if (!Boolean.parseBoolean(PropertyUtil
```
zhoujinsong, line 201: @nicochen As I know, we should not enable Logstore in catalog default property , but we may enable or disable optimize in catalog default property, so we may consider the default value in catalog properties too ?

zhoujinsong (commented):
@nicochen please take another look !

core/src/main/java/com/netease/arctic/utils/CatalogUtil.java
zhoujinsong: #745 complete this.

zhoujinsong (approved):
LGTM.

PR #721, opened 2022-11-23
[SPARK]Use timestamp with zone in iceberg format

### LINKED ISSUE #686: [Bug][Spark]: Spark write timestamp value to parquet as UTC+0 hour.
### What happened?

When table schema contains a field as timestamp without-zone,  the value spark engines write to the parquet file is UTC+0, but should be the current timezone value.

### Affects Versions

0.4.0

Spark

### How to reproduce

checked .

select via flink 

image

select via spark 

image

table schema 

image

table files 

image

seems spark problems

### Relevant log output

_No response_

### Anything else

for sql `insert into test_db.test_table values (7, 'randy', timestamp('2022-07-03 19:11:00'));

the flink connector write data 
image

spark connector  write data 

image

### Code of Conduct

hellojinsilei:

fix #686 
Use timestamp with zone in iceberg format to solve time zone problems

## How was this patch tested?

### DIFF
--- a/site/docs/ch/docker-quickstart.md
+++ b/site/docs/ch/docker-quickstart.md
@@ -126,7 +126,7 @@ CREATE TABLE cdc_source(
 INSERT INTO arctic.test_db.test_table
 SELECT id,
        name,
-       TO_TIMESTAMP(op_time) op_time
+       CAST(TO_TIMESTAMP(op_time) AS TIMESTAMP(6) WITH LOCAL TIME ZONE) op_time
 FROM cdc_source;
 
 -- 打开表的动态配置 HINT
--- a/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
+++ b/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
@@ -204,7 +204,7 @@ public class ArcticSparkCatalog implements TableCatalog, SupportsNamespaces {
   private Schema checkAndConvertSchema(StructType schema, Map<String, String> properties) {
     Schema convertSchema;
     try {
-      CatalogMeta catalogMeta = client.getCatalog(catalogName);
+      CatalogMeta catalogMeta = client.getCatalog(catalog.name());
       String catalogType = catalogMeta.getCatalogType();
       if (CATALOG_TYPE_HIVE.equals(catalogType)) {
         SparkSession sparkSession = SparkSession.active();
@@ -220,7 +220,7 @@ public class ArcticSparkCatalog implements TableCatalog, SupportsNamespaces {
         convertSchema = SparkSchemaUtil.convert(schema, false);
       }
     } catch (TException e) {
-      throw new IllegalStateException("failed when load catalog " + catalogName, e);
+      throw new IllegalStateException("failed when load catalog " + catalog.name(), e);
     }
 
     // schema add primary keys
--- a/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
+++ b/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
@@ -23,6 +23,7 @@ import com.netease.arctic.PooledAmsClient;
 import com.netease.arctic.ams.api.CatalogMeta;
 import com.netease.arctic.catalog.ArcticCatalog;
 import com.netease.arctic.catalog.CatalogLoader;
+import com.netease.arctic.hive.utils.CatalogUtil;
 import com.netease.arctic.spark.table.ArcticSparkChangeTable;
 import com.netease.arctic.spark.table.ArcticSparkTable;
 import com.netease.arctic.table.ArcticTable;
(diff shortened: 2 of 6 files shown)


spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
```diff
@@ -206,15 +201,37 @@ public Table createTable(
     }
   }
 
-  private Schema checkAndConvertSchema(Schema schema, Map<String, String> properties) {
+  private Schema checkAndConvertSchema(StructType schema, Map<String, String> properties) {
+    Schema convertSchema;
+    try {
+      CatalogMeta catalogMeta = client.getCatalog(catalog.name());
+      String catalogType = catalogMeta.getCatalogType();
+      if (CATALOG_TYPE_HIVE.equals(catalogType)) {
```
zhoujinsong, line 209: In my opinion, the hive table use timestamp without zone only, and the iceberg table's timestamp type is dependent on the configuration which is the default with zone.

zhoujinsong (commented):
@hellojinsilei  I left some comments. please take another look.

site/docs/ch/docker-quickstart.md
```diff
@@ -126,7 +126,7 @@ CREATE TABLE cdc_source(
 INSERT INTO arctic.test_db.test_table
 SELECT id,
        name,
-       TO_TIMESTAMP(op_time) op_time
+       CAST(TO_TIMESTAMP(op_time) AS TIMESTAMP(6) WITH LOCAL TIME ZONE) op_time
```
YesOrNo828, line 129: Is this field: op_time a timestamp with the time zone in the test_table?
```
create table test_db.test_table(
  id int,
  name string,
  op_time timestamp,
  primary key(id)
) using arctic partitioned by(days(op_time));
```

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
```diff
@@ -206,15 +201,39 @@ public Table createTable(
...
 
-  private Schema checkAndConvertSchema(Schema schema, Map<String, String> properties) {
+  private Schema checkAndConvertSchema(StructType schema, Map<String, String> properties) {
+    Schema convertSchema;
+    boolean defaultTimeZoneUsed;
+    try {
+      CatalogMeta catalogMeta = client.getCatalog(catalog.name());
+      String catalogType = catalogMeta.getCatalogType();
+      SparkSession sparkSession = SparkSession.active();
+      if (CATALOG_TYPE_HIVE.equals(catalogType)) {
```
zhoujinsong, line 211: Can we determine the catalog type by the `ArcticCatalog` implementation? So we may need not the client and rpc to AMS.

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
```diff
@@ -206,15 +201,39 @@ public Table createTable(
...
+        defaultTimeZoneUsed = true;
+      } else {
+        defaultTimeZoneUsed = Boolean.parseBoolean(
+            sparkSession.conf().get(USE_TIMESTAMP_WITHOUT_TIME_ZONE_IN_NEW_TABLES,
+                "false"));
```
zhoujinsong, line 216: Maybe we can change the value in `com.netease.arctic.spark.SparkSQLProperties#USE_TIMESTAMP_WITHOUT_TIME_ZONE_IN_NEW_TABLES_DEFAULT` to `false`?

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
```diff
@@ -206,15 +201,39 @@ public Table createTable(
     }
   }
 
-  private Schema checkAndConvertSchema(Schema schema, Map<String, String> properties) {
+  private Schema checkAndConvertSchema(StructType schema, Map<String, String> properties) {
+    Schema convertSchema;
+    boolean defaultTimeZoneUsed;
```
zhoujinsong, line 206: How about rename this parameter to `useTimestampWithoutZoneInNewTables`?

zhoujinsong (changes requested):
@hellojinsilei  I left some comments. PTAL.

site/docs/ch/docker-quickstart.md
zhoujinsong: Yes, After change in this PR, spark will create timestamp with zone column in test table.

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/ArcticSparkCatalog.java
hellojinsilei: Does it mean adding a new implementation in ArcticCatalog to determine the catalog type?

hellojinsilei: Currently, there is no implementation in ArcticCatalog that can determine the type of catalog

zhoujinsong: How about adding a `CatalogUtil#isHiveCatalog` implementation in the hive module?

zhoujinsong (approved):
LGTM.

baiyangtx (approved):

PR #743, opened 2022-11-24
[Bug]: Some bugs in local test for version 0.4.0 bug

### LINKED ISSUE #756: [Bug]: Failed to get optimizerGroup tables when fail to open metadata
### What happened?

AMS cannot list tables in Optimizers dashboard when fail to open metadata file:
For instances:
- Permission denied: user=root, access=EXECUTE, inode="/path":yyy:supergroup:drwxrwxrw-
- Table metadata not find(Someone manually moved/deleted it)

AMS Optimizer dashboard should display available tables as possible as it could.

### Affects Versions

master/0.4.x

AMS, Optimizer

### How to reproduce

Manually moved/deleted the metadata and try to list tables in Optimizers dashboard

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hameizi:
1. fix #756 
2. get optimized table error because there are so many native iceberg tables in the catalog, and loading these need a long time

## Brief change log

change `refreshAndListTables` to schedule

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
@@ -64,9 +64,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.stream.Collectors;
 
 public class OptimizeService extends IJDBCService implements IOptimizeService {
@@ -78,8 +77,7 @@ public class OptimizeService extends IJDBCService implements IOptimizeService {
   private final BlockingQueue<TableOptimizeItem> toCommitTables = new ArrayBlockingQueue<>(1000);
 
   private static final long DEFAULT_CACHE_REFRESH_TIME = 60_000; // 1min
-  private final Map<TableIdentifier, TableOptimizeItem> cachedTables = new HashMap<>();
-  private final ReadWriteLock tablesLock = new ReentrantReadWriteLock();
+  private final ConcurrentHashMap<TableIdentifier, TableOptimizeItem> cachedTables = new ConcurrentHashMap<>();
 
   private long refreshTime = 0;
 
@@ -97,16 +95,14 @@ public class OptimizeService extends IJDBCService implements IOptimizeService {
   }
 
   private void init() {
-    tablesLock.writeLock().lock();
     try {
       LOG.info("OptimizeService init...");
       loadTables();
       initOptimizeTasksIntoOptimizeQueue();
+      scheduleRefresh();
       LOG.info("OptimizeService init completed");
     } catch (Exception e) {
       LOG.error("OptimizeService init failed", e);
-    } finally {
-      tablesLock.writeLock().unlock();
     }
   }
 
@@ -124,7 +120,7 @@ public class OptimizeService extends IJDBCService implements IOptimizeService {
     if (checkTasks == null) {
       return;
     }
-    List<TableIdentifier> validTables = listCachedTables(true);
+    List<TableIdentifier> validTables = listCachedTables();
     checkTasks.checkRunningTask(
         new HashSet<>(validTables),
         identifier -> checkInterval,
@@ -134,16 +130,8 @@ public class OptimizeService extends IJDBCService implements IOptimizeService {
   }
 
   @Override
-  public List<TableIdentifier> listCachedTables(boolean forceRefresh) {
-    tablesLock.readLock().lock();
-    try {
-      if (!forceRefresh && !cacheExpired()) {
-        return new ArrayList<>(cachedTables.keySet());
-      }
-    } finally {
-      tablesLock.readLock().unlock();
-    }
-    return refreshAndListTables(forceRefresh);
+  public List<TableIdentifier> listCachedTables() {
+    return new ArrayList<>(cachedTables.keySet());
   }
 
   private void clearTableCache(TableIdentifier tableIdentifier) {
@@ -167,8 +155,9 @@ public class OptimizeService extends IJDBCService implements IOptimizeService {
     }
   }
 
-  private void addTableIntoCache(TableOptimizeItem arcticTableItem, Map<String, String> properties,
-                                 boolean persistRuntime) {
+  private void addTableIntoCache(
+      TableOptimizeItem arcticTableItem, Map<String, String> properties,
+      boolean persistRuntime) {
     cachedTables.put(arcticTableItem.getTableIdentifier(), arcticTableItem);
     try {
       int queueId = ServiceContainer.getOptimizeQueueService().getQueueId(properties);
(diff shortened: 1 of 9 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
```diff
@@ -189,7 +178,7 @@ private void addTableIntoCache(TableOptimizeItem arcticTableItem, Map<String, St
   public TableOptimizeItem getTableOptimizeItem(TableIdentifier tableIdentifier) throws NoSuchObjectException {
     TableOptimizeItem tableOptimizeItem = cachedTables.get(tableIdentifier);
     if (tableOptimizeItem == null) {
-      listCachedTables(true);
+      listCachedTables();
```
wangtaohz, line 181: reload is useless here

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
```diff
@@ -509,4 +494,22 @@ private List<TableOptimizeRuntime> selectTableOptimizeRuntimes() {
       return tableOptimizeRuntimeMapper.selectTableOptimizeRuntimes();
     }
   }
+
+  private void scheduleRefresh() {
+    LOG.info("start schedule refresh");
+    new Thread(() -> {
```
wangtaohz, line 500: It's not a good practice to call `new Thread`, we'd better create the thread in `ThreadPool` like the other threads in AMS.

zhoujinsong:
As I know, the pull task request will force the planner to reload iceberg tables, do we really need to reload tables every 1 minute?

hameizi:
> As I know, the pull task request will force the planner to reload iceberg tables, do we really need to reload tables every 1 minute? @hameizi @hzluting @wangtaohz

Maybe we can lengthen the reload interval, but it is necessary to reload iceberg table information periodically, otherwise we cannot obtain the new iceberg table.

wangtaohz:
> As I know, the pull task request will force the planner to reload iceberg tables, do we really need to reload tables every 1 minute? @hameizi @hzluting @wangtaohz

The `refreshAndListTables()` is called every 1 minute, but this method does not refresh all tables, it only checks if there are new tables or removed tables, and only these new tables will be loaded.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/CatalogUtil.java
```diff
@@ -40,6 +42,8 @@
  * catalog util class。cache thrift objects
  */
 public class CatalogUtil {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CatalogUtil.class);
   public static final ConcurrentHashMap<String, ArcticCatalog> catalogCache = new ConcurrentHashMap<>();
```
wangtaohz, line 47: Nit: `catalogCache` should be `CATALOG_CACHE`

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/CatalogUtil.java
```diff
@@ -40,6 +42,8 @@
  * catalog util class。cache thrift objects
```
wangtaohz, line 42: Nit: replace `。` with `.`

zhoujinsong:
> > As I know, the pull task request will force the planner to reload iceberg tables, do we really need to reload tables every 1 minute? @hameizi @hzluting @wangtaohz
> 
> @zhoujinsong Maybe we can lengthen the reload interval, but it is necessary to reload iceberg table information periodically, otherwise we cannot obtain the new iceberg table.

Do you mean that we cannot see the new table under `Optimizier` tab?
If optimize module has never planned this table , it may be reasonable not to see it here

hameizi:
> > > As I know, the pull task request will force the planner to reload iceberg tables, do we really need to reload tables every 1 minute? @hameizi @hzluting @wangtaohz
> > 
> > 
> > @zhoujinsong Maybe we can lengthen the reload interval, but it is necessary to reload iceberg table information periodically, otherwise we cannot obtain the new iceberg table.
> 
> Do you mean that we cannot see the new table under `Optimizier` tab? If optimize module has never planned this table , it may be reasonable not to see it here

The method here is the logic of optimize

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
```diff
@@ -97,16 +91,15 @@ public OptimizeService() {
   }
 
   private void init() {
-    tablesLock.writeLock().lock();
     try {
       LOG.info("OptimizeService init...");
-      loadTables();
-      initOptimizeTasksIntoOptimizeQueue();
+      new Thread(() -> {
```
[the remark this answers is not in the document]
hameizi: These operations are time-consuming. Since this logic is executed when a new OptimizeService object is created, serial operation will affect ams startup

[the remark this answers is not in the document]
zhoujinsong: sgtm

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/CatalogUtil.java
hameizi: done

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/CatalogUtil.java
hameizi: done

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -546,25 +548,30 @@ private Set<TableIdentifier> loadTablesByQueueId(int queueId) {
...
-                tablesInQueue.add(arcticTable.id());
+        try {
+          ArcticCatalog arcticCatalog =
+              CatalogLoader.load(ServiceContainer.getTableMetastoreHandler(), catalogMeta.getCatalogName());
+          List<String> databases = arcticCatalog.listDatabases();
+          for (String database : databases) {
+            for (TableIdentifier tableIdentifier : arcticCatalog.listTables(database)) {
+              try {
+                ArcticTable arcticTable = arcticCatalog.loadTable(tableIdentifier);
+                String queueName =
```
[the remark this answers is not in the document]
hameizi: Thanks for you comment. The logic `OptimizeManagerHandler.pollTask -> OptimizeQueueService.plan` is executed to determine if `optimize.enable` is true, which is that all table information should be cached in the queue. The default `optimize.enable` is true for all tables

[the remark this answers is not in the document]
XBaith: > Thanks for you comment. The logic `OptimizeManagerHandler.pollTask -> OptimizeQueueService.plan` is executed to determine if `optimize.enable` is true, which is that all table information should be cached in the queue. The default `optimize.enable` is true for all tables

But in some situations, Iceberg optimize should be `false` by default, only register tables should be optimzed. Also `optimize.enable` default `true` for all tables is confused for users, cause the property is not enabled.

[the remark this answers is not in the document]
zhoujinsong: @XBaith you may add a property `table.optimize.enable` with value `false` for the catalog to disable the optimize processing for all tables default under the catalog. And then you can alter the table property `optimize.enable` with value `true` for a certain table to enable optimize processing for that table.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/CatalogUtil.java
```diff
@@ -98,11 +102,19 @@ public static Set<TableIdentifier> loadTablesFromCatalog() {
...
+        List<String> databases = arcticCatalog.listDatabases();
+        for (String database : databases) {
+          try {
+            tables.addAll(arcticCatalog.listTables(database));
+          } catch (Exception e1) {
+            LOG.error("list tables for database {} error", database, e1);
+          }
+        }
+      } catch (Exception e2) {
+        LOG.error("load table for catalog {} error", catalogMeta.getCatalogName(), e2);
```
wangtaohz, line 117: Catching the exception (and ignoring it) will bring risk: when failing to load the catalog/database, we will delete all the cached tables in this catalog/database, which means we will delete all the optimize `runtime` `tasks` and` history`

zhoujinsong:
Compeleted in #757.

PR #759, opened 2022-11-28
[ARCTIC-683][Flink] fixed the problem that passing a null value to some fields would be given a default value

### LINKED ISSUE #683: [Bug]: When writing data with a null value to the logstore of an arctic table, a non-null default value is assigned to the field

lklhdu:
fix #683 

*When users insert data into the log store of an arctic table via Flink, some fields of the entire data have a null value, but when the data is read from the log store these fields with a null value are given a default value, which should be fixed.*

## Brief change log

  - *Added logic for null value determination to decimal and several other field types*
  - *The Flink 1.12, 1.14, and 1.15 versions are affected.*
  - *resolve lambda serialization error*

## How was this patch tested?

- [x] Run the test locally before making a pull request


YesOrNo828 (commented):
@lklhdu , Thanks for your contribution. I think adding some tests about null values during the serialization is essential.

core/pom.xml
```diff
@@ -131,6 +131,14 @@
             <scope>test</scope>
         </dependency>
 
+
```
zstraw, line 134: don't import flink into core module

core/src/main/java/com/netease/arctic/log/LogDataJsonSerialization.java
```diff
@@ -42,7 +46,12 @@
 
   private transient LogDataToJsonConverters.LogDataToJsonConverter.FormatConverterContext converterContext;
 
-  public LogDataJsonSerialization(Schema schema, LogData.FieldGetterFactory<T> fieldGetterFactory) {
+  public LogDataJsonSerialization(Schema schema, FieldGetterFactory<T> fieldGetterFactory) {
+    this.schema = schema;
+    this.fieldGetterFactory = fieldGetterFactory;
+  }
+
+  public void init() {
```
zstraw, line 54: Please add annotation to remind devs of that the method should be invoked before serialize()

zstraw (commented):
I leave some comments, please take a look.

core/src/main/java/com/netease/arctic/log/LogDataJsonSerialization.java
zhoujinsong: Or shall we automatically initialize this variable when using it?

lklhdu: > 

One way I can think of to automatically initialize this variable is to judge if the logDataToJsonConverter variable has been initialized when the serialize() method is called and initialize the variable if it has not. I wonder if you think this solution would work

zhoujinsong: Yes, the same as I think about.

zhoujinsong (approved):
LGTM.

PR #779, opened 2022-11-29
[BUG]Hotfix Spark read change store but return the result of MOR

### LINKED ISSUE #760: [Bug]: Spark read change store but return the records in base store
### What happened?

When select records from arctic keyed table' change store by spark, it returns the records of base store.
image

### Affects Versions

master/0.4.0/

AMS, Spark

### How to reproduce

_No response_

### Relevant log output

```shell
It could be reproduced follow the quick start
```

### Anything else

_No response_

### Code of Conduct

hellojinsilei:

fix #760 
Fix Spark read change store but return the result of MOR

## How was this patch tested?

### DIFF
--- a/spark/v3.1/spark/src/test/java/com/netease/arctic/spark/TestKeyedTableDML.java
+++ b/spark/v3.1/spark/src/test/java/com/netease/arctic/spark/TestKeyedTableDML.java
@@ -108,6 +108,7 @@ public class TestKeyedTableDML extends SparkTestBase {
         newRecord(keyedTable, 4, "ddd", quickDateWithZone(4)),
         newRecord(keyedTable, 5, "eee", quickDateWithZone(4))
     ));
+    writeBase(identifier,  baseFiles);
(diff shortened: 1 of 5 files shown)


spark/v3.1/spark/src/test/java/com/netease/arctic/spark/TestKeyedTableDML.java
```diff
@@ -108,6 +108,7 @@ public void testSelectChangeFiles() {
         newRecord(keyedTable, 4, "ddd", quickDateWithZone(4)),
         newRecord(keyedTable, 5, "eee", quickDateWithZone(4))
     ));
+    writeBase(identifier,  baseFiles);
```
baiyangtx, line 111: Add more test data.  The change table needs both insert data and delete data.  delete data must contain rows in the base table.

baiyangtx (changes requested):

baiyangtx (approved):
LGTM

PR #790, opened 2022-11-30
[BUG][SPARK]Catalog cache meta cause AMS catalog modify not effect at terminal sql

### LINKED ISSUE #786: [Bug][Spark]: Catalog cache meta cause AMS catalog modify not effect at terminal sql
### What happened?

modify catalog properties  `warehouse.dir`,  and execute sql on the terminal in local mode, the table will be created at the old location.

solution:

1. Core:    Catalog interface add a method to refresh the catalog meta
2. Spark:   Add a property of the catalog to control the catalog meta refresh action.  (e.g  force-refresh,   default is false)
3. Spark:   Call catalog.refresh() before any action if force-refresh is enabled.
4. AMS:    Terminal in the local mode set the property to true when creating the spark session.

### Affects Versions

0.4.0

Spark

### How to reproduce

modify catalog properties  `warehouse.dir`,  and execute sql on the terminal in local mode,  table will created at old location.

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hellojinsilei:

 fix #786 

## How was this patch tested?

### DIFF
--- a/ams/ams-api/src/main/java/com/netease/arctic/ams/api/properties/CatalogMetaProperties.java
+++ b/ams/ams-api/src/main/java/com/netease/arctic/ams/api/properties/CatalogMetaProperties.java
@@ -63,4 +63,8 @@ public class CatalogMetaProperties {
 
     // properties in table level set by catalog
     public static final String TABLE_PROPERTIES_PREFIX = "table.";
+
+    public static final String FORCE_REFRESH = "force.refresh";
+
+    public static final String FORCE_REFRESH_DEFAULT = "false";
 }
--- a/ams/ams-server/src/test/java/com/netease/arctic/ams/server/AmsTestBase.java
+++ b/ams/ams-server/src/test/java/com/netease/arctic/ams/server/AmsTestBase.java
@@ -48,6 +48,7 @@ import com.netease.arctic.ams.server.service.TestArcticTransactionService;
 import com.netease.arctic.ams.server.service.TestDDLTracerService;
 import com.netease.arctic.ams.server.service.TestFileInfoCacheService;
 import com.netease.arctic.ams.server.service.TestOptimizerService;
+import com.netease.arctic.ams.server.service.TestRefreshCatalog;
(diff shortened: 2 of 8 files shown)


core/src/test/java/com/netease/arctic/catalog/TestRefreshCatalog.java
```diff
@@ -0,0 +1,21 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import static com.netease.arctic.ams.api.MockArcticMetastoreServer.TEST_CATALOG_NAME;
+
+public class TestRefreshCatalog extends TableTestBase {
+
+  @Test
+  public void refreshCatalog() throws TException {
+    testCatalog = CatalogLoader.load(AMS.getUrl());
```
zhoujinsong, line 15: Do you need change some catalog configuration before refresh to validate this feature?

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
```diff
@@ -27,4 +27,8 @@ public class SparkSQLProperties {
 
   public static final String USE_TIMESTAMP_WITHOUT_TIME_ZONE_IN_NEW_TABLES_DEFAULT =
           "false";
+
+  public static final String FORCE_REFRESH = "force.refresh";
```
zhoujinsong, line 31: Should we start the property name with `spark.sql.arctic`?
How about rename the property name to `spark.sql.arctic.refresh-catalog-before-usage`?

zhoujinsong (commented):
@hellojinsilei I left some comments.

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
zhoujinsong: BTW, You may need to set this property to `true` in `LocalTerminalSession` in AMS.

Issue #797, opened 2022-12-01
[Improvement]: Validate required properties when creating catalog


zhoujinsong:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Arctic support creating catalogs on AMS Dashboard now. We need check the requested properties when creating catalogs.

### How should we improve?

Different catalog types need different required properties:

- Arctic Metastore requires `warehouse.dir`
- Hadoop requires `warehouse`
- Custom requires `catalog-impl`

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

YesOrNo828:
Could we unify the `warehouse.dir` and `warehouse` property names in the Arctic Metasotre and Hadoop catalog types?

zhoujinsong:
> Could we unify the `warehouse.dir` and `warehouse` property names in the Arctic Metasotre and Hadoop catalog types?

Unify the property name looks good to me, but we need to handle the compatibility problems with existed catalog.
How do you think ? @baiyangtx

PR #802, opened 2022-12-01
[ARCTIC-797]Validate catalog properties when creating

zhoujinsong:

Resove #797

## Brief change log

  - *Add property validation for Arctic Metastore and Hadoop Catalog*
  - *Change catalog property name `warehouse.dir` to `warehouse`*

## How was this patch tested?

## Documentation


Aireed (changes requested):
file sql/derby/ams-init.sql  have insert sql for add localcatalog, we should change it !
same as sql/mysql/0.4.0-init.sql

core/src/main/java/com/netease/arctic/catalog/BaseArcticCatalog.java
```diff
@@ -659,16 +659,21 @@ protected void fillTableProperties(TableMeta meta) {
...
+        String catalogWarehouse = catalogMeta.getCatalogProperties().getOrDefault(
+            CatalogMetaProperties.KEY_WAREHOUSE,null);
+        if (catalogWarehouse == null) {
+          catalogWarehouse = catalogMeta.getCatalogProperties().getOrDefault(
+              CatalogMetaProperties.KEY_WAREHOUSE_DIR,null);
         }
-        if (StringUtils.isNotBlank(catalogWarehouseDir)) {
-          return catalogWarehouseDir + '/' + identifier.getDatabase();
+        if (catalogWarehouse == null) {
+          throw new NullPointerException("Catalog warehouse is null.");
```
baiyangtx, line 669: IllegaStateExecption maybe better ?

zhoujinsong: I prefer to use NullPointerException when the required parameter is null.

Aireed (approved):

PR #828, opened 2022-12-05
[ISSUE-827] Fix the failed UTs in win.

### LINKED ISSUE #827: [Bug]: Some UTs would be failed in win.
### What happened?

Some UTs would be failed in win because of the wrong path separator.

### Affects Versions

master

_No response_

### How to reproduce

mvn clean package -pl "!trino"    in windows.

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

zstraw:
Fix #827 

## Brief change log
Fix path separators in UTs.

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/test/java/com/netease/arctic/io/IcebergFanoutPosDeleteWriterTest.java
+++ b/core/src/test/java/com/netease/arctic/io/IcebergFanoutPosDeleteWriterTest.java
@@ -18,10 +18,10 @@
 
 package com.netease.arctic.io;
 
-import com.google.common.collect.Maps;
 import com.netease.arctic.IcebergTableBase;
 import com.netease.arctic.io.writer.IcebergFanoutPosDeleteWriter;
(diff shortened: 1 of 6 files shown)


zstraw:
> Do we need to construct Path Object `new Path(FileUtil.getCombinedPath(dataDir, "data-1.parquet")).toString()` or just `FileUtil.getCombinedPath(dataDir, "data-1.parquet")` is okay?

If we just FileUtil.getCombinedPath(dataDir, "data-1.parquet"), the path returned is separated by '\' in win or '/' in linux. It will cause other Exception due to the path parsed by hard code '/' in that logic. And that method is called by other place which depend on Hadoop Path which is returned in the separator '/' in all systems.

zhoujinsong (approved):
LGTM.

zstraw:
Set flink UT parallelism to 1 to avoid blocking temporarily.

PR #836, opened 2022-12-05
[Docs]: Improvement of quick start

baiyangtx:

## Brief change log

- Add texts of warehouse to make user easier to copy from docs.
- Fix: ams container start failed for lack ams-info.log

## How was this patch tested?
-  Add some test cases that check the changes thoroughly including negative and positive cases if possible

## Documentation


zhoujinsong (approved):
LGTM.

PR #879, opened 2022-12-07
[ISSUE-561][Flink] Fix flink cannot fetch right ckpId after failover

### LINKED ISSUE #561: [Bug]: When flink job fails over, flink job will commit a wrong transaction id

### LINKED ISSUE #826: [Bug]: When a table has too many small files to be optimized, it will cause AMS OOM
### What happened?

image
When there are too many small files to be optimized with a primary key table, and then optimize causes AMS OOM
image
We can see that optimizing a small file of over 600,000 in an AMS configuration with 32G of RAM results in oom

### Affects Versions

0.3.x

AMS, Optimizer

### How to reproduce

1. Create table and write a lot of data for a long time without optimize
2. Set props ‘optimize.enable’ = true 

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

zstraw:
Fix #561 Fix #826 

the fix #568 about Flink 1.14, 1.15 was wrong due to the op cannot fetch right restore ckpId after failover if op has no state.

And the ckpId assigned to writer is not one to one correspondence, which leads to the repeatable file name.

## Brief change log

  - *The flink 1.12, 1.14 and 1.15 versions are affected.*

## How was this patch tested?

## Documentation

### DIFF
--- a/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
+++ b/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
@@ -71,7 +71,7 @@ public class ArcticFileWriter extends AbstractStreamOperator<WriteResult>
   private transient int subTaskId;
   private transient int attemptId;
   private transient String jobId;
-  private transient long checkpointId = 0;
+  private transient long checkpointId = 1;
   private transient ListState<Long> checkpointState;
   /**
    * Load table in runtime, because that table's refresh method will be invoked in serialization.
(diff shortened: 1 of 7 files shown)


zhoujinsong (commented):
Could you add some unit test case to validate this fix?

PR #892, opened 2022-12-08
[ARCTIC-890][AMS] Fix the data loss when the minor optimize in progress and restart the ams at the same time

### LINKED ISSUE #890: [Bug][ams]: Data will lose when the minor optimize in progress and restart the ams at the same time
### What happened?

For the arctic keyed table, when the minor optimize in progress and restart the ams simultaneously.
If ams started before the minor optimize is complete, the minor optimize commit will lose the insert files in the task.
Because for the minor optimize, we should add the insert files into target files. But if ams restart, the insert files cache in optimize task will be clean, so the insert files can't add to target files, which causes the optimize commit will lose the insert files.

### Affects Versions

0.3.0

_No response_

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hzluting:
fix #890 

## Brief change log

  - *Load the insert files cache in optimize task when adding the insert files into the target files*

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
@@ -317,6 +317,7 @@ public class TableOptimizeItem extends IJDBCService {
         // if minor optimize, insert files as base new files
         if (optimizeTaskItem.getOptimizeTask().getTaskId().getType() == OptimizeType.Minor &&
             !com.netease.arctic.utils.TableTypeUtil.isIcebergTableFormat(getArcticTable())) {
+          optimizeTaskItem.setFiles();
           targetFiles.addAll(optimizeTaskItem.getOptimizeTask().getInsertFiles());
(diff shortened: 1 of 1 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
```diff
@@ -317,6 +317,10 @@ public void updateOptimizeTaskStat(OptimizeTaskStat optimizeTaskStat) {
         // if minor optimize, insert files as base new files
         if (optimizeTaskItem.getOptimizeTask().getTaskId().getType() == OptimizeType.Minor &&
             !com.netease.arctic.utils.TableTypeUtil.isIcebergTableFormat(getArcticTable())) {
+          if (optimizeTaskItem.getOptimizeTask().getInsertFileCnt() !=
+              optimizeTaskItem.getOptimizeTask().getInsertFiles().size()) {
+            optimizeTaskItem.setFiles();
```
wangtaohz, line 322: We have to check the result of set Files(check insertFileCnt)

zhoujinsong (approved):

PR #908, opened 2022-12-09
[ARCTIC-907][SPARK]Duplicate keys verification for the source table should add a property to control in Spark

### LINKED ISSUE #907: [Improvement]: Duplicate keys verification for the source table should add a property to control in Spark
### Search before asking

In Spark SQL, e.g. insert into, insert overwrite, CTAS. We will check whether the primary key is not duplicated in the source table.
I think duplicate keys verification for the source table should add a property to control

### How should we improve?

Add a spark property, e.g. spark.sql.check-data-duplicates.enabled 
the default value is true

### Subtasks

_No response_

### Code of Conduct

hellojinsilei:

resolve #907 

## Brief change log

Add a spark property: spark.sql.check-data-duplicates.enabled, the default value is true 

## How was this patch tested?

## Documentation

### DIFF
--- a/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
+++ b/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
@@ -31,4 +31,8 @@ public class SparkSQLProperties {
   public static final String REFRESH_CATALOG_BEFORE_USAGE = "spark.sql.arctic.refresh-catalog-before-usage";
 
   public static final String REFRESH_CATALOG_BEFORE_USAGE_DEFAULT = "false";
+
+  public static final String CHECK_DATA_DUPLICATES_ENABLE = "spark.sql.check-data-duplicates.enabled";
+
+  public static final String CHECK_DATA_DUPLICATES_ENABLE_DEFAULT = "true";
 }
--- a/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
+++ b/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
@@ -20,7 +20,7 @@ package com.netease.arctic.spark;
 
 public class SparkSQLProperties {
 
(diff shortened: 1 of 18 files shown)


spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
```diff
@@ -31,4 +31,8 @@ public class SparkSQLProperties {
   public static final String REFRESH_CATALOG_BEFORE_USAGE = "spark.sql.arctic.refresh-catalog-before-usage";
 
   public static final String REFRESH_CATALOG_BEFORE_USAGE_DEFAULT = "false";
+
+  public static final String CHECK_DATA_DUPLICATES_ENABLE = "spark.sql.check-data-duplicates.enabled";
```
zhoujinsong, line 35: Does `spark.sql.arctic.check-source-data-uniqueness.enabled` property name is more appropriate？

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
```diff
@@ -31,4 +31,8 @@ public class SparkSQLProperties {
...
+
+  public static final String CHECK_DATA_DUPLICATES_ENABLE_DEFAULT = "true";
```
zhoujinsong, line 37: Should we change default value to `false`,  considering the big cost of this operation?

zhoujinsong (commented):
@hellojinsilei  I left some comments.

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/SparkSQLProperties.java
hellojinsilei: If we want to make the primary key unique by default, does it means this parameter should also be enabled by default?

zhoujinsong:
You may need to add this new configuration to the user documentation. @hellojinsilei

zhoujinsong (approved):
LGTM.

PR #921, opened 2022-12-12
[ARCTIC-920][Optimize] Fix the problem of data inconsistency after minor optimize

### LINKED ISSUE #920: [Bug][Optimize]:If the interval between the major optimize commit and the next minor optimize plan too short will cause data inconsistency
### What happened?

For the arctic keyed table and using the upsert mode, there are two records with the same primary key after the minor optimize.
When we do the minor optimize plan, we get the current snapshot to check whether the base table cache is all in the file cache. Still, we don't force refresh the base table, so maybe we get the old snapshot rather than the latest snapshot if the interval between this minor optimize plan and the last major optimize commit is too short(less than the time arctic table to refresh in cache). If the file cache doesn't have the latest snapshots and files, the check will pass, and we will get the old data files in the base table to do the minor optimize. 
The result is the eq-delete files will apply to the old data files rather than the latest data files. The generated pos-delete files are useless, and the deleted data will lose.

### Affects Versions

0.3.0

_No response_

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hzluting:
fix #920 

## Brief change log

  - *force refresh the base table when checking the file cache in the base table*
  - *remove isRunning in TableOptimizeRuntime and use optimize task size to check whether the table is running*
  - *generate partition running status before getting data files to avoid incorrect plan*

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BaseArcticOptimizePlan.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BaseArcticOptimizePlan.java
@@ -186,11 +186,14 @@ public abstract class BaseArcticOptimizePlan extends BaseOptimizePlan {
     if (arcticTable.isKeyedTable()) {
       long snapshotId = UnKeyedTableUtil.getSnapshotId(arcticTable.asKeyedTable().baseTable());
       this.currentBaseSnapshotId = snapshotId;
+      this.currentChangeSnapshotId = UnKeyedTableUtil.getSnapshotId(arcticTable.asKeyedTable().changeTable());
       if (snapshotId != TableOptimizeRuntime.INVALID_SNAPSHOT_ID && !snapshotIsCached.test(snapshotId)) {
         LOG.debug("File cache don't have cache snapshotId:{}," +
             "wait file cache sync latest file info", snapshotId);
         return false;
       }
+    } else {
+      this.currentBaseSnapshotId = UnKeyedTableUtil.getSnapshotId(arcticTable.asUnkeyedTable());
     }
 
     return true;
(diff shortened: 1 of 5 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BaseArcticOptimizePlan.java
```diff
@@ -183,12 +183,11 @@ protected BaseOptimizeTask buildOptimizeTask(@Nullable List<DataTreeNode> source
   }
 
   public boolean baseTableCacheAll() {
-    Snapshot snapshot;
     if (arcticTable.isKeyedTable()) {
-      snapshot = arcticTable.asKeyedTable().baseTable().currentSnapshot();
-      if (snapshot != null && !snapshotIsCached.test(snapshot.snapshotId())) {
+      long snapshotId = UnKeyedTableUtil.getSnapshotId(arcticTable.asKeyedTable().baseTable());
```
wangtaohz, line 187: The `baseTable` is refreshed twice in `tableNeedPlan()` and `baseTableCacheAll()`

hzluting: OK, I move it to baseTableCacheAll()

zhoujinsong: How about refreshing both `BaseStore` and `ChangeStore` first in `tableNeedPlan` method,  
This will make the code easier to understand.

hzluting: OK

ams/ams-server/src/test/java/com/netease/arctic/ams/server/OptimizeIntegrationTest.java
```diff
@@ -118,6 +118,9 @@ public static void beforeClass() {
     amsEnvironment.createIcebergCatalog(ICEBERG_CATALOG, ICEBERG_CATALOG_DIR);
     catalog(CATALOG).createDatabase(DATABASE);
     catalog(ICEBERG_CATALOG).createDatabase(DATABASE);
+
+    // wait for ams init service
+    Thread.sleep(1000);
```
wangtaohz, line 123: It's not guaranteed to finish initiation after waiting for 1s.

wangtaohz (approved):

wangtaohz:
LGTM

zhoujinsong (approved):
LGTM.

Issue #860, opened 2022-12-06
[Improvement]: Reduce plan memory


shidayang:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

It will used 200M memory when we plan a table with 8700 files. 
I think the 200M is too big

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wangtaohz:
What are the table type and file type?  Is it an Arctic Keyed Table?

shidayang:
@wangtaohz Keyed table

PR #929, opened 2022-12-13
[ARCTIC-860] Improve  memory usage when plan table

shidayang:
fix: #860 

## Brief change log
BaseUnkeyedTable get TableScan without includeColumnStats

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/data/DefaultKeyedFile.java
+++ b/core/src/main/java/com/netease/arctic/data/DefaultKeyedFile.java
@@ -41,8 +41,6 @@ public class DefaultKeyedFile implements PrimaryKeyedFile, Serializable {
   private final DataFile internalFile;
 
   private transient FileMeta meta;
-  private transient ChangedLsn minLsn;
-  private transient ChangedLsn maxLsn;
 
   public DefaultKeyedFile(DataFile internalFile) {
     this.internalFile = internalFile;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/netease/arctic/data/DefaultKeyedFile.java
```diff
@@ -78,18 +78,12 @@ public Long transactionId() {
 
   @Override
   public ChangedLsn minLsn() {
```
zhoujinsong, line 80: Maybe we can remove this method from the interface if it is not needed for now.

zhoujinsong (commented):
@shidayang I left some comments. BTW, I am wondering how much memory can we sav if we skip the file stat when reading the table.

zhoujinsong:
@shidayang Can you rename the PR title to match community requirements？

core/src/main/java/com/netease/arctic/table/BaseUnkeyedTable.java
```diff
@@ -101,7 +101,7 @@ public void refresh() {
 
   @Override
   public TableScan newScan() {
-    return icebergTable.newScan().includeColumnStats();
+    return icebergTable.newScan();
```
wangtaohz, line 104: Does Iceberg's TableScan not `includeColumnStats` as default?

shidayang: yes

shidayang:
@zhoujinsong From 190_989_776 to 16_682_384 in my case with about 8000+ files

zhoujinsong:
> @zhoujinsong From 190_989_776 to 16_682_384 in my case with about 8000+ files

Cool! That really is a big improvement!

zhoujinsong (approved):
LGTM.

Issue #955, opened 2022-12-20
[Bug]: Optimize status hangs in pending and minor


XBaith:
### What happened?

Hi, Optimize Iceberg has encountered two problems, and I haven't found the problem for a long time:
1. optimize group 1: The optimizer is still alive but the OptimizeQueue has no plan log, which means the queue no longer works? (It will no longer work after merging 2-3 tables successfully)

3. optimize group 2: optimize does not respond for a long time after the execution of minor, although its file size is very small. (Some tables succeed, some tables hang on or timeout retry)

And I'm not sure the bug ocurred by my change which I drop all status in `BaseFileScan` to reduce jvm heap memory(BTW, It does save a lot of memory. If this has no impact on optimization, I will trigger a new PR for this.). 

### Affects Versions

master/0.4.x

### What engines are you seeing the problem on?

AMS

### How to reproduce

just optimize a certain number of iceberg tables

### Anything else

![image](https://user-images.githubusercontent.com/54210725/208618448-5f34d7fb-a34b-4604-a6b9-2788d3bf619e.png)
![image](https://user-images.githubusercontent.com/54210725/208618547-90442eab-9110-4df8-b6c4-088e3cb2097a.png)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

majin1102:
cc @wangtaohz @hzluting

hzluting:
Hello, thanks for your feedback.
For the first problem, if you can confirm the optimizer Flink task is alive, and the plan log of this optimize queue doesn't output(maybe there are some error message in ams-error.log). You can use jstack for the AMS and search a thread named "Optimize Plan Thread Queue-xx(xx is the optimize queue id)",  maybe this thread blocked by some operations cost much time, such as the hive sync in table refresh.
For the second problem, the optimize status of the table isn't accurate. Once the table has optimize task and the optimize status is optimizing(minor or major). And we will update the logic of optimize status :
1. optimizing(minor or major): Only the table has the optimize task in executing
2. pending: The table has no executing optimize task but has pending or init optimize task, or the table hasn't optimize task now but needs to optimize
3. idle: The table has no executing and pending optimize task

majin1102:
[...]

PR #962, opened 2022-12-22
[ARCTIC-955] Improve performance of Table entries scan

wangtaohz:

As described in the #955, when process table entries scan for getting file sequence, we call `table.spec()` many times which invokes `ugi.doAs()` each time.

And also, there is no need to load column metrics if not necessary.

## Brief change log

- cache table spec for table entries scan
- only include colum stats when necessary

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/scan/TableEntriesScan.java
+++ b/core/src/main/java/com/netease/arctic/scan/TableEntriesScan.java
@@ -32,6 +32,7 @@ import org.apache.iceberg.HasTableOperations;
 import org.apache.iceberg.MetadataTableType;
 import org.apache.iceberg.MetadataTableUtils;
 import org.apache.iceberg.Metrics;
+import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.StructLike;
 import org.apache.iceberg.Table;
@@ -61,9 +62,11 @@ public class TableEntriesScan {
   private final Expression dataFilter;
   private final boolean aliveEntry;
   private final boolean allFileContent;
+  private final boolean includeColumnStats;
   private final Set<FileContent> validFileContent;
 
   private Table entriesTable;
+  private PartitionSpec spec;
   private InclusiveMetricsEvaluator lazyMetricsEvaluator = null;
   private Map<String, Integer> lazyIndexOfDataFileType;
   private Map<String, Integer> lazyIndexOfEntryType;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netease/arctic/scan/TableEntriesScan.java
```diff
@@ -61,9 +62,11 @@ public class TableEntriesScan {
   private final Expression dataFilter;
   private final boolean aliveEntry;
   private final boolean allFileContent;
+  private final boolean includeColumnStats;
   private final Set<FileContent> validFileContent;
 
   private Table entriesTable;
+  private PartitionSpec spec;
```
XBaith, line 69: By using `final` to describe the spec and initializing it in this constructor, we can avoid calling the function. 
What do you think is it better to initialize immutable variables immediately or lazily load?

wangtaohz: Honestly, they do not have much difference in this situation, but I have a preference to keep the constructor simple.

XBaith (approved):

zhoujinsong:
@wangtaohz Hi, Thanks a lot for fixing this problem.

But I am thinking `ArcticTable#spec` should not be an expensive method.
So we may should remove the `doAs` code from `ArcticTableOperations#current`.

wangtaohz:
> But I am thinking ArcticTable#spec should not be an expensive method.
So we may should remove the doAs code from ArcticTableOperations#current.

I agree. We only have to use `doAs` when `ArcticTableOperations#refresh` is invoked.

I will remove these code.

core/src/main/java/com/netease/arctic/scan/TableEntriesScan.java
```diff
@@ -162,7 +180,10 @@ public CloseableIterable<IcebergFileEntry> entries() {
           if (shouldKeep(status, fileContent)) {
             ContentFile<?> contentFile = buildContentFile(fileContent, fileRecord);
             if (metricsEvaluator().eval(contentFile)) {
-              return new IcebergFileEntry(snapshotId, sequence, contentFile);
+              ContentFile<?> copyFile = includeColumnStats ?
```
zhoujinsong, line 183: Why do need to copy the `contentFile` here?

wangtaohz: Before copy, the `contentFile` may contain column stats even if `includeColumnStats == false`, because of `dataFilter != null`.

```
private boolean needMetrics() {
    return dataFilter != null || includeColumnStats;
  }
```

So we call `contentFile.copyWithoutStats()` to remove these column stats.

However, it's not necessary to copy it in other cases.

wangtaohz: I modified the code, you may take a look.

zhoujinsong (approved):

Issue #916, opened 2022-12-12
[Improvement]: Too many Full GC after starting AMS


XBaith:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

I think too many resident objects are saved in the old generation, which leads to frequent full gc, especially when AMS is just started

![image](https://user-images.githubusercontent.com/54210725/206969226-8181fc90-7dfd-42f7-8a0a-46d79d55232b.png)

- AMS run with 10G heap memory in docker
- Handle 1253 tables
```
VM Flags:
Non-default VM flags: -XX:CICompilerCount=4 -XX:GCLogFileSize=10485760 -XX:InitialHeapSize=2147483648 -XX:+ManagementServer -XX:MaxHeapSize=10737418240 -XX:MaxNewSize=5368709120 -XX:MinHeapDeltaBytes=524288 -XX:NewRatio=1 -XX:NewSize=1073741824 -XX:NumberOfGCLogFiles=10 -XX:OldSize=1073741824 -XX:+PrintGC -XX:+PrintGCDetails -XX:+PrintGCTimeStamps -XX:SurvivorRatio=3 -XX:+UseCompressedClassPointers -XX:+UseCompressedOops -XX:+UseFastUnorderedTimeStamps -XX:+UseGCLogFileRotation -XX:+UseParallelGC 
Command line:  -Dlog.home=/usr/local/ams/arctic-0.4.0-SNAPSHOT/logs -Dlog.dir=/usr/local/ams/arctic-0.4.0-SNAPSHOT/logs -Duser.dir=/usr/local/ams/arctic-0.4.0-SNAPSHOT -Xloggc:/usr/local/ams/arctic-0.4.0-SNAPSHOT/logs/gc.log -XX:+UseGCLogFileRotation -XX:NumberOfGCLogFiles=10 -XX:GCLogFileSize=10M -Xms2048m -Xmx10240m -XX:MaxPermSize=128m -verbose:gc -XX:+PrintGCDetails -Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false -XX:NewRatio=1 -XX:SurvivorRatio=3
```
Improve AMS corresponding efficiency, reduce full gc times.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

[...]

PR #963, opened 2022-12-23
[ARCTIC-955][ARCTIC-916] Load sequence number once in optimize plan

### LINKED ISSUE #955: [Bug]: Optimize status hangs in pending and minor

XBaith:

Avoid scanning holl tables in each partition to reduce heap memory. 
fix #955 issue 1 and decrease full gc times (mentioned in #916  )

## Brief change log

In some cases, tables have more than a hundred thousand partitions (12 months and 16 bucket sizes), and scanning the table per partition will be very expensive. Actually we only need to load the current file and sequence number mapping in one optimized plan.

## How was this patch tested?

## Documentation


XBaith:
cc @wangtaohz @hzluting

wangtaohz:
Thanks for your work! 
And I think it will bring a significant performance improvement when planning with many partitions.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
```diff
@@ -440,7 +441,10 @@ public void updateTableOptimizeStatus() {
               getIcebergMinorPlan(fileScanTasks, -1, System.currentTimeMillis(), partitionIsRunning);
           List<BaseOptimizeTask> minorTasks = minorPlan.plan();
           // pending for minor optimize
-          if (CollectionUtils.isNotEmpty(minorTasks)) {
+          if (CollectionUtils.isNotEmpty(minorTasks) && TableOptimizeStatusUtil.in(
```
wangtaohz, line 444: It may be related to this issue https://github.com/NetEase/arctic/issues/959, we'd better discuss it under this issue, and then put a new PR to fix it.

wangtaohz (approved):

PR #976, opened 2023-01-04
[ARCTIC-849] Filestore reader emits keyed table data without shuffle and keeps data ordering

### LINKED ISSUE #849: [Improvement][Flink]: Filestore reader emits keyed table data without shuffle and keeps data ordering
### Search before asking

For now, Arctic file reader read Keyed table with shuffle mandatorily.
Actually, it's unnecessary. There should be another design to deal with the problem that read Keyed table unordered when nodes scale in/out.

### How should we improve?

_No response_

### Subtasks

_No response_

### Code of Conduct

YesOrNo828:
fix #849 

Keep data ordering when the ArcticSource assigns the multi-splits with different layers in the arctic tree node.
1. Arctic source read `mask=3` layer nodes.
2. Arctic enumerator finds `mask=1` layer data files are committed in the next snapshot, these files should split into the `mask=3` layer and put into the  `ShuffleSplitAssigner`.

## Brief change log

  - *Support data files move up layers or go down layers in the arctic tree nodes.*
  - *Add more UTs*
  - *Affect the Flink 1.12\1.14\1.15 versions*

## How was this patch tested?

## Documentation

  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) not documented

### DIFF
--- a/core/src/main/java/com/netease/arctic/utils/SchemaUtil.java
+++ b/core/src/main/java/com/netease/arctic/utils/SchemaUtil.java
@@ -20,7 +20,12 @@ package com.netease.arctic.utils;
 
 import com.netease.arctic.table.MetadataColumns;
 import org.apache.iceberg.Schema;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.types.TypeUtil;
+import org.apache.iceberg.types.Types;
+
+import java.util.List;
+import java.util.Set;
 
 public class SchemaUtil {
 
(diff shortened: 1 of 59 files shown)


flink/v1.12/flink/src/main/java/com/netease/arctic/flink/read/hybrid/assigner/ShuffleSplitAssigner.java
```diff
@@ -195,14 +196,8 @@ void deserializePartitionIndex(long[] shuffleSplitRelation) {
     }
   }
 
-  private long partitionAndIndexHashCode(String partition, ArcticSplit arcticSplit) {
-    return Math.abs(partition.hashCode() + getExactlyIndexOfTreeNode(arcticSplit));
-  }
-
-  @VisibleForTesting
-  public long getExactlyIndexOfTreeNode(ArcticSplit arcticSplit) {
+  public List<DataTreeNode> getExactlyTreeNodes(ArcticSplit arcticSplit) {
```
zstraw, line 199: Can you annotate for the case that a split belongs to multi tree nodes.

zstraw (commented):
Would it be better to disable read shuffle in default?

YesOrNo828:
> 

Yes, I disable the `read.distribution-mode` for the default.

zhoujinsong (commented):
I am wondering after this improvement, do we still need this configuration `read.distribution-mode` to support adding a shuffle operator after reading data?

zhoujinsong:
@YesOrNo828  After I review the source split code, I find a problem:
Do we need to make sure all the files inside a `ChangelogSplit` belong to the same data tree node?
If yes, I am afraid the `PartitionAndNodeGroup#planSplits` cannot make this happen.
I noticed that we group files by the mask here, but files with the same mask may belong to different data tree node.
We may should group files by `DataTreeNode#getId`.

YesOrNo828:
> @YesOrNo828 After I review the source split code, I find a problem: Do we need to make sure all the files inside a `ChangelogSplit` belong to the same data tree node? If yes, I am afraid the `PartitionAndNodeGroup#planSplits` cannot make this happen. I noticed that we group files by the file mask here, but files with the same mask may belong to different data tree nodes. We may group files by `DataTreeNode#getId` instead.

@zhoujinsong  Thanks for you reviewing. I've divided data files into groups by node.getId() 

And Removed the logic of the `READ_DISTRIBUTION_MODE` in the `ArcticDynamicSource`

core/src/main/java/com/netease/arctic/utils/SchemaUtil.java
```diff
@@ -30,4 +35,30 @@ public static Schema changeWriteSchema(Schema changeTableSchema) {
     );
     return TypeUtil.join(changeTableSchema, changeWriteMetaColumnsSchema);
   }
+
+  /**
+   * Convert an Iceberg Schema {@link Schema} to a {@link Schema} based on the given schema.
+   * <p>
+   * This fill-up does not assign new ids; it uses ids from the base schema.
+   * <p>
+   * If the toSchema does contain the identifierFields of the based schema, it will fill-up the identifierFields to
```
zstraw, line 44: Which one is toSchema?

zstraw (approved):
LGTM

zhoujinsong (approved):
LGTM.

Issue #979, opened 2023-01-05
[Improvement][Core]: The UpdateHiveFiles commit can't ensure atomicity for the partitions that need to alter


hzluting:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In the UpdateHiveFiles commit, we alter the partition by deleting and adding the same. We should improve it because one operation becomes two operations that can't ensure atomicity. And will have an intermediate state if the deleting was successful and the add failed.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #980, opened 2023-01-05
[ARCTIC-979][Core]Use alter to modify partition location rather than delete and add the partitions in UpdateHiveFiles commit

hzluting:
improve #979 

## Brief change log

  - *use alter to modify partition location rather than delete and add the partitions in UpdateHiveFiles commit*

## How was this patch tested?

## Documentation

### DIFF
--- a/hive/src/main/java/com/netease/arctic/hive/op/UpdateHiveFiles.java
+++ b/hive/src/main/java/com/netease/arctic/hive/op/UpdateHiveFiles.java
@@ -321,7 +321,9 @@ public abstract class UpdateHiveFiles<T extends SnapshotUpdate<T>> implements Sn
               "partition will be " +
               "delete and re-create with same location");
         } else {
-          partitions.put(entry.getKey(), entry.getValue());
+          // this partition is need to alter, rather than delete
+          partitionToAlter.put(entry.getKey(), entry.getValue());
+          partitionToDelete.remove(entry.getKey());
           continue;
         }
       }


zhoujinsong (approved):
LGTM.

@baiyangtx do you have time to review this PR?

hive/src/main/java/com/netease/arctic/hive/op/UpdateHiveFiles.java
```diff
@@ -321,7 +321,9 @@ private Map<StructLike, Partition> filterNewPartitionNonExists(
               "partition will be " +
               "delete and re-create with same location");
         } else {
-          partitions.put(entry.getKey(), entry.getValue());
+          // this partition is need to alter, rather than delete
+          partitionToAlter.put(entry.getKey(), entry.getValue());
+          partitionToDelete.remove(entry.getKey());
```
baiyangtx, line 326: need to update hive partition fields and properties.
includes lastAccessTime, transient_lastDdlTime, totalSize, numRows, numFiles.

baiyangtx (changes requested):

hive/src/main/java/com/netease/arctic/hive/op/UpdateHiveFiles.java
hzluting: > need to update hive partition fields and properties. includes lastAccessTime, transient_lastDdlTime, totalSize, numRows, numFiles.

Yes, the partition is already produced by the latest info.

zhoujinsong: As I know, the `partition` from the `partitionToCreate` map is created with a new status.

reference:https://github.com/NetEase/arctic/blob/master/hive/src/main/java/com/netease/arctic/hive/op/UpdateHiveFiles.java#L192

PR #982, opened 2023-01-05
[ARCTIC-981] Fix the Kerberos problem in the SortedPosDeleteWriter for the minor optimizing

### LINKED ISSUE #981: [Bug][Optimizer]:The optimizer can't output pos-delete files in minor optimizing because the user is incorrect
### What happened?

image
The optimizer can't output pos-delete files in minor optimizing because the user is incorrect.

### Affects Versions

0.3.0

_No response_

### How to reproduce

1. Create an arctic keyed table with user A
2. Start the optimizer with user B
3. Write some data and some deleted data into the change store and make the optimizer work
4. The pos-delete files maybe can't be produced by the optimizer, or the owner of pos-delete files is user B

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hzluting:
fix #981 

## Brief change log

  - *use table io to close in SortedPosDeleteWriter and BaseTaskWriter*

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/io/writer/SortedPosDeleteWriter.java
+++ b/core/src/main/java/com/netease/arctic/io/writer/SortedPosDeleteWriter.java
@@ -114,10 +114,7 @@ public class SortedPosDeleteWriter<T> implements Closeable {
   }
 
   public List<DeleteFile> complete() throws IOException {
-    io.doAs(() -> {
-      close();
-      return null;
-    });
+    close();
 
     return completedFiles;
   }
@@ -132,7 +129,10 @@ public class SortedPosDeleteWriter<T> implements Closeable {
 
   @Override
   public void close() throws IOException {
-    flushDeletes();
+    io.doAs(() -> {
+      flushDeletes();
+      return null;
+    });
   }
 
   private void flushDeletes() {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/netease/arctic/io/writer/SortedPosDeleteWriter.java
```diff
@@ -114,7 +114,10 @@ public void delete(CharSequence path, long pos, T row) {
   }
 
   public List<DeleteFile> complete() throws IOException {
-    close();
+    io.doAs(() -> {
```
zhoujinsong, line 117: I recommend wrapping the authentication logic around `SortedPosDeleteWriter#flushDeletes` because it will be called by other ways not only by `SortedPosDeleteWriter#complete`.

zhoujinsong (commented):
As the iceberg parquet writer may create new file from file system when close, should other writers like `BaseTaskWriter`, `IcebergFanoutPosDeleteWriter` should also add authentication logic arroundding when closing.

Reference:https://github.com/apache/iceberg/blob/master/parquet/src/main/java/org/apache/iceberg/parquet/ParquetWriter.java#L117

core/src/main/java/com/netease/arctic/io/writer/SortedPosDeleteWriter.java
hzluting: > I recommend wrapping the authentication logic around `SortedPosDeleteWriter#flushDeletes` because it will be called by other ways not only by `SortedPosDeleteWriter#complete`.

OK， I think it will be better.

zhoujinsong (approved):
LGTM.

PR #986, opened 2023-01-09
[ARCTIC-985] Introduce Table Blocker to implement concurrency controll between Writing and Optimizing

### LINKED ISSUE #985: [Feature]: Introduce a mechanism for concurrency control between Writing and Optimizing
### Description

We should introduce a mechanism to address the concurrency issues when Writing and Optimizing happen at the same time.

### Use case/motivation

There will be concurrency issues when table writing and optimizing happen at the same time in some cases:

**case1:**
Spark and Flink write into `ChangeStore` while self-optimizing is executing. TransactionId of the `ChangeStore` may be disordered in this case, which will bring data loss after self-optimizing. 

**case2:**
Spark concurrently writes into Arctic Table without conflict detection, especially when Spark writes into `ChangeStore`, because it's not easy to detect conflict since the `delete files` in `ChangeStore` is different from the standard Iceberg Table.

### Describe the solution

As described in this proposal
https://docs.google.com/document/d/13ftRtwYU9TXWojtA3jyeH9bVVgSfm2cdSjGYNzHEGKI/edit

We should introduce a mechanism to block operations on the table, like `Optimizing` and `BatchWrite`, and avoid the concurrency issue.

We call this mechanism `TableBlocker`, with the `TableBlocker`:
- the follow-up operations should be blocked if they are declared to be blocked
- the ongoing operations should not be affected

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

wangtaohz:

fix #985

## Brief change log

  - Add API of `TableBlocker`
  - Spark write with the use of `TableBlocker`
  - Implement `TableBlocker` in AMS
  - Self-Optimizing in AMS blocked by `TableBlocker`

## How was this patch tested?

## Documentation


spark/v3.1/spark/src/main/java/com/netease/arctic/spark/writer/KeyedSparkBatchWrite.java
```diff
@@ -149,13 +168,15 @@ public DataWriterFactory createBatchWriterFactory(PhysicalWriteInfo info) {
...
     public void commit(WriterCommitMessage[] messages) {
+      checkBlocker(tableBlockerManager);
       RewritePartitions rewritePartitions = table.newRewritePartitions();
       rewritePartitions.withTransactionId(txId);
 
       for (DataFile file : files(messages)) {
         rewritePartitions.addDataFile(file);
       }
       rewritePartitions.commit();
+      tableBlockerManager.release(block);
```
wangtaohz, line 179: It's better to release block like
```
try {
...
} finally {
     tableBlockerManager.release(block);
}
```

hellojinsilei: If there is an exception during the commit phase, it will execute `abort()`, the release operation is also done in `abort()`

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/FileInfoCacheService.java
```diff
@@ -147,6 +147,17 @@ public Boolean snapshotIsCached(TableIdentifier identifier, String innerTable, L
     }
   }
 
+  public Boolean snapshotIsCurrentCache(TableIdentifier identifier, String innerTable, Long snapshotId) {
+    if (snapshotId == -1) {
+      return true;
```
wangtaohz, line 152: `snapshotId` should not be -1

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -723,6 +750,14 @@ private List<OptimizeTaskItem> plan(long currentTime) {
       return Collections.emptyList();
     }
 
+    private boolean isOptimizeBlocked(TableIdentifier tableIdentifier) {
+      if (ServiceContainer.getTableBlockerService().isBlocked(tableIdentifier, BlockableOperation.OPTIMIZE)) {
+        LOG.debug("{} optimize is blocked continue", tableIdentifier);
```
wangtaohz, line 755: remove `continue`

zhoujinsong (approved):
LGTM.

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/table/ArcticSparkTable.java
```diff
@@ -182,7 +194,17 @@ public ScanBuilder newScanBuilder(CaseInsensitiveStringMap options) {
...
+    ArrayList<BlockableOperation> operations = Lists.newArrayList();
+    operations.add(BlockableOperation.BATCH_WRITE);
+    operations.add(BlockableOperation.OPTIMIZE);
+    Blocker block;
+    try {
+      block = tableBlockerManager.block(operations);
+    } catch (OperationConflictException e) {
+      throw new IllegalStateException("failed to block table " + arcticTable.id() + " with " + operations, e);
+    }
+    return new ArcticSparkWriteBuilder(arcticTable, info, tableBlockerManager, block);
```
baiyangtx, line 207: newWriteBuilder may be called during plan, request a block at buildForBatch() method, or even createBatchWriteFactory method

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/writer/KeyedSparkBatchWrite.java
baiyangtx: I think release blocker in finally is better.  the commit action may also failed.

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/writer/UnkeyedSparkBatchWrite.java
```diff
@@ -102,18 +112,30 @@ private abstract class BaseBatchWrite implements BatchWrite {
...
+                PropertyUtil.propertyAsInt(props, COMMIT_TOTAL_RETRY_TIME_MS, COMMIT_TOTAL_RETRY_TIME_MS_DEFAULT),
+                2.0 /* exponential */)
+            .throwFailureWhenFinished()
+            .run(file -> {
+              table.io().deleteFile(file.path().toString());
+            });
+      } finally {
+        tableBlockerManager.release(block);
+      }
+    }
```
baiyangtx, line 131: Does UnkeyedTable need a blocker during spark writing?  @wangtaohz @zhoujinsong 

I think the blocker is to handle PRIMARY KEY conflict between batch write and optimize

baiyangtx (changes requested):

spark/v3.1/spark/src/main/java/com/netease/arctic/spark/writer/UnkeyedSparkBatchWrite.java
wangtaohz: UnkeyedTable does not need a blocker

baiyangtx (approved):

baiyangtx (approved):

Issue #965, opened 2022-12-27
[Bug]: Kerberos authentication failure


zhongqishang:
### What happened?

ENV : k8s

Occurred during the follow three place, Kerberos authentication failure .

OptimizeQueueService.java:677
IcebergOptimizeCommit.java:165
TableOptimizeItem.java:426

```
2022-12-27 16:19:16,848 ERROR [Metastore Scheduled Commit Worker 0] [com.netease.arctic.ams.server.optimize.IcebergOptimizeCommit] [] - unexpected commit error hive_prod.ods_iceberg.xxx
org.apache.iceberg.exceptions.RuntimeIOException: Failed to open input stream for file: hdfs://xx/data/hive/warehouse/ods_iceberg.db/xxx/metadata/snap-9100413133805774208-1-ef6f8c7b-803e-42ef-9627-c0b767b6c93b.avro
        at org.apache.iceberg.hadoop.HadoopInputFile.newStream(HadoopInputFile.java:179) ~[iceberg-core-0.13.2.jar:?]
        at org.apache.iceberg.avro.AvroIterable.newFileReader(AvroIterable.java:101) ~[iceberg-core-0.13.2.jar:?]
        at org.apache.iceberg.avro.AvroIterable.iterator(AvroIterable.java:77) ~[iceberg-core-0.13.2.jar:?]
        at org.apache.iceberg.avro.AvroIterable.iterator(AvroIterable.java:37) ~[iceberg-core-0.13.2.jar:?]
        at org.apache.iceberg.relocated.com.google.common.collect.Iterables.addAll(Iterables.java:320) ~[iceberg-bundled-guava-0.13.2.jar:?]
        at org.apache.iceberg.relocated.com.google.common.collect.Lists.newLinkedList(Lists.java:237) ~[iceberg-bundled-guava-0.13.2.jar:?]
        at org.apache.iceberg.ManifestLists.read(ManifestLists.java:46) ~[iceberg-core-0.13.2.jar:?]
        at org.apache.iceberg.BaseSnapshot.cacheManifests(BaseSnapshot.java:141) ~[iceberg-core-0.13.2.jar:?]
        at org.apache.iceberg.BaseSnapshot.deleteManifests(BaseSnapshot.java:171) ~[iceberg-core-0.13.2.jar:?]
        at org.apache.iceberg.MergingSnapshotProducer.validationHistory(MergingSnapshotProducer.java:489) ~[iceberg-core-0.13.2.jar:?]
[...]

Issue #261, opened 2022-08-31
[Feature][Spark]: Support Spark 3.3


zhoujinsong:
### Description

Support more Spark version.

### Use case/motivation

Arctic support Spark 3.1 only now.
We need support more Spark version to satisfy more users' environment.

### Describe the solution

- Support Spark 3.3

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #989, opened 2023-01-10
[Arctic-506]:Upgrade iceberg version to 1.1.x

### LINKED ISSUE #506: [Feature]: support icebeg 1.0.0
### Description

iceberg 1.0.0已经发布了，尽快支持

### Use case/motivation

_No response_

### Describe the solution

_No response_

### Related issues

_No response_

### Code of Conduct

baiyangtx:
Currently, the arctic depends on iceberg v0.13.x, and this iceberg version is too old and with some problems hard to fix. 
So this PR is willing to upgrade the iceberg version to v1.1.x, and some issues will benefit from this PR:

- #965 : AMS does optimize commit for Iceberg format table failed for Kerberos authentication failure. 

This is because Iceberg Snapshot holds a HadoopTableIO object instead of an ActicFileIo,  Iceberg table loads snapshots when the table is loaded, so the wrapper of Table.io method is not worked in this situation. But since iceberg v1.0.x the Snapshot interface supports passing a FileIo object to do the mainfests files load, so this problem can be easily solved.

- #261 : Arctic Mixed format Spark connector depends on `iceberg-spark-${spark-version}` module,  it's unable to support Spark3.3 if we don't upgrade iceberg version to v1.0.x

And this PR will fix #506

## Brief change log

- Change the version of iceberg to 1.0.0 in pom.xml
- Adapt new interfaces of iceberg 1.0.x 
- Maintain the iceberg code depended by spark-2.3 module, and remove the dependence on iceberg-spark2 module 
- Maintain the iceberg code depended by flink-1.12 module, and remove the dependence on iceberg-flink-1.12 module 
- Refactor code of ams-server/ModifyTableSequence

## How was this patch tested?

## Documentation


baiyangtx:
@lklhdu Adapt Flink modules to Iceberg 1.0.x at this PR.

flink/v1.12/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java
```diff
@@ -68,30 +79,61 @@ public class ScanContext implements Serializable {
...
 
   public static final ConfigOption<Duration> MONITOR_INTERVAL =
-      ConfigOptions.key("monitor-interval").durationType().defaultValue(Duration.ofSeconds(10));
+      ConfigOptions.key("monitor-interval").durationType().defaultValue(Duration.ofSeconds(300));
+
+  public static final ConfigOption<Boolean> INCLUDE_COLUMN_STATS =
+      ConfigOptions.key("include-column-stats").booleanType().defaultValue(false);
+
+  public static final ConfigOption<Integer> MAX_PLANNING_SNAPSHOT_COUNT =
+      ConfigOptions.key("max-planning-snapshot-count").intType().defaultValue(Integer.MAX_VALUE);
```
YesOrNo828, line 88: missing the description. 
Please also add descriptions to other ConfigOtions.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/read/source/ArcticScanContext.java
```diff
@@ -49,6 +51,8 @@ public class ArcticScanContext extends ScanContext implements Serializable {
   protected ArcticScanContext(
       boolean caseSensitive,
       Long snapshotId,
+      StreamingStartingStrategy startingStrategy,
```
YesOrNo828, line 54: does it duplicate with `scanStartupMode`?

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/read/source/ArcticScanContext.java
```diff
@@ -49,6 +51,8 @@ public class ArcticScanContext extends ScanContext implements Serializable {
...
+      Long startSnapshotTimestamp,
```
YesOrNo828, line 55: what's the different with `asOfTimestamp`?

YesOrNo828 (commented):
@lklhdu  Thanks for your contribution. I left some comments.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/read/source/ArcticScanContext.java
lklhdu: asOfTimestamp is used for BATCH scan mode, and startSnapshotTimestamp is used for INCREMENTAL_APPEND scan mode.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/read/source/ArcticScanContext.java
lklhdu: There is no duplication between them. The scanStartupMode parameter is an Arctic-defined parameter that controls the startup method with the primary key table read; the startingStrategy is an Iceberg-native parameter that is only passed to the Iceberg ScanContext constructor in the ArcticScanContext class and is not actually used.

shidayang:
Iceberg version maybe 1.1.x?

core/src/main/java/com/netease/arctic/scan/TableEntriesScan.java
```diff
@@ -147,14 +149,21 @@ public Builder project(Schema schema) {
       return this;
     }
 
+    public Builder withMetadataTableType(MetadataTableType metadataTableType) {
```
zhoujinsong, line 152: AFAIK, this class is used to scan table entries table, why should we add this method here?

hellojinsilei: I consider using MetadataTableType this parameter to control the type of file to be scanned.
Otherwise, the original class can only get files of type `MetadataTableType.ENTRIES`.

hellojinsilei: It's also considered to extend the use of `TableEntriesScan`

zhoujinsong (approved):
LGTM.

PR #991, opened 2023-01-10
[ARCTIC-966][Core]Check the written row number before closing the file writer

### LINKED ISSUE #966: [Bug]: Failed to process files of length 0 when optimizing
### What happened?

Flink stream ingestion job committed an equality-delete file of length 0 after failover, then self-optimizing caught an exception when reading this file.

image

### Affects Versions

master/0.3.0

Optimizer, Flink

### How to reproduce

1.After Flink job failover, it occasionally commits a file of length 0.
2.Then the self-optimizing will fail, and we can find the exception in optimizer.

### Relevant log output

_No response_

### Anything else

I think we should avoid committing a file of length 0, then no exceptions will occur when MOR/Optimizing.

### Code of Conduct

zhoujinsong:

To fix #966, check the written row number before closing the file writer.

## Brief change log

  - *Calculate the written row number in task writer*
  - *Complete only if written row number > 0*

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/io/writer/BaseTaskWriter.java
+++ b/core/src/main/java/com/netease/arctic/io/writer/BaseTaskWriter.java
@@ -84,7 +84,7 @@ public abstract class BaseTaskWriter<T> implements TaskWriter<T> {
     } else {
       writer = dataWriterMap.get(writerKey);
     }
-    writer.write(row);
+    write(writer, row);
 
     if (shouldRollToNewFile(writer)) {
       writer.close();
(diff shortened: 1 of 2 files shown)


wangtaohz (approved):
LGTM. 

It's necessary to make sure rows > 0 when writer is closed.

However, I checked the code, but I'm still not sure why Flink job committed a 0 length file with the code before.

zhoujinsong:
> LGTM.
> 
> It's necessary to make sure rows > 0 when writer is closed.
> 
> However, I checked the code, but I'm still not sure why Flink job committed a 0 length file with the code before.

Maybe we can merge this PR and keep watching if the same problem still happens.

Issue #925, opened 2022-12-13
[Improvement]: Introduce a disk spillable map for optimizer to avoid OOM


majin1102:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Each optimizer loads delete files in a map to filter insert records, and when delete files are too large, may causes OOM error. In AMS view, optimizers which have already OOM may be busy or hanged.

we shall introduce a disk spillable map to store delete records when the size reaches a threshold too avoid OOM problem. This implemention will refer to ExternalSpillableMap of HUDI

### How should we improve?

Use  a disk spillable map instead of a memory storing map with features as below:

* fits both mix format and iceberg format
* fits both self-optimizing, merge on read and optimize command in future
* use memory map before map size hits a configurable threshold
* use rocksdb as spilled map backend
* fits partial upsert in the future(column family feature)
* for optimizers, rocksdb backend stays after first loaded and could be use cyclically

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #992, opened 2023-01-10
[ARCTIC-925][Optimizer] The optimize group adds config to support spill map and improve spill map with the custom directory

hzluting:
improve #925 

## Brief change log

  - *The optimize group adds spill map config to start optimizer*
  - *The spill map supports the custom directory*

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/utils/map/RocksDBBackend.java
+++ b/core/src/main/java/com/netease/arctic/utils/map/RocksDBBackend.java
@@ -50,17 +50,29 @@ import java.util.stream.Collectors;
 
 public class RocksDBBackend {
   private static final Logger LOG = LoggerFactory.getLogger(RocksDBBackend.class);
-  private static final String BACKEND_BASE_DIR = System.getProperty("rocksdb.dir");
+  private static final String BACKEND_BASE_DIR = System.getProperty("java.io.tmpdir");
   private static final ThreadLocal<RocksDBBackend> instance =
           new ThreadLocal<>();
 
   public static RocksDBBackend getOrCreateInstance() {
     RocksDBBackend backend = instance.get();
     if (backend == null) {
-      backend = create();
+      backend = create(BACKEND_BASE_DIR);
     }
     if (backend.closed) {
-      backend = create();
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/netease/arctic/utils/map/RocksDBBackend.java
```diff
@@ -50,17 +50,29 @@
...
+    return backend;
+  }
+
+  public static RocksDBBackend getOrCreateInstance(String backendBaseDir) {
+    RocksDBBackend backend = instance.get();
+    if (backend == null) {
+      backend = create(backendBaseDir);
+    }
+    if (backend.closed) {
+      backend = create(backendBaseDir);
```
shidayang, line 75: The instant is share with thread, so the backendBaseDir can not be modify once created in the same thread

core/src/main/java/com/netease/arctic/utils/map/StructLikeCollections.java
```diff
@@ -26,26 +26,34 @@ public class StructLikeCollections {
...
       this.maxInMemorySizeInBytes = maxInMemorySizeInBytes;
     }
   }
 
+  public StructLikeCollections(boolean enableSpillableMap, Long maxInMemorySizeInBytes, String backendBaseDir) {
+    if (enableSpillableMap) {
+      this.maxInMemorySizeInBytes = maxInMemorySizeInBytes;
+      this.backendBaseDir = backendBaseDir;
+    }
+  }
```
shidayang, line 42: maxInMemorySizeInBytes may be need a default value?

shidayang (changes requested):

core/src/main/java/com/netease/arctic/utils/map/RocksDBBackend.java
hzluting: > The instant is share with thread, so the backendBaseDir can not be modify once created in the same thread

I think the backendBaseDir should be modified with new backendBaseDir

shidayang (approved):

PR #1001, opened 2023-01-12
[WIP][ARCTIC-1000][AMS] Persist on the property of task after the update

### LINKED ISSUE #1000: [Bug][AMS]: The optimize task property doesn't persist in MySQL after set 'max-execute-time'
### What happened?

The optimize task property adds the key named 'max-execute-time' that doesn't persist in MySQL.
So if the AMS restarts,  the task will lose the property 'max-execute-time' value

### Affects Versions

0.3.0

AMS

### How to reproduce

1. create an arctic table
2. write the data into the table, and confirm optimizing task execution time over 30min
3. set the `self-optimizing.execute.timeout` as 3600000
4. restart the AMS when the optimizer executes the task
5. the task will always be failed because the execution expired

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hzluting:
fix #1000 

## Brief change log

  - *persist the property of the task after updating the 'max-execute-time' when the task on executing*

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/mapper/OptimizeTaskRuntimesMapper.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/mapper/OptimizeTaskRuntimesMapper.java
@@ -18,6 +18,7 @@
 
 package com.netease.arctic.ams.server.mapper;
 
+import com.netease.arctic.ams.server.model.BaseOptimizeTask;
 import com.netease.arctic.ams.server.model.BaseOptimizeTaskRuntime;
 import com.netease.arctic.ams.server.mybatis.Long2TsConvertor;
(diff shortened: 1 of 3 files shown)


zhoujinsong:
cc @wangtaohz

wangtaohz (changes requested):
It's not rigorous to update the 'max-execute-time' when the task on executing. Because we are unable to send it to Optimizer who is using this property to execute the task.

hzluting:
> It's not rigorous to update the 'max-execute-time' when the task on executing. Because we are unable to send it to Optimizer who is using this property to execute the task.

Yes, this only affects the next execution after updating the 'max-execute-time'. I don't think we need to stop the task once someone updates the `self-optimizing.execute.timeout`.

PR #1010, opened 2023-01-12
[ARCTIC-994] Introduce design of `TransactionId` generation to resolving data conflicts

### LINKED ISSUE #994: [Feature]: A new design of resolving data conflicts without relying on AMS to generate TransactionId
### Description

To resolve the data conflict of Arctic KeyedTable(Mixed-format table) with concurrent batch write and stream write, we introduce a design of version for each record, in order to keep the data with a higher version when row conflicts arise.

### Use case/motivation

In the case of a concurrent batch write and stream write, row conflicts are likely to arise, especially when batch write takes a long time, and the batch write will fail to commit.

We have introduced `TransactionId` and `offset` to determine the record version, but the generation of `TransactionId` relies on the AMS, which leads to reliability issues for data writing.

### Describe the solution

The design of the record version:
Version = (TransactionId, offset)
 - TransactionId: the sequence of each Transaction
 - offset: the offset of record in a Transaction, to determine the record version in a transaction

Two ways to generate `TransactionId`:
- generate on transaction begin: commit an empty snapshot to ChangeTable,  and get the snapshot sequence as TransactionId
- generate on transaction commit: get the change snapshot sequence when commit as TransactionId

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

wangtaohz:

fix #994 

## Brief change log
1. generate `TransactionId` from change snapshot sequence
2. spark operation begins a transaction with new `TransactionId`
3. flink operation does not begin a transaction, and uses snapshot sequence as `TransactionId` when committing. The `TxId` in fileName is set to be 0, and the correct `TransactionId` should get from iceberg metadata in this case
4. flink incremental pull using `TableEntriesScan` instead of the iceberg `TableScan`, to get the sequence number from iceberg metadata when `TxId` in fileName is 0
5. import a UUID into the file name to distinguish files with the same `TxId` = 0, in `CommonOutputFileFactory` and `AdaptHiveOutputFileFactory`
6. Introduce `com.netease.arctic.io.FileNameHandle` to resolve file name.
7. Introduce `com.netease.arctic.data.file.ContentFileWithSequence` to wrap ContentFile.
8. Make `com.netease.arctic.scan.BaseChangeTableIncrementalScan` can return SequenceNumber.
9. make the using change snapshot sequence as `TransactionId` be compatible with old `TransactionId`  (make sure snapshot sequence >= old transactionId) when ams start, in `UpdateTool`

## How was this patch tested?

## Documentation


wangtaohz:
When `TransactionId` is generated on commit, the `TransactionId` in the file name is 0, so we have to import a UUID into the file name to distinguish files.

core/src/main/java/com/netease/arctic/data/DefaultKeyedFile.java
```diff
@@ -175,11 +174,15 @@ public int hashCode() {
...
+
+    private static final String KEYED_FILE_NAME_PATTERN_STRING = "(\\d+)-(\\w+)-(\\d+)-(\\d+)-(\\d+)-(\\d+)\\.\\w+";
+    private static final Pattern KEYED_FILE_NAME_PATTERN = Pattern.compile(KEYED_FILE_NAME_PATTERN_STRING);
+
     private final long transactionId;
     private final DataFileType type;
     private final DataTreeNode node;
 
-    public FileMeta(Long transactionId, DataFileType type, DataTreeNode node) {
+    private FileMeta(Long transactionId, DataFileType type, DataTreeNode node) {
```
wangtaohz, line 185: `Long transactionId` may be `long transactionId` to avoid NPE

wangtaohz:
cc @YesOrNo828

core/src/main/java/com/netease/arctic/table/BaseKeyedTable.java
```diff
@@ -170,11 +172,11 @@ public UpdateProperties updateProperties() {
 
   @Override
   public long beginTransaction(String signature) {
-    try {
-      return client.allocateTransactionId(this.tableMeta.getTableIdentifier(), signature);
-    } catch (TException e) {
-      throw new IllegalStateException("failed begin transaction", e);
-    }
+    AppendFiles appendFiles = changeTable.newAppend();
+    // TODO add summary, use fast append?
```
zhoujinsong, line 176: I don't think fast append has any difference with append here.
BTW， we may need add some comment here to describe this impelementation.

wangtaohz: OK, I will add some comment

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
```diff
@@ -148,16 +148,7 @@ private void initTaskWriterFactory(Long mask) {
...
-    if (table.isKeyedTable()) {
-      String signature = BaseEncoding.base16().encode((jobId + checkpointId).getBytes());
-      transaction = table.asKeyedTable().beginTransaction(signature);
-      LOG.info("table:{}, signature:{}, transactionId:{}. From jobId:{}, ckpId:{}", table.name(), signature,
-          transaction, jobId, checkpointId);
-    } else {
-      transaction = null;
-    }
-    return transaction;
+    return null;
```
YesOrNo828, line 151: The variable transactionId is always null in the flink writer operator lifecycle, so this method and [`setTransactionId`](https://github.com/NetEase/arctic/blob/master/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticRowDataTaskWriterFactory.java#L56) should be removed because they seemed redundant.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticRowDataTaskWriterFactory.java
```diff
@@ -59,12 +59,7 @@ public void setTransactionId(Long transactionId) {
 
   @Override
   public void initialize(int taskId, int attemptId) {
-    if (table.isKeyedTable()) {
-      Preconditions.checkNotNull(transactionId, "TransactionId should be set first. " +
-          "Invoke setTransactionId() before this method");
-    } else {
-      Preconditions.checkArgument(transactionId == null, "TransactionId should be null for unkeyed table.");
-    }
+    Preconditions.checkArgument(transactionId == null, "TransactionId should be null for table.");
```
YesOrNo828, line 62: I think the variable transactionId is the `final` variable which is better.
`private transient final Long transactionId = null;`

core/src/main/java/com/netease/arctic/io/FileNameHandle.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+    DataTreeNode node = DataTreeNode.ofId(nodeId);
+    return new DefaultKeyedFile.FileMeta(transactionId, type, node);
+  }
+
+  /**
+   * Path writen by hive can not be pared by arctic format. so we set it transactionId as 0.
+   * @param path file path
+   */
+  public static DefaultKeyedFile.FileMeta parseBase(String path) {
```
YesOrNo828, line 92: Could you add more detail about the param path? E.g., path format and demo. I think it will help devs easier to understand what this parsing tries to do.

YesOrNo828 (commented):
@wangtaohz  Thanks for your contribution. I left some comments.

@zstraw  Do you have a time to take a look?

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticRowDataTaskWriterFactory.java
zstraw, line 62: Now that transactionId should be null, why setTransactionId() still be exposed? Would it be better to remove all transactionId's setting in this factory?

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
```diff
@@ -148,16 +148,7 @@ private void initTaskWriterFactory(Long mask) {
   }
 
   private Long getTransactionId() {
```
zstraw, line 150: After removing transaction in ArcticFileWriter.java, it's useless to initTaskWriterFactory(null) in processElement. And the input parameter of mask can be replaced with long in initTaskWriterFactory().

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
wangtaohz: OK, I will remove it

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
wangtaohz: I removed it.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticRowDataTaskWriterFactory.java
wangtaohz: I removed it.

wangtaohz: I added `final` to it.

core/src/main/java/com/netease/arctic/io/FileNameHandle.java
```diff
@@ -0,0 +1,151 @@
...
+import com.netease.arctic.io.writer.TaskWriterKey;
+import com.netease.arctic.utils.IdGenerator;
+import com.netease.arctic.utils.TableFileUtils;
+import org.apache.iceberg.FileFormat;
+
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class FileNameHandle {
```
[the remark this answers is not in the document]
wangtaohz: `FileNameHandle` is for both `KeyedTable` and `UnkeyedTable`, so `KeyedFileNameGenerator` may be ambiguous.

BTW, I think `FileNameHandler` is better than `FileNameHandle`.

cc @shidayang

core/src/main/java/com/netease/arctic/io/FileNameHandle.java
wangtaohz: I add some comments for this class to show more details. like 
```
/**
 * File name pattern:${tree_node_id}-${file_type}-${transaction_id}-${partition_id}-${task_id}-{operation_id}-{count}
 * like:
 * 1-B-100-0-0-4217271085623029157-0.parquet
 * 4-ED-101-0-0-9009257362994691056-1.parquet
 * 4-I-101-0-0-9009257362994691056-2.parquet
 */
public class FileNameHandle {
```

zstraw (approved):

wangtaohz:
This PR also fix #1045

core/src/main/java/com/netease/arctic/io/FileNameHandle.java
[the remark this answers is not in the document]
wangtaohz: The class name is changed to `FileNameGenerator`

spark/v2.3/spark/src/test/java/com/netease/arctic/spark/SparkTestContext.java
```diff
@@ -276,9 +275,7 @@ public static GenericRecord newRecord(Schema schema, Object... val) {
 
   public static void writeChange(TableIdentifier identifier, ChangeAction action, List<Record> rows) {
     KeyedTable table = SparkTestContext.catalog(identifier.getCatalog()).loadTable(identifier).asKeyedTable();
-    long txId = table.beginTransaction(System.currentTimeMillis() + "");
     try (TaskWriter<Record> writer = GenericTaskWriters.builderFor(table)
-        .withTransactionId(txId)
         .withChangeAction(action)
         .buildChangeWriter()) {
```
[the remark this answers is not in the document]
wangtaohz: Here, it writes some data into ChangeStore, there is no need to begin a transaction here, so I remove the `transactionId`.

baiyangtx (approved):

Issue #1009, opened 2023-01-12
[Subtask][AMS]: Add the config of the spill map in optimize group for the optimizer to enable the spill map


hzluting:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### Description

I want to add some config of the spill map in optimize group as below for the optimizer to enable the spill map:

1.enable_spill_map(whether enable the spill map, true or false, default value is false)
2.backend_base_dir(the base directory for the disk spill map, default value is System.getProperty("java.io.tmpdir"))
3.max_delete_memory_size_in_byte(max delete map size in memory for optimizer)

### Parent issue

#925

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

majin1102:
unify properties naming as : https://arctic.netease.com/ch/configurations/

recommand:

1. spillable.map.enabled
2. spillable.map.dir
3. spillable.memory.limit

hzluting:
> Let's unify properties naming as : https://arctic.netease.com/ch/configurations/
> 
> recommand:
> 
> 1. spillable.map.enabled
> 2. spillable.map.dir
> 3. spillable.memory.limit

OK.

PR #1017, opened 2023-01-16
[ARCITC-1009][AMS] Modify spill map config in optimize group

hzluting:
improve #1009 

## Brief change log

  - *modify spill map config keys in optimize group*
  - *modify the shell for the local optimizer*

## How was this patch tested?

## Documentation

### DIFF
--- a/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
+++ b/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
@@ -78,16 +78,16 @@ public class FlinkOptimizer implements StatefulOptimizer {
     }
     String cmd = flinkHome + "/bin/flink run -m yarn-cluster ";
 
-    //add flink executor config
+    // add flink executor config
     int tmMemory = groupProperties.getInteger("taskmanager.memory");
     cmd += " -ytm " + tmMemory;
     int jmMemory = groupProperties.getInteger("jobmanager.memory");
     cmd += " -yjm " + jmMemory;
 
     // spill map config
-    Boolean enableSpillMap = groupProperties.getBoolean("enable_spill_map");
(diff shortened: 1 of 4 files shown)


optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
```diff
@@ -78,16 +78,16 @@ public void start() throws IOException {
...
     int tmMemory = groupProperties.getInteger("taskmanager.memory");
     cmd += " -ytm " + tmMemory;
     int jmMemory = groupProperties.getInteger("jobmanager.memory");
     cmd += " -yjm " + jmMemory;
 
     // spill map config
-    Boolean enableSpillMap = groupProperties.getBoolean("enable_spill_map");
-    String backendBaseDir = groupProperties.getString("backend_base_dir");
-    Long maxDeleteMemorySize = groupProperties.getLong("max_delete_memory_size_in_byte");
+    Boolean enableSpillMap = groupProperties.getBoolean("spillable.map.enabled");
```
zhoujinsong, line 88: Should we manage these property names in a static class?

zhoujinsong:
I am wondering about the property names like `splillable.map.enable` is easy for a user to understand.

optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
hzluting: Yes

hzluting:
> I am wondering about the property names like `splillable.map.enable` is easy for a user to understand.

I think It is easy for the user who knows the principle of reading the delete data. 
But as a parameter of the optimizer group is worth considering.
Because from the user's perspective, this parameter is used to avoid the delete map OOM.

zhoujinsong (approved):
LGTM.

PR #1046, opened 2023-02-02
[ARCTIC-1016] Moving files will fail if the target parent directory doesn't exist in HDFS

### LINKED ISSUE #1016: [Bug][AMS]: Moving files will fail if the target parent directory doesn't exist in HDFS
### What happened?

For the unKeyed MIXED_HIVE table with partition, we will move the result files to the hive partition location to support the hive. The rename method will return false and doesn't throw the exception in HDFS when the target parent directory doesn't exist, but we don't handle it according to the return value.
So, the major optimizing commit will commit the file that isn't moved. The "File not found" exception will throw when we read or optimize this table.

### Affects Versions

0.4.x

_No response_

### How to reproduce

1. create an unKeyed MIXED_HIVE table with partition
2. trigger a major optimizing for this table and optimize the partition that doesn't exist in the hive
3. read this table will throw the "File not found" exception

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

shendanfengg:

Fix #1016 

## Brief change log

*Exception thrown when FileSystem returns false.
*If the new path does not exist during Support Hive Commit,  and then create it.
*Fix some simple grammar errors.

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
+++ b/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
@@ -89,7 +89,7 @@ public class ArcticHadoopFileIO extends HadoopFileIO implements ArcticFileIO {
         result = false;
       }
       if (result == false) {
-        LOG.warn("Fail to delete file " + path + " and file system return false, false, need to check the hdfs path");
+        LOG.warn("Fail to delete file " + path + " and file system return false, need to check the hdfs path");
       }
       return result;
     });
--- a/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
+++ b/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
@@ -78,7 +78,8 @@ public class ArcticHadoopFileIO extends HadoopFileIO implements ArcticFileIO {
   }
 
   @Override
-  public boolean deleteFileWithResult(String path, boolean recursive) {
+  public boolean deleteFileWith
+  (String path, boolean recursive) {
     return tableMetaStore.doAs(() -> {
       Path toDelete = new Path(path);
       FileSystem fs = getFs(toDelete);
@@ -88,7 +89,7 @@ public class ArcticHadoopFileIO extends HadoopFileIO implements ArcticFileIO {
       } catch (IOException e) {
         result = false;
       }
-      if (!result ) {
+      if (!result) {
         LOG.warn("Fail to delete file " + path + " and file system return false, need to check the hdfs path");
       }
       return result;
--- a/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
+++ b/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
@@ -78,8 +78,7 @@ public class ArcticHadoopFileIO extends HadoopFileIO implements ArcticFileIO {
   }
 
   @Override
-  public boolean deleteFileWith
-  (String path, boolean recursive) {
+  public boolean deleteFileWith(String path, boolean recursive) {
     return tableMetaStore.doAs(() -> {
       Path toDelete = new Path(path);
       FileSystem fs = getFs(toDelete);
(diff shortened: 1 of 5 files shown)


hzluting:
cc @zhoujinsong

core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
```diff
@@ -165,9 +172,14 @@ public boolean rename(String src, String dts) {
...
+              " and file system return false, need to check the hdfs path");
+        } else {
+          return true;
+        }
       } catch (IOException e) {
-        throw new UncheckedIOException("Failed to rename: from " + src + " to " + dts, e);
+        throw new UncheckedIOException("Fail to rename: from " + src + " to " + dts, e);
       }
     });
   }
```
wangtaohz, line 185: If we choose to throw an exception when fail to rename, this method should return `void` instead of return `boolean`.

hive/src/main/java/com/netease/arctic/hive/CachedHiveClientPool.java
```diff
@@ -73,7 +73,7 @@ private synchronized void init() {
   @Override
   public <R> R run(Action<R, HMSClient, TException> action) throws TException, InterruptedException {
     try {
-      return tableMetaStore.doAs(() -> clientPool().run(action));
+      return tableMetaStore.doAs(() -> clientPool().run(action, true));
```
wangtaohz, line 76: What problem is this going to fix? Is there another PR to fix this?

wangtaohz (commented):
@shendanfengg Thanks for your work.

I left some comment, please take a look.

hive/src/main/java/com/netease/arctic/hive/CachedHiveClientPool.java
shendanfengg: I have also committed this change in another pr and I will delete it in this pr.

wangtaohz (approved):

core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
```diff
@@ -84,6 +88,9 @@ public boolean deleteFileWithResult(String path, boolean recursive) {
       } catch (IOException e) {
         result = false;
       }
+      if (result == false) {
```
zhoujinsong, line 91:
```suggestion
      if (!result ) {
```

core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
```diff
@@ -159,16 +166,20 @@ public boolean isEmptyDirectory(String location) {
...
-  public boolean rename(String src, String dts) {
-    return tableMetaStore.doAs(() -> {
+  public void rename(String src, String dts) {
+    tableMetaStore.doAs(() -> {
       Path srcPath = new Path(src);
       Path dtsPath = new Path(dts);
       FileSystem fs = getFs(srcPath);
       try {
-        return fs.rename(srcPath, dtsPath);
+        if (fs.rename(srcPath, dtsPath) == false) {
```
zhoujinsong, line 175:
```suggestion
        if (!fs.rename(srcPath, dtsPath)) {
```

core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
```diff
@@ -185,21 +196,25 @@ public boolean exists(String path) {
...
   @Override
-  public boolean mkdirs(String path) {
-    return tableMetaStore.doAs(() -> {
+  public void mkdirs(String path) {
+    tableMetaStore.doAs(() -> {
       Path filePath = new Path(path);
       FileSystem fs = getFs(filePath);
       try {
-        return fs.mkdirs(filePath);
+        if (fs.mkdirs(filePath) == false) {
```
zhoujinsong, line 210:
```suggestion
        if (!fs.mkdirs(filePath)) {
```

zhoujinsong (commented):
I left some suggestion, please task a look.

wangtaohz:
I have cherry-picked it to 0.3.x

PR #1048, opened 2023-02-02
[ARCTIC-1047][Core] Fix the the logic of estimating size in the SimpleSpillableMap

### LINKED ISSUE #1047: [Bug][Core]: The estimate of memory size is incorrect in the SimpleSpillableMap
### What happened?

When I tested the performance of the spill map, I found the performance of the memory map in the SimpleSpillableMap significantly worse than the performance of the SimpleMemoryMap.
I found the estimate of memory size is incorrect in the SimpleSpillableMap because the key is the StructLikeWrapper which contains extra content(such as Comparator). When we estimate the key's size of the memory map in memory, we should only consider the record's size in the StructLikeWrapper rather than its all size in the memory map because some content in the StructLikeWrapper references the same instance.

### Affects Versions

0.4.x

Core

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hzluting:
fix #1047

## Brief change log

  - *fix the logic of estimating size in the SimpleSpillableMap*
  - *unified the units in optimizer group config*

## How was this patch tested?

## Documentation


core/src/main/java/com/netease/arctic/utils/map/SizeEstimator.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netease.arctic.utils.map;
+
+import org.openjdk.jol.info.GraphLayout;
```
wangtaohz, line 21: useless import

core/src/main/java/com/netease/arctic/utils/map/SimpleSpillableMap.java
```diff
@@ -32,32 +31,44 @@
   private static final int RECORDS_TO_SKIP_FOR_ESTIMATING = 200;
   private final long maxInMemorySizeInBytes;
   private final String backendBaseDir;
+  // Size Estimator for key type
+  private final SizeEstimator<K> keySizeEstimator;
+  // Size Estimator for key types
+  private final SizeEstimator<T> valueSizeEstimator;
   private Map<K, T> memoryMap;
   private Optional<SimpleSpilledMap<K, T>> diskBasedMap = Optional.empty();
   private Long currentInMemoryMapSize;
```
wangtaohz, line 40: `long currentInMemoryMapSize` may be better, since the `null` value is not used.

core/src/main/java/com/netease/arctic/utils/map/StructLikeWrapperSizeEstimator.java
```diff
@@ -0,0 +1,32 @@
...
+import com.netease.arctic.iceberg.optimize.StructLikeWrapper;
+import com.netease.arctic.utils.ObjectSizeCalculator;
+
+/**
+ * Size Estimator for StructLikeWrapper record payload.
+ */
+public class StructLikeWrapperSizeEstimator implements SizeEstimator<StructLikeWrapper> {
+  @Override
+  public long sizeEstimate(StructLikeWrapper structLikeWrapper) {
+    return ObjectSizeCalculator.getObjectSize(structLikeWrapper.get());
```
wangtaohz, line 30: If `structLikeWrapper == null`, it will incur NPE.

wangtaohz (commented):
@hzluting Thanks for your contribution. I left some comments.

core/src/main/java/com/netease/arctic/utils/map/SizeEstimator.java
hzluting: OK

core/src/main/java/com/netease/arctic/utils/map/SimpleSpillableMap.java
hzluting: Yes

core/src/main/java/com/netease/arctic/utils/map/StructLikeWrapperSizeEstimator.java
hzluting: OK

wangtaohz:
LGTM

wangtaohz (approved):

zhoujinsong (approved):
The codes look good to me.
But I am wondering the spillablemap performance now.

core/src/main/java/com/netease/arctic/utils/map/StructLikeWrapperSizeEstimator.java
```diff
@@ -0,0 +1,36 @@
...
+ * Size Estimator for StructLikeWrapper record payload.
+ */
+public class StructLikeWrapperSizeEstimator implements SizeEstimator<StructLikeWrapper> {
+  @Override
+  public long sizeEstimate(StructLikeWrapper structLikeWrapper) {
+    if (structLikeWrapper == null) {
+      return 0;
+    }
+
+    return ObjectSizeCalculator.getObjectSize(structLikeWrapper.get());
```
shidayang, line 34: The object size is also incorrect, because StructLike is an interface. You only can calculate object size implementation

shidayang (changes requested):

core/src/main/java/com/netease/arctic/utils/map/StructLikeWrapperSizeEstimator.java
zhoujinsong: StructLike had multiple implementations, should we implement different calculation process for different situation?

hzluting: I find the key Serializer in the spill map uses the StructLikeCopy to serialize. The key in the StructLikeBaseMap also can use StructLikeCopy. And the size of the StructLikeCopy instance is affirmatory.

shidayang (approved):

PR #1058, opened 2023-02-06
[ARCTIC-994] fix read legacy transaction id when MOR

wangtaohz:

Now, the transaction ids of each partition are stored in the base table's property, and the legacy transaction id is supposed to be read when the transaction id is not present, in order to be compatible with the legacy tables from 0.3.*.

However, when the transaction id is not present, we return `-1`, which is error-prone, and causes compatibility problem for MOR.

## Brief change log

  - return `null` instead of `-1` when transaction id is not present.
  - refactor `OverwriteBaseFiles` when handling transaction id
  - for data consistency, allow `KeyedPartitionRewrite` to set the transaction id to a smaller value 

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/op/OverwriteBaseFiles.java
+++ b/core/src/main/java/com/netease/arctic/op/OverwriteBaseFiles.java
@@ -134,9 +134,9 @@ public class OverwriteBaseFiles extends PartitionTransactionOperation {
 
   @Override
   protected StructLikeMap<Long> apply(Transaction transaction, StructLikeMap<Long> partitionMaxTxId) {
-    applyDeleteExpression();
     Preconditions.checkState(this.dynamic != null,
         "updateMaxTransactionId() or updateMaxTransactionIdDynamically() must be invoked");
+    applyDeleteExpression();
 
     StructLikeMap<Long> changedPartitionTransactionId = null;
     if (this.dynamic) {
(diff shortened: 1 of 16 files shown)


core/src/main/java/com/netease/arctic/op/OverwriteBaseFiles.java
```diff
@@ -158,15 +167,12 @@ protected StructLikeMap<Long> apply(Transaction transaction, StructLikeMap<Long>
...
-        for (DeleteFile d : this.addDeleteFiles) {
-          partitionData = keyedTable.spec().isUnpartitioned() ? TablePropertyUtil.EMPTY_STRUCT : d.partition();
-          partitionMaxTxId.put(partitionData, getPartitionMaxTxId(partitionData));
+        if (dynamicPartition) {
+          for (DeleteFile d : this.addDeleteFiles) {
+            changedPartitionTransactionId.put(d.partition(), this.transactionId);
+          }
         }
 
-        for (DataFile d : this.deleteFiles) {
```
wangtaohz, line 166: I don't this `this.deleteFiles` should be handled here, so I remove these code

wangtaohz:
cc @baiyangtx @zhoujinsong @shendanfengg

baiyangtx (approved):

zhoujinsong (approved):

PR #1076, opened 2023-02-08
[ARCTIC-1075][AMS]When use native iceberg, update catalog properties does not take effect in terminal

### LINKED ISSUE #1075: [AMS][Bug]: When use native iceberg, update catalog properties does not take effect in terminal
### What happened?

When use native iceberg, update catalog properties then use terminal to execute sql, properties does not take effect

### Affects Versions

0.4.1

AMS

### How to reproduce

step1: create native iceberg catalog
step2: use terminal to execute sql
step3: update catalog properties
step4: use terminal to execute SQL agian 

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hellojinsilei:

fix #1075 

## Brief change log

add cleanSession() method

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/terminal/TerminalManager.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/terminal/TerminalManager.java
@@ -139,7 +139,7 @@ public class TerminalManager {
     return new LogInfo(sessionContext.getStatus().name(), sessionContext.getLogs());
   }
 
-  public void removeSessionById() {
+  public void cleanSession() {
     sessionMap.clear();
   }
 
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/terminal/TerminalManager.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/terminal/TerminalManager.java
@@ -140,7 +140,7 @@ public class TerminalManager {
   }
 
   public synchronized void cleanSession(String catalogName) {
-    for(String key : sessionMap.keySet()) {
+    for (String key : sessionMap.keySet()) {
       if (key.endsWith("-" + catalogName)) {
         sessionMap.remove(key);
       }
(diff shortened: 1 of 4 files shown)


baiyangtx (changes requested):
Unreasonable， the action of modify catalog metadata may from other user. The cookie is different. 

You should modify TerminalManager when catalog meta is changed, and TerminalManager will scan all Session and find out the related session and close them.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/terminal/TerminalManager.java
```diff
@@ -139,6 +139,10 @@ public LogInfo getExecutionLog(String sessionId) {
     return new LogInfo(sessionContext.getStatus().name(), sessionContext.getLogs());
   }
 
+  public void cleanSession() {
+    sessionMap.clear();
+  }
+
```
baiyangtx, line 145: should be synchronized

baiyangtx (changes requested):

ams/ams-server/src/main/java/com/netease/arctic/ams/server/terminal/TerminalManager.java
```diff
@@ -139,6 +139,10 @@ public LogInfo getExecutionLog(String sessionId) {
     return new LogInfo(sessionContext.getStatus().name(), sessionContext.getLogs());
   }
 
+  public synchronized void cleanSession() {
+    sessionMap.clear();
+  }
+
```
baiyangtx, line 145: I don't think sessionMap.clear is a good idea.

The session is a heavy object and maybe hold some external resources. close() method should be call if you don't need the session any more.

but don't call session.close() directly in controller.  because: 
- session.close() is a heavy method and maybe failed.
- session.close() may cost times.
- some session may be in running state and can't closed.
mark the session as deperated and close them in gc thread.

baiyangtx (changes requested):

baiyangtx:
This issue is more complex than think, in the local mode, the terminal can't control the catalog already loaded by the spark engine. so a better idea is solve this by using ArcticSparkCatalog load native iceberg table.

Issue #1066, opened 2023-02-07
[Spark][Bug]: Insert overwrite select from view will throw exception 


hellojinsilei:
### What happened?

when executing SQL : Insert overwrite select, and source is a view, it will throw an exception. But it will not happen when source is an arctic table
![image](https://user-images.githubusercontent.com/52364847/217169962-9ffa0638-c841-4968-8af5-b773499df776.png)

### Affects Versions

0.4.1

### What engines are you seeing the problem on?

Spark

### How to reproduce

step1: create a view
step2: create arctic table
step3: insert overwrite arctic_table select * from view

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1079, opened 2023-02-08
[Arctic-1066][Spark]:Fix ordered write error when using MixedHive unkeyedTable

baiyangtx:

fix: #1066 

## Brief change log

- Set default value of write options `writer.distributed-and-ordered` to false for the unkeyed table.
- Set the right value of FileIndexBucketMask
- Do optimize write for Batch Insert 

## How was this patch tested?

## Documentation


zhoujinsong (commented):
This change looks fine to me.
But we may need add some comment for `OptimizeWriteRule` to describe how we deal with all operations and why.

zhoujinsong (approved):

PR #1087, opened 2023-02-08
[ARCTIC-985] avoid NoSuchObjectException when renew blocker

wangtaohz:

avoid NoSuchObjectException when renew blocker, like 
image

## Brief change log

  - check blocker exist before renewing blocker
  - MockArcticMetastoreServer support renewBlocker

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/table/blocker/RenewableBlocker.java
+++ b/core/src/main/java/com/netease/arctic/table/blocker/RenewableBlocker.java
@@ -28,10 +28,12 @@ import org.slf4j.LoggerFactory;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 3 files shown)


wangtaohz:
cc @hellojinsilei

core/src/main/java/com/netease/arctic/table/blocker/RenewableBlocker.java
```diff
@@ -93,10 +95,16 @@ public void renewAsync() {
 
   private void doRenew() {
     try {
+      Set<String> validBlockers = amsClient.getBlockers(tableIdentifier.buildTableIdentifier())
```
zhoujinsong, line 98: Why this change is needed?
The new implementation require two rpc each time when renew, as the old implementation looks fine to me.

wangtaohz: The old implementation brings annoying error logs both in the spark client and AMS. I can change the log level to avoid part of these error logs.

wangtaohz: I found all these annoying errors could be avoided by using appropriate log level, I will change the code.

zhoujinsong (approved):

Issue #1060, opened 2023-02-07
[Optimizing][Improvement]: Clear optimizing tasks if task failed to execute after retry


wangtaohz:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

If the optimizing task failed to execute and is not recoverable, the optimizing process will be blocked and the tasks need to be cleared from sysdb, which is tedious and error-prone.

### How should we improve?

Clear these unrecoverable optimizing tasks automatically.

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1104, opened 2023-02-14
[ARCTIC-1103][AMS] Clear the optimizing tasks when the optimizing commit failed

### LINKED ISSUE #1103: [Improvement][AMS]: Automatic retry optimizing if commit failed
### Search before asking

Now, the optimizing will be hanged when the optimizing commit failed.

### How should we improve?

We can clear the task and automatic retry optimizing rather than waiting for the manual operation.

### Subtasks

_No response_

### Code of Conduct

hzluting:
fix #1103 

## Brief change log

  - *Clear the optimizing tasks when the optimizing commit failed*
  - *Clear the optimizing tasks when the table has failed tasks after max retry*

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
@@ -173,14 +173,14 @@ public class OptimizeService extends IJDBCService implements IOptimizeService {
     TableOptimizeItem tableItem = cachedTables.remove(tableIdentifier);
     optimizeQueueService.release(tableIdentifier);
     try {
+      tableItem.optimizeTasksClear(false);
+    } catch (Throwable t) {
+      LOG.debug("failed to delete " + tableIdentifier + " optimize task, ignore", t);
+    }
+    try {
       deleteTableOptimizeRuntime(tableIdentifier);
     } catch (Throwable t) {
       LOG.debug("failed to delete  " + tableIdentifier + " runtime, ignore", t);
-    }
-    try {
-      tableItem.clearOptimizeTasks();
-    } catch (Throwable t) {
-      LOG.debug("failed to delete " + tableIdentifier + " optimize task, ignore", t);
     }
     try {
       deleteOptimizeRecord(tableIdentifier);
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
@@ -626,6 +626,10 @@ public class TableOptimizeItem extends IJDBCService {
           getMapper(sqlSession, TableOptimizeRuntimeMapper.class);
 
       try {
+        // reset snapshot id in table runtime, because the tasks were cleared rather than committed.
+        // So need to plan again.
+        tableOptimizeRuntime.setCurrentSnapshotId(TableOptimizeRuntime.INVALID_SNAPSHOT_ID);
+        tableOptimizeRuntime.setCurrentChangeSnapshotId(TableOptimizeRuntime.INVALID_SNAPSHOT_ID);
         // persist partition optimize time
         tableOptimizeRuntimeMapper.updateTableOptimizeRuntime(tableOptimizeRuntime);
       } catch (Throwable t) {
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
@@ -629,7 +629,7 @@ public class TableOptimizeItem extends IJDBCService {
         // persist partition optimize time
         tableOptimizeRuntimeMapper.updateTableOptimizeRuntime(tableOptimizeRuntime);
       } catch (Throwable t) {
-        LOG.warn("failed to persist tableOptimizeRuntime after commit, ignore. " + getTableIdentifier(), t);
+        LOG.warn("failed to persist tableOptimizeRuntime after commit failed, ignore. " + getTableIdentifier(), t);
         sqlSession.rollback(true);
       }
 
@@ -659,6 +659,8 @@ public class TableOptimizeItem extends IJDBCService {
 
       sqlSession.commit(true);
     }
+
+    updateTableOptimizeStatus();
   }
 
   private void optimizeTasksCommitted(BasicOptimizeCommit optimizeCommit,
@@ -940,6 +942,7 @@ public class TableOptimizeItem extends IJDBCService {
           long commitTime = System.currentTimeMillis();
           optimizeTasksCommitted(optimizeCommit, commitTime);
         } else {
+          LOG.warn("{} commit failed, clear optimize tasks", tableIdentifier);
           optimizeTasksClear(optimizeCommit);
         }
       } else {
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
@@ -674,6 +674,9 @@ public class TableOptimizeItem extends IJDBCService {
           getMapper(sqlSession, InternalTableFilesMapper.class);
       TableOptimizeRuntimeMapper tableOptimizeRuntimeMapper =
           getMapper(sqlSession, TableOptimizeRuntimeMapper.class);
+      TaskHistoryMapper taskHistoryMapper =
+          getMapper(sqlSession, TaskHistoryMapper.class);
+
 
       TableOptimizeRuntime oldTableRuntime = tableOptimizeRuntime.clone();
       try {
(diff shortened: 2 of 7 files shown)


wangtaohz:
After clearing the optimizing tasks, the `currentChangeSnapshotId` and `currentSnapshotId` in `TableOptimizeRuntime` should also be reset, or it will skip the next plan if the table does not change.

It's better to be fixed in this PR.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
```diff
@@ -673,10 +673,14 @@ private void optimizeTasksClear(BasicOptimizeCommit optimizeCommit) {
...
       try {
+        // reset snapshot id in table runtime, because the tasks were cleared rather than committed.
+        // So need to plan again.
+        tableOptimizeRuntime.setCurrentSnapshotId(TableOptimizeRuntime.INVALID_SNAPSHOT_ID);
+        tableOptimizeRuntime.setCurrentChangeSnapshotId(TableOptimizeRuntime.INVALID_SNAPSHOT_ID);
         // persist partition optimize time
         tableOptimizeRuntimeMapper.updateTableOptimizeRuntime(tableOptimizeRuntime);
       } catch (Throwable t) {
-        LOG.warn("failed to persist tableOptimizeRuntime after commit, ignore. " + getTableIdentifier(), t);
+        LOG.warn("failed to persist tableOptimizeRuntime after commit failed, ignore. " + getTableIdentifier(), t);
```
wangtaohz, line 683: There is a standard procedure to update tableOptimizeRuntime to keep the consistency of memory and sysdb.

We should stick to it.

The procedure is like
```
oldRuntime = tableOptimizeRuntime.clone();
try {
    // update tableOptimizeRuntime

} catch () {
    tableOptimizeRuntime.restore(oldRuntime);
    sqlSession.rollback(true);
}
```

BTW, there are too `failed` in the warn log.

hzluting: OK

wangtaohz (approved):
LGTM

hzluting:
Now, there are two main exceptions causing optimizing hang. These exceptions can continue running if we clear the tasks.
1. Commit exception
- Missing required files to delete
- Found new delete for replaced data file
- For the unKeyed table, the file cache isn't the latest, handling the hive partition in the commit will throw an exception
- It took a long time to complete the task, the snapshot with the plan expired when committing

2. Execute exception
- If the task executes timeout, we modify the property named "self-optimizing.execute.timeout", it can continue running
- If the task plan with incorrect file info, causing the task throws the exception "file not found"

cc @wangtaohz @shendanfengg
Do you have any new situations to add?

wangtaohz:
> 2.Execute exception
If the task executes timeout, we modify the property named "self-optimizing.execute.timeout", it can continue running If the task plan with incorrect file info, causing the task throws the exception "file not found"

I support it as a default behavior to clean all tasks after any task fails(after retry for several times).

Even if it can't recover automatically after cleaning and planning, it's better than keeping hang up.

And this issue https://github.com/NetEase/arctic/issues/1060 could be fixed here.

hzluting:
> > 2.Execute exception
> > If the task executes timeout, we modify the property named "self-optimizing.execute.timeout", it can continue running If the task plan with incorrect file info, causing the task throws the exception "file not found"
> 
> I support it as a default behavior to clean all tasks after any task fails(after retry for several times).
> 
> Even if it can't recover automatically after cleaning and planning, it's better than keeping hang up.
> 
> And this issue #1060 could be fixed here.

OK

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
```diff
@@ -661,9 +662,9 @@ public void addNewOptimizeTasks(List<BasicOptimizeTask> newOptimizeTasks)
     }
   }
 
-  private void optimizeTasksClear(BasicOptimizeCommit optimizeCommit) {
+  public void optimizeTasksClear(Boolean refreshOptimizeStatus) {
```
wangtaohz, line 665: Now, `optimizeTasksClear` also clears tasks in `Executing` status, we have to handle the `optimize_task_history` of these tasks, or it will affect the quota occupation of this table.

BTW, `boolean` may be better than`Boolean`.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
```diff
@@ -173,14 +173,14 @@ private void clearTableCache(TableIdentifier tableIdentifier) {
...
     optimizeQueueService.release(tableIdentifier);
     try {
-      deleteTableOptimizeRuntime(tableIdentifier);
+      tableItem.optimizeTasksClear(false);
     } catch (Throwable t) {
-      LOG.debug("failed to delete  " + tableIdentifier + " runtime, ignore", t);
+      LOG.debug("failed to delete " + tableIdentifier + " optimize task, ignore", t);
     }
     try {
-      tableItem.clearOptimizeTasks();
```
wangtaohz, line 181: There are two similar methods in `TableOptimizeItem`:
- clearOptimizeTasks
- optimizeTasksClear

I think we should remove one of them.

wangtaohz (commented):
@hzluting I add some comments, plesase take a look.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
hzluting: OK

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/OptimizeService.java
hzluting: OK

XBaith:
> 2. Execute exception
> 
> * If the task executes timeout, we modify the property named "self-optimizing.execute.timeout", it can continue running
> * If the task plan with incorrect file info, causing the task throws the exception "file not found"
> 
> cc @wangtaohz @shendanfengg Do you have any new situations to add?
* HDFS cluster network issue also lead to failure to obtain file info

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
```diff
@@ -794,13 +798,22 @@ private void optimizeTasksCommitted(BasicOptimizeCommit optimizeCommit,
...
+        long endTime = System.currentTimeMillis();
         tasks.values().stream().flatMap(Collection::stream).map(OptimizeTaskItem::getTaskId)
             .forEach(optimizeTaskId -> {
               OptimizeTaskItem removed = optimizeTasks.remove(optimizeTaskId);
               if (removed != null) {
                 removedList.add(removed);
                 optimizeTasksMapper.deleteOptimizeTask(optimizeTaskId.getTraceId());
                 internalTableFilesMapper.deleteOptimizeTaskFile(optimizeTaskId);
+
+                // set end time and cost time in optimize_task_history
```
wangtaohz, line 810: Why should we set the cost time after committing?

hzluting: It should be in the optimizeTasksClear method.

XBaith:
> Now, there are two main exceptions causing optimizing hang. These exceptions can continue running if we clear the tasks.
> 
> 1. Commit exception
> 
> 2. Execute exception
> 
> cc @wangtaohz @shendanfengg Do you have any new situations to add?
`org.apache.hadoop.security.AccessControlException: Permission denied` exception can occur in both cases.

wangtaohz (approved):
LGTM

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
```diff
@@ -922,6 +928,27 @@ public void checkTaskExecuteTimeout() {
...
 
+  /**
+   * If task execute failed after max retry, clear tasks.
+   */
+  public void clearFailedTasks() {
+    tasksLock.lock();
+    try {
+      int maxRetry = optimizeMaxRetry();
+      Optional<OptimizeTaskItem> failedTask =
+          optimizeTasks.values().stream().findAny().filter(task -> task.getOptimizeRuntime().getRetry() > maxRetry);
```
zhongqishang, line 939: @hzluting 
- Should `findAny()` be placed at the end
- Add filter `OptimizeStatus.Failed == task.getOptimizeRuntime().getStatus()`

What do you think?

zhongqishang (commented):
Sorry for the late comment.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/TableOptimizeItem.java
hzluting: Yes, thank you for your comment

hzluting: @zhongqishang I fix it in #1168

PR #1115, opened 2023-02-14
[Hotfix] 0.3.x fix concurrent plan in different optimizing group

wangtaohz:

When changing the optimizing group, it may plan concurrently in different groups.

## Brief change log

  - refresh and check optimizing group is correct before optimize

## How was this patch tested?

## Documentation


wangtaohz:
Also, we have to avoid concurrent commits with this https://github.com/NetEase/arctic/issues/1124

PR #1122, opened 2023-02-16
[ARCTIC-1105][AMS] Fix snapshot expire and orphan clean misdelete the DataFiles 

### LINKED ISSUE #1105: [BUG]: The expire snapshot and the orphan file clean scan related files from metadata
### Description

The snapshot expires and the TTL scans files from metadata with TableScan API.
The orphan file clean will check whether the files in the change store are also indexed by the base store.

### Use case/motivation

Now, the expire clean checks useless files by scan files from the file info cache, making the data consistency rely too much on the correctness of the file info cache.
The orphan file clean doesn't consider the same file in the change store and the base store.

### Describe the solution

Load MetadataTableType.ALL_MANIFESTS to get all the content file entries in manifest files, like what org.apache.iceberg.spark.actions.BaseSparkAction#buildValidDataFileDF does.

### Subtasks

_No response_

### Related issues

#883 

### Code of Conduct

hzluting:
fix #1105 

## Brief change log

  - *The snapshot expire check valid DataFiles according to scan from metadata files*
  - *The orphan file clean check valid DataFiles according to scan from metadata files*

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ContentFileUtil.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ContentFileUtil.java
@@ -49,7 +49,6 @@ import java.util.Set;
  * Tools for handling the ContentFile which in Iceberg
  */
 public class ContentFileUtil {
-  private static final Logger LOG = LoggerFactory.getLogger(ContentFileUtil.class);
 
   public static ContentFileWithSequence<?> buildContentFile(DataFileInfo dataFileInfo,
                                                 PartitionSpec partitionSpec,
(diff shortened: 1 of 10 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ContentFileUtil.java
```diff
@@ -73,4 +95,32 @@ public static ContentFileWithSequence<?> buildContentFile(DataFileInfo dataFileI
 
     return WrapFileWithSequenceNumberHelper.wrap(contentFile, dataFileInfo.sequence);
   }
+
+  public static Set<String> getAllContentFilePath(UnkeyedTable innerTable) {
```
wangtaohz, line 99: We should append some unit test cases for this new method.

wangtaohz:
The unit test cases for orphan files clean in `TestOrphanFileClean` are too simple, and the cases where files are shared by ChangeStore and BaseStore are not covered. 

We should add more test cases to ensure it meets our expectations.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/HiveLocationUtils.java
```diff
@@ -45,7 +46,7 @@ public static Set<String> getHiveLocation(ArcticTable table) {
         try {
           Table hiveTable = ((SupportHive) table).getHMSClient().run(client ->
               client.getTable(table.id().getDatabase(), table.id().getTableName()));
-          hiveLocations.add(hiveTable.getSd().getLocation());
+          hiveLocations.add(TableFileUtils.getUriPath(hiveTable.getSd().getLocation()));
```
wangtaohz, line 49: Why change to use `TableFileUtils.getUriPath()`, does it bring problems before the change?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ContentFileUtil.java
```diff
@@ -22,12 +22,34 @@
...
+import org.slf4j.LoggerFactory;
 
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Tools for handling the ContentFile which in Iceberg
+ */
 public class ContentFileUtil {
```
wangtaohz, line 51: It may not be a good idea to put `getAllContentFilePath()` into `ContentFileUtil`.

`UnKeyedTableUtil` may be a better choice, because it seems to be an operation on `Table` as opposed to `File`.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/HiveLocationUtils.java
hzluting: Because all file locations use TableFileUtils.getUriPath() to remove the URI, I think the Hive location also should use to avoid potential risks.

hzluting:
> 

OK, I will add some cases.

wangtaohz (approved):

wangtaohz:
LGTM

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ContentFileUtil.java
```diff
@@ -22,11 +22,32 @@
...
 import org.apache.iceberg.ContentFile;
+import org.apache.iceberg.DataFile;
 import org.apache.iceberg.DataFiles;
 import org.apache.iceberg.FileMetadata;
+import org.apache.iceberg.HasTableOperations;
+import org.apache.iceberg.MetadataTableType;
+import org.apache.iceberg.MetadataTableUtils;
 import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.Table;
+import org.apache.iceberg.data.GenericRecord;
```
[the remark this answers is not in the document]
hzluting: OK

Issue #1093, opened 2023-02-09
[Feature]: Self-Optimizing scan files from metadata instead of from file info cache


wangtaohz:
### Description

Self-Optimizing scan files from metadata with `TableScan` API.

### Use case/motivation

Now, Self-Optimizing of `KeyedTable` and `UnkeyedTable` scan files from the file info cache, making the data consistency rely too much on the correctness of the file info cache, and affecting the stability of Self-Optimizing.

### Describe the solution

For `KeyedTable`, use `Table.newScan()` API to get all data files and delete files.

For `ArcticKeyed`, use `KeyedTable.newScan()` API to get all insert files, delete files and base files.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1125, opened 2023-02-16
[ARTIC-1903] fix Minor Optimizing scans files with iceberg sequence and add more test cases for hive support table

wangtaohz:

hot-fix for #1093 

## Brief change log

  - add planFilesWithSequence for ChangeTableIncrementalScan
  - add task property `move-files-to-hive-location` for hive support major optimize to avoid move files to hive location unexpectedly
  - fix hive support major optimize keeping optimizing with 1 small file
  - fix expireTaskHistory NPE
  - refactor OptimizingIntegrationTest and add more test case for hive support table
  - remove misuse of org.glassfish.jersey.internal.guava.*

## How was this patch tested?

## Documentation


ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/AbstractArcticOptimizePlan.java
```diff
@@ -208,38 +212,37 @@ protected void addChangeFilesIntoFileTree() {
...
+      markChangeSequence(partition, file.getSequenceNumber());
 
       addCnt.getAndIncrement();
     });
     LOG.debug("{} ==== {} add {} change files into tree, total files: {}." + " After added, partition cnt of tree: {}",
         tableId(), getOptimizeType(), addCnt, unOptimizedChangeFiles.size(), partitionFileTree.size());
   }
 
-  private long getMaxTransactionIdLimit(List<PrimaryKeyedFile> unOptimizedChangeFiles) {
+  private long getMaxSequenceLimit(List<ContentFileWithSequence<?>> unOptimizedChangeFiles) {
```
zhoujinsong, line 245: Minor optimize change files with the order of the change file sequence will result in changes files with larger transaction Id are merged before changes files with smaller transaction Id when streaming and batch writing occur simultaneously.

How should we solve this problem?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BasicOptimizeCommit.java
```diff
@@ -228,6 +228,7 @@ private void minorCommit(ArcticTable arcticTable,
       AtomicInteger addedPosDeleteFile = new AtomicInteger(0);
       StructLikeMap<Long> partitionOptimizedSequence =
           TablePropertyUtil.getPartitionOptimizedSequence(arcticTable.asKeyedTable());
+      AtomicBoolean shouldCommit = new AtomicBoolean();
```
zhoujinsong, line 231: Maybe we can support `skipEmptyCommit` method in `OverwriteBaseFiles`.

wangtaohz: It seems to be more reasonable. I will change it.

zhoujinsong (approved):
LGTM.

PR #1149, opened 2023-02-23
[ARCTIC-1148] Confirm the file cache is latest when optimize plan for the unKeyed table support hive

### LINKED ISSUE #1148: [Bug][Optimize]: Optimizing commit failed
### What happened?

For the unKeyed partition table support hive, the optimizing commit failed as below.
image

### Affects Versions

0.3.0

Optimizer

### How to reproduce

1. create an unKeyed partition table support hive
2. insert files into the table
3. execute a major optimize, and the produced files not in the file cache
4. execute a full optimize

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hzluting:
fix #1148 

## Brief change log

  - *Confirm the file cache is latest when optimize plan for the unKeyed table support hive*

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BaseOptimizePlan.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BaseOptimizePlan.java
@@ -38,7 +38,6 @@ import com.netease.arctic.table.TableProperties;
 import com.netease.arctic.utils.SerializationUtil;
 import org.apache.iceberg.DataFile;
 import org.apache.iceberg.DeleteFile;
-import org.apache.iceberg.Snapshot;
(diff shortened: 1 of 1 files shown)


zhoujinsong:
@hzluting Is this bug fixed in master? If yes, can you add the PR?

hzluting:
> @hzluting Is this bug fixed in master? If yes, can you add the PR?

No, because Self-Optimizing scans files from metadata instead of the file info cache in the master(#1100).
We don't need to consider the file cache.

hzluting:
This condition is too strict for the unKeyed arctic table because the Flink will stream-produce the data into the base store, and it will cause the file cache of the base store is always not the latest.

PR #1151, opened 2023-02-23
[ARCTIC-1150] Minor optimizing with many small files and TransactionId not in order

### LINKED ISSUE #1150: [Bug]: Minor optimizing breaks data consistency with many small files and TransactionId not in order
### What happened?

When there are many files in the change store, these files will split into several batches to be Minor optimized.
If the transaction Ids of these files are not in order, and they affect the same record, the Minor optimizing may break the data consistency, because it doesn't optimize these files in the order of transaction id.

### Affects Versions

master/0.4.0/

_No response_

### How to reproduce

I have write a test case here `com.netease.arctic.ams.server.optimize.MixedIcebergOptimizingTest#testKeyedTableTxIdNotInOrder`.

1.create Arctic Keyed table
2.stop self-optimizing
3.set `self-optimizing.max-file-count` = 5
4.follows the write order in the test case, to write some data with Transaction Ids not in order
5.open self-optimizing
6.after optimizing, we will find duplicate id 2 by MOR

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

wangtaohz:

fix #1150 

## Brief change log

  - ChangeTableIncrementalScan support `toSequence`, corresponding to `fromSequence`
  - Minor optimizing use `formSequence` and `toSequence` to limit the file cnt it scans
  - Minor optimizing scan all change files grouped by snapshot and calculate the max sequence for the `toSequence`

## How was this patch tested?

## Documentation


ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ChangeFilesUtil.java
```diff
@@ -72,4 +73,114 @@ private static void deleteChangeFiles(KeyedTable keyedTable, List<DataFile> chan
...
+    for (int i = 0; i < snapshotFileGroups.size(); i++) {
+      fileCnt += snapshotFileGroups.get(i).getFileCnt();
+      if (fileCnt <= maxFileCntLimit) {
+        index = i;
+      } else {
+        break;
+      }
+    }
+    // all files cnt <= maxFileCntLimit, return all files
+    if (fileCnt <= maxFileCntLimit) {
```
zhoujinsong, line 106: `fileCnt` seems to be always smaller than `maxFileCntLimit ` according to line 99 to 103.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ChangeFilesUtil.java
```diff
@@ -72,4 +73,114 @@ private static void deleteChangeFiles(KeyedTable keyedTable, List<DataFile> chan
...
+   * Selected-Files as possible, and also
+   * - the cnt of these Selected-Files must <= maxFileCntLimit
+   * - the max TransactionId of the Selected-Files must > the min TransactionId of all the left files
+   *
+   * @param snapshotFileGroups snapshotFileGroups
+   * @param maxFileCntLimit    maxFileCntLimit
+   * @return the max sequence of selected file, return Long.MAX_VALUE if all files should be selected,
+   * Long.MIN_VALUE means no files should be selected
+   */
+  public static long getMaxSequenceLimit(List<SnapshotFileGroup> snapshotFileGroups, long maxFileCntLimit) {
```
zhoujinsong, line 88: I don't think this is a good util method as it is hard to reuse.
Maybe we can abstract the change store file planning process to a util method like `planOptimizingFiles`.

core/src/main/java/com/netease/arctic/data/file/FileNameGenerator.java
```diff
@@ -223,6 +223,23 @@ public static long parseTransactionId(String path) {
     }
   }
 
+  /**
+   * Parse transaction id of change file.
+   *
+   * @param path path
+   * @return transactionId, return 0 if path is not arctic file format.
+   */
+  public static long parseChangeTransactionId(String path, long sequenceNumber) {
```
zhoujinsong, line 232: Can we implement this method dependent on `parseTransactionId`?

wangtaohz: Ok, I will change it.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ChangeFilesUtil.java
wangtaohz: This method is not a good abstraction. I will refactor it.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ChangeFilesUtil.java
wangtaohz:
```
    // 2.find the max index where all file cnt <= maxFileCntLimit
    int index = -1;
    int fileCnt = 0;
    for (int i = 0; i < snapshotFileGroups.size(); i++) {
      fileCnt += snapshotFileGroups.get(i).getFileCnt();
      if (fileCnt <= maxFileCntLimit) {
        index = i;
      } else {
        break;
      }
    }
    // all files cnt <= maxFileCntLimit, return all files
    if (fileCnt <= maxFileCntLimit) {
      return Long.MAX_VALUE;
    }
```

According to the complete code, when `fileCnt > maxFileCntLimit`, it will break the traversal, so `fileCnt` is not always smaller than `maxFileCntLimit`.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/model/SnapshotFileGroup.java
```diff
@@ -0,0 +1,60 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netease.arctic.ams.server.model;
+
+/**
+ * Files grouped by snapshot, but only with the file cnt.
+ */
+public class SnapshotFileGroup implements Comparable<SnapshotFileGroup> {
```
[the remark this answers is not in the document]
wangtaohz: Ok. I moved `SnapshotFileGroup ` to `ChangeFilesUtil` as a static inner class.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/ChangeFilesUtil.java
wangtaohz: I moved this method from `ChangeFilesUtils` to `TableOptimizeItem`, corresponding to scan base files.

zhoujinsong (approved):

PR #1156, opened 2023-02-24
[ARCTIC-1153][Optimize] Put task back into the queue if polling tasks failed

### LINKED ISSUE #1153: [Bug][Optimize]: Optimize task status is always pending
### What happened?

The optimizing task status is always pending after the poll task failed.
image

### Affects Versions

0.3.0

AMS

### How to reproduce
1. AMS running on sysdb mysql
2. Then make AMS failing to connect to mysql
3. Start an optimizer to poll task from AMS
4. Task's status is still Pending, but can't be executed by optimizer, unless restart AMS.

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hzluting:
fix #1153 

## Brief change log

  - *Put task back into the queue if polling tasks failed*

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
@@ -570,10 +570,17 @@ public class OptimizeQueueService extends IJDBCService {
               tableTaskHistory = task.onExecuting(jobId, attemptId);
             } catch (Exception e) {
               task.clearFiles();
-              LOG.error("{} failed to load files from sysdb, try put task back into queue", task.getTaskId(), e);
+              LOG.error("{} handle sysdb failed, try put task back into queue", task.getTaskId(), e);
               if (!tasks.offer(task)) {
                 LOG.error("{} failed to put task back into queue", task.getTaskId());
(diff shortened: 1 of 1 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -561,19 +561,22 @@ public OptimizeTask poll(JobId jobId, final String attemptId, long waitTime) {
...
+              task.setMaxExecuteTime();
+              tableTaskHistory = task.onExecuting(jobId, attemptId);
             } catch (Exception e) {
               task.clearFiles();
               LOG.error("{} failed to load files from sysdb, try put task back into queue", task.getTaskId(), e);
               if (!tasks.offer(task)) {
+                LOG.error("{} failed to put task back into queue", task.getTaskId());
                 task.onFailed(new ErrorMessage(System.currentTimeMillis(), "failed to put task back into queue"), 0);
               }
+              continue;
```
wangtaohz, line 578: We should sleep for like 1 second, before `continue` (poll from the tasks again).

hzluting: OK

wangtaohz (approved):
LGTM

PR #1168, opened 2023-02-27
[Arctic-1103][AMS] fix the condition of checking the tasks whether they can be clear (improve #1104)

hzluting:
improve #1104

## Brief change log

  - *Modify the condition of checking the tasks whether they can be clear*
  - *Add some related unit tests*

## How was this patch tested?

## Documentation


wangtaohz (approved):
LGTM

PR #1178, opened 2023-03-01
[ARCTIC-930][AMS] Snapshot expire service should expire snapshot with the lastest checkpoint id

### LINKED ISSUE #930: [Bug]: Flink cannot restore from old state if the snapshot with max commited ckp id info lost
### What happened?

This problem is similar to [Iceberg-4137](https://github.com/apache/iceberg/issues/4137) .
The old state means that there are some successful ckps after that and it has been committed.

If the max committed checkpoint-id is lost due to snapshot expired and removed by Optimizer, the job may not be able to restore from old state because that the max committed checkpoint-id is lost, then flink would commit the pending Manifest again(refer to https://github.com/apache/iceberg/blob/master/flink/v1.14/flink/src/main/java/org/apache/iceberg/flink/sink/IcebergFilesCommitter.java#L184), but the pending manifests had been removed because they were committed successfully before.

### Affects Versions

master

_No response_

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

wangtaohz:

To fix #930 

## Brief change log
  - the last snapshot with `flink.max-committed-checkpoint-id` should not be expired

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
@@ -45,6 +45,8 @@ import org.apache.hadoop.fs.Path;
 import org.apache.iceberg.ContentFile;
 import org.apache.iceberg.DataFile;
 import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.util.StructLikeMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -137,7 +139,8 @@ public class TableExpireService implements ITableExpireService {
             // in the base store
             Set<String> baseExclude = UnKeyedTableUtil.getAllContentFilePath(changeTable);
             baseExclude.addAll(finalHiveLocation);
-            expireSnapshots(baseTable, startTime - baseSnapshotsKeepTime, baseExclude);
+            long baseOlderThan = getExpireTime(baseTable, startTime - baseSnapshotsKeepTime);
+            expireSnapshots(baseTable, baseOlderThan, baseExclude);
             long baseCleanedTime = System.currentTimeMillis();
             LOG.info("[{}] {} base expire cost {} ms", traceId, arcticTable.id(), baseCleanedTime - startTime);
 
@@ -151,14 +154,16 @@ public class TableExpireService implements ITableExpireService {
             // in the change store
             Set<String> changeExclude = UnKeyedTableUtil.getAllContentFilePath(baseTable);
             changeExclude.addAll(finalHiveLocation);
-            expireSnapshots(changeTable, startTime - changeSnapshotsKeepTime, changeExclude);
+            long changeOlderThan = getExpireTime(changeTable, startTime - changeSnapshotsKeepTime);
+            expireSnapshots(changeTable, changeOlderThan, changeExclude);
             return null;
           });
           LOG.info("[{}] {} expire cost total {} ms", traceId, arcticTable.id(),
               System.currentTimeMillis() - startTime);
         } else {
           UnkeyedTable unKeyedArcticTable = arcticTable.asUnkeyedTable();
-          expireSnapshots(unKeyedArcticTable, startTime - baseSnapshotsKeepTime, hiveLocation);
+          long baseOlderThan = getExpireTime(unKeyedArcticTable, startTime - baseSnapshotsKeepTime);
+          expireSnapshots(unKeyedArcticTable, baseOlderThan, hiveLocation);
           long baseCleanedTime = System.currentTimeMillis();
           LOG.info("[{}] {} unKeyedTable expire cost {} ms", traceId, arcticTable.id(), baseCleanedTime - startTime);
         }
@@ -166,6 +171,17 @@ public class TableExpireService implements ITableExpireService {
         LOG.error("[" + traceId + "] unexpected expire error of table " + tableIdentifier, t);
       }
     }
+  }
+
+  public static long getExpireTime(UnkeyedTable table, long olderThan) {
+    ArrayList<Snapshot> snapshots = Lists.newArrayList(table.snapshots());
+    for (int i = snapshots.size() - 1; i >= 0; i--) {
+      Snapshot snapshot = snapshots.get(i);
+      if (snapshot.summary().containsKey("flink.max-committed-checkpoint-id")) {
+        return Math.min(snapshot.timestampMillis(), olderThan);
+      }
+    }
+    return olderThan;
   }
 
   public static void deleteChangeFile(KeyedTable keyedTable, List<DataFileInfo> changeDataFiles) {
@@ -229,7 +245,7 @@ public class TableExpireService implements ITableExpireService {
   public static void expireSnapshots(UnkeyedTable arcticInternalTable,
                                      long olderThan,
                                      Set<String> exclude) {
-    LOG.debug("start expire snapshots, the exclude is {}", exclude);
+    LOG.debug("start expire snapshots older than {}, the exclude is {}", olderThan, exclude);
     final AtomicInteger toDeleteFiles = new AtomicInteger(0);
     final AtomicInteger deleteFiles = new AtomicInteger(0);
     Set<String> parentDirectory = new HashSet<>();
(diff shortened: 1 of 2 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
```diff
@@ -168,6 +173,17 @@ public void run() {
     }
   }
 
+  public static long getExpireTime(UnkeyedTable table, long olderThan) {
+    ArrayList<Snapshot> snapshots = Lists.newArrayList(table.snapshots());
```
YesOrNo828, line 177: How about using Iterator to iterate the snasphots? it could reduce the time complex to `n`, but now it is `n to 2n`

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
```diff
@@ -168,6 +173,17 @@ public void run() {
     }
   }
 
+  public static long getExpireTime(UnkeyedTable table, long olderThan) {
```
YesOrNo828, line 176: How about renaming the method name to `fetchLastestFlinkCommitedSnapshotTime`?

YesOrNo828 (commented):
@wangtaohz  Thanks for your contribution. I left some comments.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
wangtaohz: It makes sense. I will modify it.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
```diff
@@ -103,69 +101,94 @@ public void run() {
...
-              return null;
-            }
-            UnkeyedTable changeTable = keyedArcticTable.changeTable();
-            if (changeTable == null) {
-              LOG.warn("[{}] Change table is null: {}", traceId, tableIdentifier);
-              return null;
-            }
+  public static void expireArcticTable(ArcticTable arcticTable) {
+    TableIdentifier tableIdentifier = arcticTable.id();
+    final String traceId = UUID.randomUUID().toString();
```
zhoujinsong, line 113: What is the usage of this `traceId`?
It looks as if it is used to correlate related content when troubleshooting logs, and if so whether the table name would serve the same purpose.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
```diff
@@ -103,69 +101,94 @@ public void run() {
...
-            return null;
-          });
-          LOG.info("[{}] {} expire cost total {} ms", traceId, arcticTable.id(),
-              System.currentTimeMillis() - startTime);
-        } else {
-          UnkeyedTable unKeyedArcticTable = arcticTable.asUnkeyedTable();
-          expireSnapshots(unKeyedArcticTable, startTime - baseSnapshotsKeepTime, hiveLocation);
-          long baseCleanedTime = System.currentTimeMillis();
-          LOG.info("[{}] {} unKeyedTable expire cost {} ms", traceId, arcticTable.id(), baseCleanedTime - startTime);
+    Set<String> hiveLocation = new HashSet<>();
```
zhoujinsong, line 127: Change parameter name to `hiveLocations` or `hiveLocationSet`?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
```diff
@@ -103,69 +101,94 @@ public void run() {
...
+    if (TableTypeUtil.isHive(arcticTable)) {
+      hiveLocation = HiveLocationUtils.getHiveLocation(arcticTable);
+    }
+
+    if (arcticTable.isKeyedTable()) {
+      KeyedTable keyedArcticTable = arcticTable.asKeyedTable();
+      Set<String> finalHiveLocation = hiveLocation;
+      keyedArcticTable.io().doAs(() -> {
+        UnkeyedTable baseTable = keyedArcticTable.baseTable();
+        if (baseTable == null) {
```
zhoujinsong, line 137: When will a table store be null?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
```diff
@@ -103,69 +101,94 @@ public void run() {
...
-        LOG.error("[" + traceId + "] unexpected expire error of table " + tableIdentifier, t);
+        UnkeyedTable changeTable = keyedArcticTable.changeTable();
+        if (changeTable == null) {
+          LOG.warn("[{}] Change table is null: {}", traceId, tableIdentifier);
+          return null;
+        }
+
+        // get valid files in the change store which shouldn't physically delete when expire the snapshot
+        // in the base store
+        Set<String> baseExclude = UnKeyedTableUtil.getAllContentFilePath(changeTable);
```
zhoujinsong, line 149: Change parameter to `baseExcludePaths` or `baseExcludePathSet`?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
```diff
@@ -103,69 +101,94 @@ public void run() {
...
+        baseExclude.addAll(finalHiveLocation);
+        long latestBaseFlinkCommitTime = fetchLatestFlinkCommittedSnapshotTime(baseTable);
```
zhoujinsong, line 151: Can you add some comment help other coder understand your logic here?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
```diff
@@ -103,69 +101,94 @@ public void run() {
...
+          hiveLocation);
+      long baseCleanedTime = System.currentTimeMillis();
+      LOG.info("[{}] {} unKeyedTable expire cost {} ms", traceId, arcticTable.id(), baseCleanedTime - startTime);
+    }
+  }
+
+  public static long fetchLatestFlinkCommittedSnapshotTime(UnkeyedTable table) {
+    long latestCommitTime = Long.MAX_VALUE;
+    for (Snapshot snapshot : table.snapshots()) {
+      if (snapshot.summary().containsKey("flink.max-committed-checkpoint-id")) {
```
zhoujinsong, line 187: I suggest using a static variable to hold this value and link it to the iceberg flink writer class in the comment.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
wangtaohz: `traceId` is useless now, and should be removed.

It was used to trace a http request before, but now the orphan files clean can't be triggered from http request.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
wangtaohz: I changed it to `hiveLocations`

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
wangtaohz: `ChangeTable` and `BaseTableIt` of `KeyedTable` will never be null, I will remove the check.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
wangtaohz: I changed it to `baseExcludePaths`.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
wangtaohz: OK, I added some comment for `fetchLatestFlinkCommittedSnapshotTime`.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/TableExpireService.java
wangtaohz: I import a static variable for this constant and add some comment linking to `org.apache.iceberg.flink.sink.IcebergFilesCommitter#MAX_COMMITTED_CHECKPOINT_ID`, but the `{@link xxx}` doesn't work here because it is inaccessible.

zhoujinsong (approved):
LGTM.

PR #1180, opened 2023-03-01
[ARCTIC-1025][FLINK] Fix data duplication even if there is a primary key table with upsert enabled

### LINKED ISSUE #1025: [Bug]: The same _file_offset in the Change Store causes data duplication even if there is a primary key table with upsert enabled
### What happened?

For tables with a primary key and the property 'write.upsert.enabled' set, when the same transaction id commits multiple changes to a primary key data, it can lead to data duplication. And the '_file_offset' of these duplicate data must be duplicated.
image

### Affects Versions

0.3.x

Flink

### How to reproduce

1.Create an Arctic table with a primary key and start a Flink ingestion job
2.The same primary key is used for Insert and Update operations during a single transaction commit

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

zstraw:
FIX #1025 

To make sure only one primary key in key table with upsert enabled, it's necessary to delete data firstly when update_after not followed by update_before.

It may lead to some unnecessary delete data if there are some data interspersed with other primary key data. e.g. K1-UB, K2-UB, K1-UA, K2-UB

## Brief change log

  - *The flink 1.12 and 1.14, 1.15 versions are affected.*

## How was this patch tested?

## Documentation

### DIFF
--- a/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
+++ b/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
@@ -79,7 +79,7 @@ public class ArcticFileWriter extends AbstractStreamOperator<WriteResult>
    */
   private transient ArcticTable table;
   /**
-   * Track whether there is update_before before update_after or not neglecting primary key.
+   * Track whether there is update_before before update_after or not neglecting primary key if upsert enabled.
    */
   private transient boolean hasUpdateBefore = false;
 
@@ -219,7 +219,7 @@ public class ArcticFileWriter extends AbstractStreamOperator<WriteResult>
 
   /**
    * If upsert is enabled, turn update_after to insert if there isn't update_after followed by update_before.
-   * It may lead to some unnecessary delete data if there are some data interspersed with other primary key data.
+   * But it may lead to some unnecessary delete data if there are some data interspersed with other primary key data.
    * e.g. K1-UB, K2-UB, K1-UA, K2-UB
    */
   private void processMultiUpdateAfter(RowData row) {
--- a/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
+++ b/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
@@ -19,6 +19,8 @@
 package com.netease.arctic.flink.write;
 
 import com.netease.arctic.data.DataTreeNode;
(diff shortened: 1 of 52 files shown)


flink/v1.12/flink/src/test/java/com/netease/arctic/flink/table/TestKeyed.java
```diff
@@ -441,12 +441,14 @@ public void testPartitionSinkFile() throws IOException {
   @Test
   public void testFileUpsert() throws IOException {
```
YesOrNo828, line 442: Could you add a UT when an upsert table functionality is disabled?

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
```diff
@@ -213,6 +217,22 @@ public void processElement(StreamRecord<RowData> element) throws Exception {
     });
   }
 
+  /**
+   * If upsert is enabled, turn update_after to insert if there isn't update_after followed by update_before.
+   * But it may lead to some unnecessary delete data if there are some data interspersed with other primary key data.
+   * e.g. K1-UB, K2-UB, K1-UA, K2-UB
+   */
+  private void processMultiUpdateAfter(RowData row) {
```
YesOrNo828, line 225: Maybe we should also consider this scenario:
Key1-UB, Key1-UA, Key2-UB, Key1-UA.

YesOrNo828 (commented):
@zstraw Thanks for your contribution, I left some comments.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
```diff
@@ -213,6 +217,22 @@ public void processElement(StreamRecord<RowData> element) throws Exception {
...
+    if (!upsert) {
```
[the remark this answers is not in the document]
zstraw: It makes sense to me.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/ArcticFileWriter.java
zstraw: I will track ua before in keys set to fix this case.

flink/v1.12/flink/src/test/java/com/netease/arctic/flink/table/TestKeyed.java
zstraw: Ok

PR #1191, opened 2023-03-03
[ARCTIC-1188][Optimizer] Flink optimizer retry reporting task result after reporting failed

### LINKED ISSUE #1188: [Improvement]: Flink Optimizer should retry reporting the task result after reporting failed
### Search before asking

When AMS is restarting, the Optimizer will fail to report the task result. 

However, Flink Optimizer doesn't retry reporting the task result, which makes the status of the task in AMS to be `Executing` until timeout(default 30min). So the `Self-Optimizing` will be blocked for at least 30min in this case.

### How should we improve?

Flink Optimizer should retry reporting the task result after reporting failed, to avoid blocking `Self-Optimizing`.

I think keeping retry for 5 min is reasonable (>30min is useless since the default timeout is 30min), so the default value is designed to be retrying 20 times with 15s intervals.

### Subtasks

_No response_

### Code of Conduct

wangtaohz:

resolve #1188 

## Brief change log

  -  optimizer makes retry reporting task result after reporting failed as a default behavior
  -  add `OperatorFactory` for the facilitation of unit test

## How was this patch tested?

## Documentation


zhoujinsong:
@wangtaohz Can you finish the pr description about the test and document part?

wangtaohz:
> @wangtaohz Can you finish the pr description about the test and document part?

OK, I added some descriptions.

zhoujinsong (approved):
LGTM.

PR #1192, opened 2023-03-03
[ARCTIC-1189][Optimizer] Optimizer reports a failed result to AMS when failed to load table

### LINKED ISSUE #1189: [Bug]: Optimizer doesn't report a failed result when failed to load table
### What happened?

When the Optimizer fails to load a table after polling a task and before executing it, we can see an error log in the optimizer, but AMS doesn't receive any report.

In this case, the task in AMS will be in `Executing` until timeout(default 30min), and the Self-Optimizing will be blocked for at least 30min.

### Affects Versions

master/0.3.0/,master/0.4.0/

Optimizer

### How to reproduce

We need an environment, which will fail to load ArcticTable, like having some network or authorization problem.

1. Start an optimizer in this environment, making it the only Optimizer of an optimizer group.
2. Create an Arctic Table and change its optimizer group.
3. Write some data into the ArcticTable
4. The ArcticTable will be in the status of `Optimizing` for several hours and we can find `Executing` optimize tasks from sysdb.

### Relevant log output

_No response_

### Anything else

This problem will occur only when Optimizer failed to load ArcticTable, and it will happen every time.

### Code of Conduct

wangtaohz:

fix #1189 

## Brief change log

  - Optimizer reports a failed result to AMS when failed to load the table
  - Add some unit tests for `BaseTaskExecutor`

## How was this patch tested?

## Documentation

### DIFF
--- a/optimizer/src/main/java/com/netease/arctic/optimizer/local/LocalOptimizer.java
+++ b/optimizer/src/main/java/com/netease/arctic/optimizer/local/LocalOptimizer.java
@@ -287,7 +287,7 @@ public class LocalOptimizer implements StatefulOptimizer {
           LOG.info("get task to execute {}", task.getTask().getTaskId());
           OptimizeTaskStat result = baseTaskExecutor.execute(task);
           LOG.info("execute {} {}", result.getStatus(), task.getTask().getTaskId());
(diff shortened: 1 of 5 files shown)


zhoujinsong (approved):
LGTM.

PR #1199, opened 2023-03-07
[ARCTIC-1197][Optimizer] Optimizer should wait for a while between each polling task request

### LINKED ISSUE #1197: [Improvement]: When `Optimizer` polls tasks from AMS, it occupies too many thrift worker threads in AMS
### Search before asking

Too many thrift worker threads will be occupied by `Optimizer` depending on the parallelism of `Optimizer`; because when a thread in `Optimizer` polls a task from AMS, it will occupy a thrift worker thread and wait for a while(10s) until getting a task or returning null when getting no task. Then, the thread in `Optimizer` will trigger another polling request with no delay.

### How should we improve?

The thrift worker thread should not wait for a while and return null immediately when getting no task. The `Optimizer` should sleep for a while before the next request when getting no task.

### Subtasks

_No response_

### Code of Conduct

wangtaohz:

fix #1197 

## Brief change log

  - the polling task request should return immediately
  - `Optimizer` should wait for a while between each polling task request

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
@@ -490,12 +490,9 @@ public class OptimizeQueueService extends IJDBCService {
         task = waitForTask(startTime, waitTime);
         if (task == null) {
           return null;
-        } else {
-          return onExecuteOptimizeTask(task, jobId, attemptId);
-        }
-      } else {
-        return onExecuteOptimizeTask(task, jobId, attemptId);
-      }
+        } 
+      } 
+      return onExecuteOptimizeTask(task, jobId, attemptId);
     }
 
     private OptimizeTask onExecuteOptimizeTask(OptimizeTaskItem task, JobId jobId, String attemptId) {
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
@@ -484,7 +484,7 @@ public class OptimizeQueueService extends IJDBCService {
 
     public OptimizeTask poll(JobId jobId, final String attemptId, long waitTime) {
       long startTime = System.currentTimeMillis();
-      OptimizeTaskItem task = tasks.poll();
+      OptimizeTaskItem task = pollValidTask();
       if (task == null) {
         tryPlanAsync(jobId, attemptId);
         task = waitForTask(startTime, waitTime);
@@ -520,20 +520,33 @@ public class OptimizeQueueService extends IJDBCService {
       return task.getOptimizeTask();
     }
 
+    private OptimizeTaskItem pollValidTask() {
+      while (true) {
+        OptimizeTaskItem task = tasks.poll();
+        if (task == null) {
+          return null;
+        }
+        if (tables.contains(task.getTableIdentifier())) {
+          return task;
+        } else {
+          LOG.warn("get task {} from queue {} but table {} not in this queue, continue",
+              task.getTaskId(), queueName(), task.getTableIdentifier());
+        }
+      }
+    }
+
     private OptimizeTaskItem waitForTask(long startTime, long waitTime) {
+      if (waitTime <= 0) {
+        return null;
+      }
       while (true) {
         lock();
         try {
           // if timeout, return null, await support zero or a negative value, and return false
           if (planThreadCondition.await(waitTime - (System.currentTimeMillis() - startTime),
               TimeUnit.MILLISECONDS)) {
-            OptimizeTaskItem task = tasks.poll();
+            OptimizeTaskItem task = pollValidTask();
             if (task != null) {
-              if (!tables.contains(task.getTableIdentifier())) {
-                LOG.warn("get task {} from queue {} but table {} not in this queue",
-                    task.getTaskId(), queueName(), task.getTableIdentifier());
-                continue;
-              }
               return task;
             }
           } else {
@@ -546,6 +559,7 @@ public class OptimizeQueueService extends IJDBCService {
         } finally {
           unlock();
         }
+        // check wait timeout
         long duration = System.currentTimeMillis() - startTime;
         if (duration > waitTime) {
           LOG.info("pool task cost too much time {} ms, return null", duration);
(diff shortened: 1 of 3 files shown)


optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkConsumer.java
```diff
@@ -67,7 +72,7 @@ public void run(SourceContext<TaskWrapper> sourceContext) throws Exception {
           retry = 0;
           LOG.error("flink source has tried too many times, and the subscription message is suspended." +
               " Please check for errors");
-          Thread.sleep(10000);
+          Thread.sleep(2 * POLL_INTERVAL);
```
zhoujinsong, line 75: Is it possible to wait uniformly if no task is obtained or if the task fails to be obtained?

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -484,11 +484,16 @@ public String queueName() {
...
       while (true) {
-        long duration = System.currentTimeMillis() - startTime;
-        if (duration > waitTime) {
-          LOG.warn("pool task cost too much time {} ms, return null", duration);
-          return null;
+        if (first) {
+          first = false;
+        } else {
+          long duration = System.currentTimeMillis() - startTime;
+          if (duration > waitTime) {
```
zhoujinsong, line 493: I suggest putting the timeout judgment at the end of the method. In addition, the current method body is too long. It is recommended to abstract some complex logic into private methods to shortage this method.

wangtaohz: OK, I will refactor it.

optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkConsumer.java
wangtaohz: Yes, I will modify it.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -484,117 +484,137 @@ public String queueName() {
...
+        tryPlanAsync(jobId, attemptId);
+        task = waitForTask(startTime, waitTime);
+        if (task == null) {
+          return null;
+        } else {
+          return onExecuteOptimizeTask(task, jobId, attemptId);
+        }
+      } else {
+        return onExecuteOptimizeTask(task, jobId, attemptId);
+      }
```
zhoujinsong, line 498:
```suggestion
      if (task == null) {
        tryPlanAsync(jobId, attemptId);
        task = waitForTask(startTime, waitTime);
        if (task == null) {
          return null;
        } 
      } 
      return onExecuteOptimizeTask(task, jobId, attemptId);
```

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -484,117 +484,137 @@ public String queueName() {
...
+      try {
+        insertTableTaskHistory(tableTaskHistory);
+      } catch (Exception e) {
+        LOG.error("failed to insert tableTaskHistory, {} ignore", tableTaskHistory, e);
+      }
+      return task.getOptimizeTask();
+    }
+
+    private OptimizeTaskItem waitForTask(long startTime, long waitTime) {
       while (true) {
```
zhoujinsong, line 527: When `waitTime` is 0, we should avoid doing unnecessary waits.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -484,117 +484,137 @@ public String queueName() {
...
+        lock();
+        try {
+          // if timeout, return null, await support zero or a negative value, and return false
+          if (planThreadCondition.await(waitTime - (System.currentTimeMillis() - startTime),
+              TimeUnit.MILLISECONDS)) {
+            OptimizeTaskItem task = tasks.poll();
+            if (task != null) {
+              if (!tables.contains(task.getTableIdentifier())) {
```
zhoujinsong, line 535: It seems that we not only need to check the legality of the task here, but also the task obtained at the beginning.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -484,117 +484,137 @@ public String queueName() {
...
+
+    private void tryPlanAsync(JobId jobId, String attemptId) {
+      if (!planThreadStarted.compareAndSet(false, true)) {
+        // a plan tread is working now
+        return;
+      }
+      try {
+        Thread planThread = new Thread(() -> {
+          int retry = 0;
+          boolean isHaveTask = false;
```
zhoujinsong, line 568: Is `planSuccess` better here as a parameter? 
Or we should just check the size of `tasks` instead of create a new parameter.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
```diff
@@ -484,117 +484,137 @@ public String queueName() {
...
-                if (!planThreadCondition.await(waitTime - (System.currentTimeMillis() - startTime),
-                    TimeUnit.MILLISECONDS)) {
-                  LOG.debug("The queue {} has no task have planned", optimizeQueue.getOptimizeQueueMeta().getQueueId());
-                  return null;
-                }
-              } finally {
-                unlock();
+                Thread.sleep(retryInterval);
+              } catch (InterruptedException e) {
+                LOG.error("Internal Thread Interrupted", e);
```
zhoujinsong, line 586: We may need to break from the loop when got an interrupted exception.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
wangtaohz: I created a method pollValidTask() to check legality while polling tasks.
```
    private OptimizeTaskItem pollValidTask() {
      while (true) {
        OptimizeTaskItem task = tasks.poll();
        if (task == null) {
          return null;
        }
        if (tables.contains(task.getTableIdentifier())) {
          return task;
        } else {
          LOG.warn("get task {} from queue {} but table {} not in this queue, continue",
              task.getTaskId(), queueName(), task.getTableIdentifier());
        }
      }
    }
```

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
wangtaohz: OK, I checked it the first step in `waitForTask()`.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
wangtaohz: I removed the flag `isHaveTask`.

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizeQueueService.java
wangtaohz: Yes, it should be breaked.

zhoujinsong (approved):
LGTM.

PR #1221, opened 2023-03-13
[ARCTIC-1025][FLINK] Fix data duplication even if there is a primary key table with upsert enabled

### LINKED ISSUE #1025: [Bug]: The same _file_offset in the Change Store causes data duplication even if there is a primary key table with upsert enabled

lklhdu:
FIX #1025 

To make sure only one primary key in key table with upsert enabled, it's necessary to delete data firstly when update_after not followed by update_before.

It may lead to some unnecessary delete data if there are some data interspersed with other primary key data. e.g. K1-UB, K2-UB, K1-UA, K2-UB
## Brief change log

  - The flink 1.12 and 1.14, 1.15 versions are affected.

## How was this patch tested?

## Documentation


core/src/main/java/com/netease/arctic/io/writer/BaseTaskWriter.java
```diff
@@ -123,6 +123,10 @@ public void close() throws IOException {
     writerHolder.close();
   }
 
+  public PrimaryKeyData getPrimaryKey() {
```
zhoujinsong, line 126: I'd recommend making this method to be `protected`.

zhoujinsong (approved):
LGTM.

flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/FlinkChangeTaskWriter.java
```diff
@@ -76,4 +96,25 @@ protected ChangeAction action(RowData data) {
...
+      primaryKey.primaryKey(asStructLike(row));
+      PrimaryKeyData copyKey = primaryKey.copy();
+
+      if (RowKind.UPDATE_AFTER.equals(rowKind)) {
+        if (!hasUpdateBeforeKeys.contains(copyKey)) {
+          row.setRowKind(RowKind.INSERT);
+        }
+        hasUpdateBeforeKeys.remove(copyKey);
+      } else {
+        hasUpdateBeforeKeys.add(copyKey);
```
YesOrNo828, line 116:
```suggestion
        PrimaryKeyData copyKey = primaryKey.copy();
        hasUpdateBeforeKeys.add(copyKey);
```

PR #1223, opened 2023-03-14
[ARCTIC-1167][core][hive] fix Trash for restoring deleted files

### LINKED ISSUE #1167: [Feature]: Support move files to `Trash` when deleting files and restore files from `Trash` later
### Description

We should introduce a new feature to improve table security in the situation of deleting files, like snapshots expiration, and orphan files clean, making it easy to restore files when they are deleted by mistake.

### Use case/motivation

When files are deleted by mistake, whether they are metadata files or data files, it will cause serious problems for the iceberg table, like loading table errors or reading/writing errors.

And also, it's very difficult to recover the table from the abnormal state above; sometimes, restoring the files is the only way to recover the table.

### Describe the solution

1. We should introduce a new feature of Trash to restore deleted files
2. The `Trash` is an optional config for each Table
3. When deleting files of a table with `Trash` enabled, files are supposed to be moved to a specific directory, instead of deleted directly from the file system;  the `Trash` directory is also a config for each Table
4. We can restore deleted files from `Trash` when needed
5. The `Trash` directory is cleaned periodically with an expiration time by AMS

Get more details from the doc https://docs.google.com/document/d/1CYBb2u9KmhTcBVy-uUn_Lifu3osNnAgSkiD20LQhmko/edit#heading=h.74a2yp1p0du8

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

wangtaohz:

fix #1167, add some hot-fix

## Brief change log

  - make TableTrashManager serializable
  - fix the ArcticHadoopFileIO cast error for hive sync
  - moving the file to Trash should overwrite the file that already exists
  - remove useless ArcticFileIoDummy

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

PR #1228, opened 2023-03-15
[ARCTIC-1227] should not be case sensitive when reading the hive upgrade table schema

### LINKED ISSUE #1227: [Bug]: should not be case sensitive when reading the hive upgrade table schema
### Search before asking

optimize after upgrading the arctic table will fail if the case of the field names in the hive table schema and the parquet schema do not match
### How should we improve?
should not be case sensitive when reading the hive upgrade table schema

_No response_

### Subtasks

_No response_

### Code of Conduct

hameizi:

resolve #1227 

## Brief change log

Case differences are ignored when matching parquet file schema and namemapping

## How was this patch tested?

## Documentation

### DIFF
--- a/hive/pom.xml
+++ b/hive/pom.xml
@@ -123,6 +123,7 @@
                 <configuration>
                     <includes>
                         <include>**/ArcticHiveTestMain.java</include>
+                        <include>**/TestAdaptHiveReadConf.java</include>
                     </includes>
                 </configuration>
             </plugin>
--- a/hive/src/test/java/org/apache/iceberg/parquet/TestAdaptHiveReadConf.java
+++ b/hive/src/test/java/org/apache/iceberg/parquet/TestAdaptHiveReadConf.java
@@ -18,33 +18,149 @@
 
 package org.apache.iceberg.parquet;
 
-import com.google.common.collect.Lists;
-import org.apache.iceberg.mapping.MappedField;
-import org.apache.iceberg.mapping.MappedFields;
+import java.util.concurrent.atomic.AtomicInteger;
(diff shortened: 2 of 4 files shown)


shidayang (approved):

hive/src/test/java/org/apache/iceberg/parquet/TestAdaptHiveReadConf.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.parquet.schema.PrimitiveType;
+import org.apache.parquet.schema.Type;
+import org.apache.parquet.schema.Types;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestAdaptHiveReadConf {
+
+  @Test
+  public void testConvertNameMapping() {
```
zhoujinsong, line 35: We may add more struct types in test cases like struct, map, and array.

hive/pom.xml
```diff
@@ -123,6 +123,7 @@
                 <configuration>
                     <includes>
                         <include>**/ArcticHiveTestMain.java</include>
+                        <include>**/TestAdaptHiveReadConf.java</include>
```
zhoujinsong, line 126: Or we can add this new test case to the `ArcticHiveTestMain` suit?

hive/src/test/java/org/apache/iceberg/parquet/TestAdaptHiveReadConf.java
hameizi: added in new commit

PR #1248, opened 2023-03-17
FlinkOptimizer automatically retries after failover 

Sixkd:
# Background
https://github.com/NetEase/arctic/issues/725
When a optimize task is executed by optimizer, changes (such as failover, attempt or manual restart) occur, and the optimizing process is interrupted. The ams does not know about the interrupts. The ams changes the optimize status to failed and tries again after the merger task times out,delaying the optimize task 

# Goals
For Flink-external optimzer ,ams cans know that the optimizer has experienced a state change (such as failover, attempt or manual restart), and promptly sets the previous optimize process to fail and retries, reducing latency

# Non-goals
Optimizer other than Flinks-external-Optimzer

# Overall Design
The Reporter class of flink-optimizer  and AMS heartbeat processing

# Implementation Plan
Flink-external-optimizer
1、Flink-optimizer args adds flink api's url prefix
2、Get the last_modification via flink's rest api and report the last_modification to ams via heartbeat as a status identification

AMS-OptimizerService.checkOptimizerRetry
1、If the newReportData status identification exists and the new and old status identification is inconsistent, the status of the optimzer has changed
2、Obtain the optimizer-id and query the task with create_time smaller than last_modification being executed by the optimzer. Set the task to failed
3、Update OptimizerState

# Test Plan
Restart flink-optimzer. Observe that the task corresponding to optimize_task is set to failed because the optimizer job has occurred retry.

# Compatibility, Deprecation, and Migration Plan
This parameter is valid only for flink tasks with a rest-prefix and a specified id in the args


ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizerService.java
```diff
@@ -226,5 +236,52 @@ private Optimizer fillContainerType(Optimizer optimizer) {
...
+      try (SqlSession sqlSession = getSqlSession(true)) {
+        OptimizeTasksMapper optimizeTasksMapper = getMapper(sqlSession, OptimizeTasksMapper.class);
+        //Gets the task in execution of the optimizer
+        List<BaseOptimizeTask> baseOptimizeTasks = optimizeTasksMapper
+                .selectOptimizeTasksByJobIDAndStatus(optimizerId, OptimizeStatus.Executing.name());
+        for (BaseOptimizeTask baseOptimizeTask : baseOptimizeTasks) {
+
+          //Since optimzer may poll tasks before heartbeat,
+          //only tasks whose createtime is less than status Identification are set to fail
+          if (baseOptimizeTask.getCreateTime()  < Long.parseLong(statusIdentification)) {
```
wangtaohz: Checking `create time of task` < `last modify time of optimizer` is not safe, I recommend checking `execute time of task` < `last modify time of optimizer`.

We can get the execute time of task by 
```
OptimizeTaskItem optimizeTask = XXX;
optimizeTask.getOptimizeRuntime().getExecuteTime();
```

wangtaohz (changes requested):
@Sixkd Thanks for your contribution. I have left some comments, please take a look.

wangtaohz:
I am wondering if there is a simpler and more general way to get the `last-modification` time of the flink job instead of the REST API of yarn.

For example, we can keep the time when `FlinkReporter` is created, like 
```
  private long createTime;

  public FlinkReporter(BaseTaskReporter taskReporter, BaseToucher toucher, OptimizerConfig optimizerConfig) {
    this.taskReporter = taskReporter;
    this.toucher = toucher;
    this.heartBeatInterval = optimizerConfig.getHeartBeat();
    this.createTime = System.currentTimeMillis();
  }

```

then we can use the `createTime` of `FlinkReporter` as the `last-modification`, since there is only one `FlinkReporter` instance for a flink job.

Do you think it's practicable? @Sixkd

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OptimizerService.java
```diff
@@ -132,8 +142,8 @@ public void updateOptimizerState(OptimizerStateReport reportData) throws NoSuchO
         } else {
           stateInfo = reportData.optimizerState;
         }
-        optimizerMapper.updateOptimizerState(reportData.optimizerId, factory.serialize(instance),
-            stateInfo, TableTaskStatus.RUNNING.name());
+        optimizerMapper.updateOptimizerState(reportData.optimizerId, optimizerinstance,
```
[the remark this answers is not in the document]
hameizi: Yes, for external optimizer we don't store optimizerinstance

optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkReporter.java
```diff
@@ -36,9 +36,18 @@ public class FlinkReporter extends AbstractStreamOperator<Void>
...
   public static final String STATE_JOB_ID = "flink-job-id";
 
+  private static final String JOB_OVERVIEW_REST_API = "/jobs/overview";
+
+  private static final String HIGH_AVAILABILITY_CLUSTER_ID = "high-availability.cluster-id";
+
+  private static final String STATUS_IDENTIFICATION = "status_identification";
+
+  private static final String LAST_MODIFICATION = "last-modification";
+
```
wangtaohz, line 46: `JOB_OVERVIEW_REST_API` `HIGH_AVAILABILITY_CLUSTER_ID` `LAST_MODIFICATION` are useless, we should remove them.

optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkReporter.java
```diff
@@ -36,9 +36,18 @@ public class FlinkReporter extends AbstractStreamOperator<Void>
...
   private final BaseTaskReporter taskReporter;
   private final BaseToucher toucher;
   private final long heartBeatInterval;
+  private long createTime;
```
wangtaohz, line 50: The field `createTime` may be `final`.
It's better to refactor the constructors to 
```
  public FlinkReporter(BaseTaskReporter taskReporter, BaseToucher toucher, OptimizerConfig optimizerConfig) {
    this.taskReporter = taskReporter;
    this.toucher = toucher;
    this.heartBeatInterval = optimizerConfig.getHeartBeat();
    this.createTime = System.currentTimeMillis();
  }

  public FlinkReporter(OptimizerConfig config) {
    this(new BaseTaskReporter(config), new BaseToucher(config), config);
  }
```

wangtaohz (commented):
I left some comment, please take a look @Sixkd

wangtaohz (approved):
LGTM.

Issue #1241, opened 2023-03-16
[Subtask]: Change the Arctic source splits discovery and assigning thread model


YesOrNo828:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### Description

Currently, there are two threads that operate the ConcurrentHashMap `readersAwaitingSplit`, one thread iterates the map and removes an entry via readersAwaitingSplit.entrySet().iterator().remove(), and the other thread puts an entry into the map. These threads may operate the same key. So conflicts may happen.

I'm going to change the async thread model to the synchronous thread model and discard the scheduled thread. It would improve the speed at which readers fetch splits.

### Parent issue

#1230 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1251, opened 2023-03-20
[ARCTIC-1241] Change the Arctic source splits discovery and assigning thread model

YesOrNo828:

fix: #1241 
## Brief change log

  - *Change the ShuffleSplitAssigner's on discovery and waiting to assign splits to readers logic*
  - *The flink 1.1, 1.14, and 1.15 versions are affected.*
  - *Add the static and continuous splits assigned UTs*

## How was this patch tested?

## Documentation


flink/v1.12/flink/src/main/java/com/netease/arctic/flink/read/hybrid/assigner/SplitAssigner.java
```diff
@@ -35,8 +37,11 @@ default void open() {
 
   Optional<ArcticSplit> getNext();
```
zhoujinsong, line 38: Should we change the result type of this method too?

YesOrNo828: Yes, I return a `Split` instead of `Optional<ArcticSplit>`

YesOrNo828:
@zhoujinsong  I mainly change the getNext API in the SplitAssigner class.

zhoujinsong (approved):
LGTM.

Issue #1233, opened 2023-03-16
[Improvement]:  arctic sink task can't restore from checkpoint after modify sink parallelism


Aireed:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

In some production scenarios, the failure to account for traffic bursts leads to backpressure of tasks at a certain time. We need to increase the concurrency of tasks to solve this problem. Currently it is not possible to recover from checkpoint due to the way the state is implemented.

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

YesOrNo828:
![image](https://user-images.githubusercontent.com/6017591/227157446-3aa84c30-f416-4ec1-a606-e694cd4a46b0.png)
Thanks, @Aireed for mentioning this improvement. I met this exception after reducing or increasing the config option `sink. parallelism` and then restoring the flink job from a checkpoint.

PR #1275, opened 2023-03-24
[ARCTIC-1233] Arctic filestore write task can't been restored from checkpoint after modifying the sink parallelism

YesOrNo828:

fix: #1233 

This patch #1010 has changed the txId generation rule, the flink writer operator doesn't need to fetch a transaction with a checkpoint before committing a snapshot from AMS. Then, the checkpoint no longer needs to be generated and maintained in the state.

## Brief change log

  - *Remove the useless code in the file writer operator*
  - *Remove the state which contains a checkpoint in the file writer operator*
  - *The flink 1.12, 1.14, and 1.15 versions are affected.*

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

Issue #1277, opened 2023-03-24
[Bug]: corresponding the actual arctic table schema and flink table schema when using flink datastream API


YesOrNo828:
### What happened?

```
DataStreamSink<?> ds = FlinkSink
          .forRowData(dataStream)
          .table(table)
          .flinkSchema(flinkTable.getSchema())
          .tableLoader(tableLoader)
          .overwrite(overwrite)
          .build();
```
When the flinkSchema is out-of-order with the arctic table schema, the arcticWriter may get a wrong field value to write into the arctic.

[FlinkTaskWriterBuilder.java](https://github.com/NetEase/arctic/blob/6af772f5f403c9a5c416d7746de5331a44bd9d81/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/write/FlinkTaskWriterBuilder.java#L180):
flinkSchemaWithMeta can not be convert from arctic table schema, instead of flink table schema.
```
Schema selectSchema = TypeUtil.reassignIds(
        FlinkSchemaUtil.convert(FlinkSchemaUtil.toSchema(flinkSchema)), keyedTable.baseTable().schema());
    Schema changeSchemaWithMeta = SchemaUtil.changeWriteSchema(keyedTable.baseTable().schema());
    RowType flinkSchemaWithMeta = FlinkSchemaUtil.convert(changeSchemaWithMeta);

    OutputFileFactory outputFileFactory = new CommonOutputFileFactory(keyedTable.changeLocation(),
        keyedTable.spec(), fileFormat, keyedTable.io(), keyedTable.baseTable().encryption(), partitionId,
        taskId, transactionId);
    FileAppenderFactory<RowData> appenderFactory = TableTypeUtil.isHive(table) ?
        new AdaptHiveFlinkAppenderFactory(changeSchemaWithMeta, flinkSchemaWithMeta,
            keyedTable.properties(), keyedTable.spec()) :
        new FlinkAppenderFactory(
        changeSchemaWithMeta, flinkSchemaWithMeta, keyedTable.properties(), keyedTable.spec());
```

### Affects Versions

master

### What engines are you seeing the problem on?

Flink

### How to reproduce

Using Flink Datastream Api write data into the arctic table;
Define a flink schema that is out-of-order with the arctic table schema, for example, the flink table fields size is more than the arctic table fields size;

### Relevant log output
![image](https://user-images.githubusercontent.com/6017591/227512824-487d66e0-f735-427f-91b4-16d002e7f51e.png)
[...]

PR #1297, opened 2023-03-29
[ARCTIC-1277] Support writing partial and out-of-order field data into mixed format tables

YesOrNo828:

fix: #1277 

Due to initiating a task writer with an incorrect schema when writing partial field data into a mixed format table, task writers will have a field type cast exception.
For example: 
```
Arctic Table Schema:
id int,
name string,
age,
opt timestamp

Flink Table Schema:
age,
id,
opt timestamp
```
Flink task only writes three field data into an arctic table and also the order of fields is different from the arctic table's fields.
## Brief change log

  - *Modify the correct schema in the FlinkTaskWriterBuilder class*
  - *The flink 1.12, 1.14, and 1.15 versions are affected.*
  - *Add the mixed-hive and mixed-iceberg table writing and reading tests*

## How was this patch tested?

## Documentation


shidayang (approved):

shidayang:
LGTM

YesOrNo828:
This patch contains a partial-update feature which needs a proposal, so close it.

PR #1329, opened 2023-04-11
update jacoco version to 0.8.9

baiyangtx:

Just a test to fix EOF problem in github ci

## Brief change log

- Upgrade jacoco version to 0.8.9


baiyangtx:
It doen't matter with the version of jacoco

Issue #1166, opened 2023-02-27
[Feature]: Trino supports table statistics


HuangFru:
### Description

Trino supports table statistics for better query performance.

### Use case/motivation

Table statistics can help the query engine generate a better query plan which may improve our performance.

### Describe the solution

Take the statistics information from the Arctic-Core module and handle the information and return them to Trino.

### Related issues

To develop this feature, we must solve #1143.

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1344, opened 2023-04-13
[ARCITC-1166][Trino] Trino supports table statistics for the keyed table

### LINKED ISSUE #1143: [Bug]: Reuse containers in ManifestReader will cause one IcebergFileEntry in the Java map covered by another
### What happened?

I'm trying to develop the Table Statistics feature in Arctic-Trino #1166. When I add the includeColumnStats builder in the planning stage in Arctic-Core, I found the statistics information is wrong in the map which is used to group files by partition.

When Java map stores an object which is not immutable, it only stores references to objects. So if the object changed, the value in the map will change too. When planning tasks in the change store, we get the entries of manifest by rows() method in ManifestReadTask in BaseEntriesTable. This method calls the open() method in ManifesrReader which will reuse containers. Once it reuses the container, the value in the Java map will be changed for it did not create a new object but just reuse the last object. So the next value's statistics will cover the last one's statistics.

This is the map mentioned above:
image

This is the rows() method mentioned above and where we call it:
image
image

This is the open() method mentioned above:
截屏2023-02-22 17 35 18

### Affects Versions

May not affect in master, but it will.

Core

### How to reproduce

Because we didn't use the statistics information in the code after the map was created, there's no problem for now. But we're going to develop the Table statistics feature, so it will be a problem.
About reproduce:
Just write some files into a table's change store and debug the plan task stage adding debug breakpoint to the map mentioned above and watching it.

Here's my debug information:
This is 5-I-2-00000-0-0000000002.parquet which is first put in the map:
image
This is 4-I-2-00000-0-0000000001.parquet which is put in the map after 5-I-2-00000-0-0000000002.parquet:
image

We can see that they have the same object about statistics. When the latter value was put in, the former one's statistics information will be covered.

### Relevant log output

```shell
See the pictures in 'How to reproduce'.
```

### Anything else

_No response_

### Code of Conduct

HuangFru:
Support table statistics for the keyed table in Trino for better query performance. #1166 
Resolve #1143 

## Brief change log

  - *Add the 'includeColumnStats' interface for KeyedTableScan and implements it in BasicKeyedTableScan.*
  - *Make statistics always new in TableEntriesScan to resolve #1143.*
  - *Support table statistics in KeyedConnectorMeta in Trino.*
  - *Rebuild the ArcticSessionProperties to include IcebergSessionProperties in it and ArcticConfig could be set in the session.*

## How was this patch tested?

## Documentation


shidayang:
We should try to reuse the Iceberg implementation by only counting the base table or masquerading the ArcticTable as a Table

shidayang (approved):

PR #1355, opened 2023-04-17
[Arctic-1364]: Disable jacoco in github CI

baiyangtx:
disable jacoco when run github ci for not blocking code development.

This PR releated to #1364 but not fixed it totally, so this PR will not close the issue. In the future we should find out the reason of the EOFException and enable the test cover check.

 
## Brief change log

- Add `-Djacoco.skip=true` for core ci
- Add `-B ` for core ci to diable print colorful code.
- Add `--no-transfer-progress` to not print download process

### DIFF
--- a/.github/workflows/core-ci.yml
+++ b/.github/workflows/core-ci.yml
@@ -48,7 +48,7 @@ jobs:
           cache: maven
 
       - name: Build all module with Maven
-        run: mvn clean install -pl '!trino' -Djacoco.skip=true -B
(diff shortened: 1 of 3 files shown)


zhoujinsong (approved):

zhoujinsong:
Due to the current Flink module throwing errors during unit test coverage calculation, the coverage plugin has been temporarily disabled and will be re-enabled after the issue is identified and fixed.

PR #1379, opened 2023-04-20
[Arctic-1120]we return bytes array when we response file load request

Aireed:
## Brief change log
fix: https://github.com/NetEase/arctic/issues/1120

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/PlatformFileInfoService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/PlatformFileInfoService.java
@@ -42,19 +42,6 @@ public class PlatformFileInfoService extends IJDBCService {
     }
   }
 
-  /**
-   * get file content
-   * @param fileId
-   * @return
-   */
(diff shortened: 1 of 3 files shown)


wangtaohz (approved):
LGTM

ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/PlatformFileInfoService.java
```diff
@@ -55,4 +55,12 @@ public String getFileContentById(Integer fileId) {
     }
   }
 
+  public byte[] getFileContentBytesById(Integer fileId) {
```
zhoujinsong, line 58: Is the method `getFileContentById ` still needed?

Aireed: There is no call of the method getFileContentById left. 
 i will remove it.

zhoujinsong (approved):
LGTM.

Issue #1372, opened 2023-04-19
[Improvement]: AMS module refactoring in 0.5


majin1102:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

AMS Refactoring Goals:

1. More elegant module designing, such as merging optimizer module into ams module and unifying ams-related interfaces into ams-api.
2. More elegant interface abstraction, reorganizing modules and implementations around table service and optimizing service.
3. Unify optimizing planning logics for mixed format and native Iceberg format, with mixed format planning inheriting from native Iceberg.
4. Abstract resource type to prepare for providing more plugin functionality in the future.
5. Optimize optimizer code logic to prepare for extending more table formats in the future.
6. Improve code simplicity and readability, especially for system database operations.
7. Optimize optimizing service interface and provide standardized exception codes.
8. More refactoring work oriented towards future expansion…

Compatibility:
- Table service interfaces are compatible, which means flink, spark, and trino interfaces are compatible.
- Optimizer interfaces are not compatible and require clearing all optimizers and restarting.

### How should we improve?

This issue and PR would be a total refactoring for ams.This work would be mainly followed by me, @zhoujinsong @hameizi @shidayang 

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Subtasks

- [ ] #1411 

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1392, opened 2023-04-24
Replace the fastjson of the ArcticMetaStore class in ams-server with jackson

329724814:
There are many high-risk vulnerabilities in the previous version of fastjosn,We need to replace fastjson with jackson

## Brief change log
Replace the fastjson of the ArcticMetaStore class in ams-server with jackson,

## How was this patch tested?

## Documentation

### DIFF
--- a/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/reader/ArcticSparkKeyedDataReader.java
+++ b/spark/v3.1/spark/src/main/java/com/netease/arctic/spark/reader/ArcticSparkKeyedDataReader.java
@@ -44,7 +44,7 @@ public class ArcticSparkKeyedDataReader extends AbstractAdaptHiveArcticDataReade
       String nameMapping,
       boolean caseSensitive) {
     super(fileIO, tableSchema, projectedSchema, primaryKeySpec, nameMapping, caseSensitive,
-        ArcticSparkUtils::convertConstant, false);
+        ArcticSparkUtils::convertConstant, true);
   }
 
   @Override
@@ -58,7 +58,8 @@ public class ArcticSparkKeyedDataReader extends AbstractAdaptHiveArcticDataReade
   protected Function<Schema, Function<InternalRow, StructLike>> toStructLikeFunction() {
     return schema -> {
       final StructType structType = SparkSchemaUtil.convert(schema);
-      return row -> new SparkInternalRowWrapper(structType).wrap(row);
+      SparkInternalRowWrapper wrapper = new SparkInternalRowWrapper(structType);
+      return row -> wrapper.wrap(row);
     };
   }
 }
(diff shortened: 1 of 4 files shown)


ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/JacksonUtils.java
```diff
@@ -0,0 +1,133 @@
...
+        JsonNode value = jsonNode.get(fieldName);
+        if(value.isInt()){
+            return value.intValue();
+        }
+        throw new ClassCastException("can not cast to int, value : " + value);
+    }
+
+    public static Boolean getBoolean(JsonNode jsonNode, String fieldName){
+        if(jsonNode == null){
+            throw new NullPointerException();
```
shidayang, line 111: You can use Preconditions.checkNotNull

ams/ams-server/src/main/java/com/netease/arctic/ams/server/utils/JacksonUtils.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+        if(jsonNode.get(fieldName) == null){
```
shidayang, line 113: We should create a local variable like this: JsonNode jsonNode = jsonNode.get(fieldName), because the jsonNode.get(fieldName) method is being called twice.

ams/ams-server/src/test/java/JacksonTest.java
```diff
@@ -0,0 +1,121 @@
...
+        JsonNode yamlConfig = load(configPath);
+        ObjectNode systemConfig =(ObjectNode) yamlConfig.get(ConfigFileProperties.SYSTEM_CONFIG);
+        systemConfig.put("test", "test1");
+        LinkedHashMap<String, Object> config = new LinkedHashMap<String, Object>();
+        config.putAll(JacksonUtils.parseObject(JacksonUtils.toJSONString(systemConfig), Map.class));
+        JsonNode catalogs = yamlConfig.findValues(ConfigFileProperties.CONTAINER_LIST).get(0);
+        for(JsonNode jsonNode : catalogs){
+            System.out.println(jsonNode.get("name").asText());
+        }
+        System.out.println(systemConfig.has("arctic.ams.expire.thread.pool-size"));
```
shidayang, line 30: We cannot simply output at this point, we need to assert the correctness of the value. You can refer to other unit tests for how to do this.

ams/ams-server/src/test/java/JacksonTest.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.InputStream;
+import java.util.*;
+
+public class JacksonTest {
+
+
+
+    @Test
+    public void jacksonTest(){
+        String configPath = System.getProperty("user.dir") + "/src/test/resources/test.yaml";
```
shidayang, line 20: You can find this file from classpath

shidayang (changes requested):

zhoujinsong:
@329724814  Thank you for your outstanding work！

Since the PR mainly involves changes to the AMS module, I suggest waiting for #1372 to be resolved before merging this PR, otherwise, there may be many conflicts.

baiyangtx:
Cloud you please to handle the merge conflict with master branch?  @329724814

Issue #414, opened 2022-09-26
[Improvement]: Refresh the arctic properties in the FlinkTablePropertiesInvocationHandler after table refreshed


YesOrNo828:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?page=4&q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

As of now, the Flink table properties should combine the flink properties and arctic table properties. And the flink properties will replace the same key of the arctic table properties.

```
tableLoader = ArcticTableLoader.of(PK_TABLE_ID, catalogBuilder);

    tableLoader.open();
    ArcticTable arcticTable = tableLoader.loadArcticTable();

    UpdateProperties updateProperties = arcticTable.updateProperties();
    updateProperties.set(LOG_STORE_CATCH_UP, String.valueOf(LOG_STORE_CATCH_UP_DEFAULT));
    updateProperties.set(LOG_STORE_CATCH_UP_TIMESTAMP, String.valueOf(LOG_STORE_CATCH_UP_TIMESTAMP_DEFAULT));
    updateProperties.commit();

    arcticTable.refresh();
    Map<String, String> properties = arcticTable.properties();
    Assert.assertSame("" + LOG_STORE_CATCH_UP_DEFAULT, properties.get(LOG_STORE_CATCH_UP));
    Assert.assertSame("" + LOG_STORE_CATCH_UP_TIMESTAMP_DEFAULT, properties.get(LOG_STORE_CATCH_UP_TIMESTAMP));
```

### How should we improve?

Add a new proxy method(refresh) of the arctic table in the `FlinkTablePropertiesInvocationHandler`

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1393, opened 2023-04-24
[Arctic-414][Improvement]: Refresh the arctic properties in the FlinkTablePropertiesInvocationHandler after table refreshed

YesOrNo828:
fixed: #414 

Arctic Table Properties has been updated, but the arctic table properties in the FlinkTablePropertiesInvocationHandler didn't be refreshed.
Reproduced by the follow ut:
```java
{
    ArcticTableLoader tableLoader = ArcticTableLoader.of(PK_TABLE_ID, catalogBuilder);

    tableLoader.open();
    ArcticTable arcticTable = tableLoader.loadArcticTable();
    boolean catchUp = true;
    String catchUpTs = "1";

    UpdateProperties updateProperties = arcticTable.updateProperties();
    updateProperties.set(LOG_STORE_CATCH_UP.key(), String.valueOf(catchUp));
    updateProperties.set(LOG_STORE_CATCH_UP_TIMESTAMP.key(), catchUpTs);
    updateProperties.commit();

    arcticTable.refresh();
    Map<String, String> properties = arcticTable.properties();
    Assert.assertEquals(String.valueOf(catchUp), properties.get(LOG_STORE_CATCH_UP.key()));
    Assert.assertEquals(catchUpTs, properties.get(LOG_STORE_CATCH_UP_TIMESTAMP.key()));
  }
``` 

## Brief change log

  - *add rewrite method in the FlinkTablePropertiesInvocationHandler*
  - *The flink 1.12, 1.14 and 1.15 versions are affected.*
  - *add some UTs*

## How was this patch tested?

## Documentation

### DIFF
--- a/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/interceptor/FlinkTablePropertiesInvocationHandler.java
+++ b/flink/v1.12/flink/src/main/java/com/netease/arctic/flink/interceptor/FlinkTablePropertiesInvocationHandler.java
@@ -34,16 +34,18 @@ import java.util.Map;
 public class FlinkTablePropertiesInvocationHandler implements InvocationHandler, Serializable {
 
   private ArcticTable arcticTable;
-  private Map<String, String> flinkTableProperties = new HashMap<>();
+  private final Map<String, String> flinkTableProperties = new HashMap<>();
+  protected Map<String, String> tablePropertiesCombined = new HashMap<>();
 
   public FlinkTablePropertiesInvocationHandler(Map<String, String> flinkTableProperties,
                                                ArcticTable arcticTable) {
-    this.flinkTableProperties.putAll(arcticTable.properties());
+    this.tablePropertiesCombined.putAll(arcticTable.properties());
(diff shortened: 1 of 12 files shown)


zhoujinsong (approved):
These codes look good to me.
However, I don't think using reflection to solve Flink custom table properties is a good solution.
Can we use a more reasonable approach later to solve this problem? How do you think @YesOrNo828

YesOrNo828:
> These codes look good to me. However, I don't think using reflection to solve Flink custom table properties is a good solution. Can we use a more reasonable approach later to solve this problem? How do you think @YesOrNo828

I agree that using reflection to solve Flink custom table properties is not the most ideal solution. A better approach would be to define a structured object that can be easily serialized and deserialized. This would provide better type safety and make it easier to read and manage custom table properties.

Additionally, wrapping the Arctic table is an approach to managing custom table properties in Flink. This approach can provide a more seamless integration with Arctic, and can make it easier to manage custom table properties in a centralized location.

PR #1395, opened 2023-04-25
[ARCTIC-1394] Optimize the execution time of drop table

### LINKED ISSUE #1394: [Improvement]: The execution time of delete table operation is too long
### Search before asking

After performing the drop table operation, if there are many hdfs files in the Arctic table, the file cleaning time will be very long, even up to 1-2 hours. If the user rebuilds the table during this process, the hdfs file of the new table may be cleaned up by the previous table deletion operation.

### How should we improve?

Optimize the operation of deleting tables.

### Subtasks

_No response_

### Code of Conduct

ShawHee:
fix #1394 

After performing the drop table operation, if there are many hdfs files in the Arctic table, the file cleaning time will be very long, even up to 1-2 hours. If the user rebuilds the table during this process, the hdfs file of the new table may be cleaned up by the previous table deletion operation.

## Brief change log
core/src/main/java/com/netease/arctic/catalog/BasicArcticCatalog.java

The doDropTable method has a purge parameter, which represents whether to clean up the file. The current operation is that regardless of whether the purge is true or false, the table deletion method of iceberg will be called first to delete the change and base tables respectively, and finally the table location directory will be cleared.
However, iceberg's table deletion method is slow to execute, and if it is decided to clear the table file, the previous steps are not required, and the table location directory in hdfs can be cleared directly.

**Before：**

```
if purge == true:
      1. delete change table (by iceberg)
      2. delete base table (by iceberg)
      3. delete table location (hdfs operations)
else:
      1. delete change table (by iceberg)
      2. delete base table (by iceberg)
```

**After modification:**

```
if purge == true:
      1. delete table location (hdfs operations)
else:
      1. delete change table (by iceberg)
      2. delete base table (by iceberg)
```

In the test, to delete an arctic table with 20,000 files, it takes about 10 minutes to use the previous method, but it only takes 1 second to use the modified method.

## How was this patch tested?


zhoujinsong:
@ShawHee Thanks a lot for your contribution!
From your description, the file system delete recursively API has a much better performance than Iceberg Tables drop table API.So deleting table location recursively when the purge configuration is enabled is indeed a good optimization method.

However, this optimization method still has a limitation, which is that all manifest/data files must be located under the table directory. In the current implementation of Arctic, this condition is satisfied. However, in future implementations, this limitation may be broken.

Therefore, I think this optimization can be merged in at the current stage, but additional comments should be added to explain the limitation.

shidayang (approved):

shidayang:
LGTM

ShawHee:
@zhoujinsong Thanks for your suggestions. I've added a comment to the code explaining the limitations of this optimization.

zhoujinsong (approved):
LGTM.

PR #1409, opened 2023-04-27
[ARCTIC-1408][Ams]Disable authentication for ZooKeeper client.

### LINKED ISSUE #1408: [Improvement]: Disable authentication for ZooKeeper client.
### Search before asking

This issue is related to #1346 

The previous approach was to solve the problem by clearing the Kerberos authentication information, but now a more user-friendly and convenient way has been found, which is to disable authentication for the ZooKeeper client to solve the problem.

### How should we improve?

When an exception is thrown, catch the exception and set `zookeeper.sasl.client` = `false`.

### Subtasks

_No response_

### Code of Conduct

hellojinsilei:

resolve #1408 
The previous approach was to solve the problem by clearing the Kerberos authentication information, but now a more user-friendly and convenient way has been found, which is to disable authentication for the ZooKeeper client to solve the problem.

## Brief change log

System.setProperty("zookeeper.sasl.client", "false");

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-api/src/main/java/com/netease/arctic/ams/api/client/ArcticThriftUrl.java
+++ b/ams/ams-api/src/main/java/com/netease/arctic/ams/api/client/ArcticThriftUrl.java
@@ -143,16 +143,7 @@ public class ArcticThriftUrl {
           retryCount++;
           logger.error(String.format("Caught exception, retrying... (retry count: %s)", retryCount),
               authFailedException);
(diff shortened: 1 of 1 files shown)


hellojinsilei:
I have validated it in the environment and suggest that it be merged as soon as possible because the impact of the previous solution to clear Kerberos authentication is unclear.

PR #1425, opened 2023-05-06
Commit the DeleteFile delete operation separately in Iceberg Format

shidayang:
Deleting the old DeleteFile operation often causes the commit to fail, but this does not affect the accuracy of the data, so we want to perform this operation alone to avoid Optimizing always failing

## Brief change log
Commit the DeleteFile delete operation separately in IcebergCommit

## How was this patch tested?

## Documentation


majin1102 (commented):
LGTM

PR #1433, opened 2023-05-09
[AMS-Refactor] fix runtime bug

hameizi:

## Brief change log

fix runtime bug

## How was this patch tested?

## Documentation


majin1102 (commented):
LGTM

majin1102 (commented):
LGTM

PR #1443, opened 2023-05-16
[Arctic-1362]: Refactor ArcticFileIO to decouple it from Hadoop 

### LINKED ISSUE #1362: [Subtask]: Remove dir operation from ArcticFileIo in order to support S3 storage.
### Search before asking

### Description

Currently, the ArcticTable.io() return ArcticFileIo interface, this interface is hadoop oriented and some mehods are not supported when using OSS storage.

So refactor this code and move the filesystem based methods to Hive module.

### Parent issue

https://github.com/NetEase/arctic/issues/1061

### Code of Conduct

baiyangtx:

fix #1362

 
## Brief change log

- Remove dir operartion from ArcticFileIO interface.
- Add SupportFileSytemOperations interface for directory operation.
- Using ArcticHadoopFileIO instead of ArcticFileIO for MixedHive table.
- The orphan file cleaner will using SupportPrefixOperations interface if the file io dosen't support listDirectory. 
- Remove FileStatus from interface and using PathInfo instead.

## How was this patch tested?


ams/server/src/main/java/com/netease/arctic/server/table/executor/OrphanFilesCleaningExecutor.java
```diff
@@ -167,85 +170,97 @@ private static Set<String> getValidContentFiles(ArcticTable arcticTable) {
...
-    if (internalTable.io().exists(dataLocation)) {
-      for (FileStatus fileStatus : internalTable.io().list(dataLocation)) {
-        deleteFilesCnt += deleteInvalidContentFiles(
-            internalTable.io(),
-            fileStatus,
-            lastTime,
-            exclude);
+
+    try (ArcticFileIO io = internalTable.io()) {
+      if (io.supportFileSystemOperations()) {
```
zhoujinsong, line 177: Why do we need to differentiate between `FileSystemIO` and `PrefixIO` here?

ams/server/src/test/java/com/netease/arctic/server/optimizing/TestMixedHiveOptimizing.java
```diff
@@ -136,7 +140,10 @@ private List<Record> readHiveTableData() throws TException, IOException {
   }
 
   private List<String> filesInLocation(String location) {
-    return arcticTable.io().list(location).stream().map(fileStatus -> fileStatus.getPath().toString())
-        .collect(Collectors.toList());
+    try (ArcticHadoopFileIO io = ((SupportHive) arcticTable).io()) {
```
zhoujinsong, line 143: As far as I know, we don't need to close the IO after use because the current implementation is empty, and the table IO is shared, so it should not be closed after usage.

core/src/main/java/com/netease/arctic/utils/TableFileUtil.java
```diff
@@ -107,4 +107,9 @@ public static String getNewFilePath(String newDirectory, String filePath) {
   public static String getUriPath(String path) {
     return URI.create(path).getPath();
   }
+
+  public static String getParent(String path) {
```
zhoujinsong, line 111: add java doc for util method.

PR #1481, opened 2023-05-26
[Arctic-1476]: Support load and optimizing iceberg table on s3

### LINKED ISSUE #1476: [Feature]: Support Iceberg on S3 or other non-hadoop storage system.
### Description

Currently, Arctic only supports running within the Hadoop ecosystem, and it cannot be used effectively with other storage types like Glue + S3. We should support more types of storage systems.

### Use case/motivation

We should support managing Iceberg tables with Arctic in S3 + Glue scenarios, optimize it for better performance, and enable features like cleaning up orphan files.

### Describe the solution

1. ArcticFileIO interface should be refactored to reduce its strong dependency on Hadoop.
2. The Optimizer logic needs to be retested to ensure that it supports non-Hadoop systems.
3. The logic for cleaning orphan files needs to be adapted to non-file system scenarios. In extreme cases where the storage system does not support list operations, this functionality may need to be downgraded or even unavailable.
4. Due to the fact that cloud object storage is usually billed based on the number of API calls, the logic of the optimizer checker needs to be adjusted to avoid generating unnecessary costs from frequent checks.

### Related issues

- https://github.com/NetEase/arctic/issues/1362

### Code of Conduct

baiyangtx:

fix #1476 to support load and optimzing iceberg table with s3 or other non-hadoop storage system.

## Brief change log

- Add ArcticFileIOIcebergAdapter 
- Using ArcticFileIOIcebergAdapter instead of HadoopFileIO when it's not a hadoop file system.

## How was this patch tested?


core/src/main/java/com/netease/arctic/table/BasicTableBuilder.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.iceberg.relocated.com.google.common.collect.Maps;
+
+import java.util.Map;
+
+
+/**
+ * A base class of interface {@link TableBuilder}
+ * @param <ThisT> self class
+ */
+public abstract class BasicTableBuilder<ThisT extends TableBuilder> implements TableBuilder {
```
zhoujinsong, line 36: Should the `BasicArcticCatalog#ArcticTableBuilder` extend this basic class too?

baiyangtx: In the future, this class will inherit from this parent class, but this part of the change is not involved in the current PR, and there will be other PRs to do this in the future

zhoujinsong (approved):
LGTM.

PR #1488, opened 2023-05-30
[Refactor-AMS] Fix TableRuntime recreate

HuangFru:
TableRuntime will recreate and persistent to DB if the optimizer group is not set correctly.

## Brief change log

  - *Make sync table do as transaction*
  - *Add a unique key for table runtime with table name, DB name and catalog name.*

## How was this patch tested?

## Documentation


ams/server/src/main/resources/mysql/ams-mysql-init.sql
```diff
@@ -118,7 +118,8 @@ CREATE TABLE `table_runtime`
     `optimizer_group`               varchar(64) NOT NULL,
     `table_config`                  mediumtext,
     `optimizing_config`             mediumtext,
-    PRIMARY KEY (`table_id`)
+    PRIMARY KEY (`table_id`),
+    UNIQUE KEY `table_index` (`catalog_name`,`db_name`,`table_name`)
```
zhoujinsong, line 122: add unique index for debery  init sql too.

PR #1490, opened 2023-05-30
[ARCTIC-1385][Bug]: fix transaction and action fields value error

### LINKED ISSUE #1385: [Bug]: The column ‘_transaction_id’ and '_change_action' will be null when querying change table with spark
### What happened?

The column ‘_transaction_id’ and '_change_action' will be null when querying change table

### Affects Versions

0.4.1

Spark

### How to reproduce

Using spark to query change table with ‘_transaction_id’  '_change_action' columns. 
such as:
`select _transaction_id, _change_action from {tableName}.change`

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

baiyangtx:
 
fix #1385 
 
## Brief change log

- Add class ArcticChangeTableScan which extend Iceberg  DataTableScan.
- Add class ArcticManifestGroup which extend iceberg ManifestGroup
- Use ArcticChangeTableScan instead of BasicChangeTableIncreamentScan
- Remove BasicChangeTableIncreamentScan

## How was this patch tested?

 

core/src/main/java/org/apache/iceberg/ArcticChangeTableScan.java
```diff
@@ -0,0 +1,178 @@
...
+    super(ops, table, schema, context);
+  }
+
+  @Override
+  public ArcticChangeTableScan useSnapshot(long scanSnapshotId) {
+    super.useSnapshot(scanSnapshotId);
+    Schema snapshotSchema = SnapshotUtil.schemaFor(table(), scanSnapshotId);
+    return newRefinedScan(
+        tableOps(), table(), snapshotSchema, context().useSnapshotId(scanSnapshotId));
+  }
```
shidayang, line 54: This method is the same as super class, May be don't need to rewrite

core/src/main/java/org/apache/iceberg/ArcticManifestGroup.java
```diff
@@ -0,0 +1,105 @@
...
+  ArcticManifestGroup scanMetrics(ScanMetrics metrics) {
+    super.scanMetrics(metrics);
+    return this;
+  }
+
+  @Override
+  ArcticManifestGroup ignoreDeleted() {
+    super.ignoreDeleted();
+    return this;
+  }
```
shidayang, line 69: These method don't need to rewrite

shidayang (changes requested):

core/src/main/java/org/apache/iceberg/ArcticManifestGroup.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.Map;
+
+public class ArcticManifestGroup extends ManifestGroup {
+
+  ArcticManifestGroup(FileIO io, Iterable<ManifestFile> dataManifests, Iterable<ManifestFile> deleteManifests) {
+    super(io, dataManifests, deleteManifests);
+  }
+
+  @Override
+  ArcticManifestGroup caseSensitive(boolean newCaseSensitive) {
```
zhoujinsong, line 36: What is the reason to override these methods if we just call the super method?

core/src/main/java/org/apache/iceberg/ArcticManifestGroup.java
```diff
@@ -0,0 +1,105 @@
...
+    return plan(ArcticManifestGroup::createContentFileWithSequence);
+  }
+
+  private static CloseableIterable<ContentFileScanTaskWithSequence> createContentFileWithSequence(
+      CloseableIterable<ManifestEntry<DataFile>> entries, TaskContext ctx) {
+    return CloseableIterable.transform(
+        entries,
+        entry -> {
+          DataFile dataFile = entry.file().copy(ctx.shouldKeepStats());
+          return new ContentFileScanTaskWithSequence(dataFile, entry.fileSequenceNumber());
```
zhoujinsong, line 82: we may use the data sequence number rather than file sequence number?

zhoujinsong: Besides,  is the data sequence number value would be null if the entry status is `ADD`?

core/src/main/java/org/apache/iceberg/ArcticManifestGroup.java
```diff
@@ -0,0 +1,105 @@
...
+        });
+  }
+
+
+  public static class ContentFileScanTaskWithSequence implements ScanTask {
```
zhoujinsong, line 87: Can you simplify the class name to make it consist of 3/4 word?

core/src/main/java/org/apache/iceberg/ArcticManifestGroup.java
baiyangtx: the return type is different. if we don't overwrite the method, the chain call is not available

core/src/main/java/org/apache/iceberg/ArcticManifestGroup.java
baiyangtx: the return type is different. if we don't overwrite the method, the chain call is not available.

core/src/main/java/org/apache/iceberg/ArcticManifestGroup.java
baiyangtx: the entry's file seq inherit from mainfest list if the status is ADDED

baiyangtx: > we may use the data sequence number rather than file sequence number?

The code I refactored used the file seq num, what's the different?

zhoujinsong (approved):
LGTM.

Issue #1500, opened 2023-06-01
[Bug]: failed to delete orphaned files


xujiangfeng001:
### What happened?

Deployed `HDFS dual cluster` in the working environment, Arctic on `cluster A` and iceberg table files on `cluster B`. Arctic failed to delete orphaned files for iceberg.

### Affects Versions

0.4.1

### What engines are you seeing the problem on?

AMS

### How to reproduce

Deployed `HDFS dual cluster` in the working environment, Arctic on `cluster A` and iceberg table files on `cluster B`. Arctic failed to delete orphaned files for iceberg.

### Anything else

Every Time

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1501, opened 2023-06-01
[ARCTIC-1500][Bug]: failed to delete orphaned files

xujiangfeng001:

To solve the problem of `arctic `being unable to delete `Iceberg `orphaned files, please refer to: https://github.com/NetEase/arctic/issues/1500

## How was this patch tested?

## Documentation

nothing

### DIFF
--- a/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
+++ b/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
@@ -71,7 +71,9 @@ public class ArcticHadoopFileIO extends HadoopFileIO implements ArcticFileIO {
       FileSystem fs = getFs(toDelete);
 
       try {
-        fs.delete(toDelete, false);
+        if (!fs.delete(toDelete, false)) {
+          throw new IOException("fileSystem delete " + path + " is failed");
+        }
       } catch (IOException e) {
         throw new UncheckedIOException("Fail to delete file: " + path, e);
       }
--- a/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
+++ b/core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
@@ -71,9 +71,7 @@ public class ArcticHadoopFileIO extends HadoopFileIO implements ArcticFileIO {
       FileSystem fs = getFs(toDelete);
 
       try {
-        if (!fs.delete(toDelete, false)) {
-          throw new IOException("fileSystem delete " + path + " is failed");
-        }
+        fs.delete(toDelete, false);
       } catch (IOException e) {
         throw new UncheckedIOException("Fail to delete file: " + path, e);
       }
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OrphanFilesCleanService.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/service/impl/OrphanFilesCleanService.java
@@ -215,7 +215,7 @@ public class OrphanFilesCleanService implements IOrphanFilesCleanService {
                                                FileStatus fileStatus,
                                                Long lastTime,
                                                Set<String> exclude) {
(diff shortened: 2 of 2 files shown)


baiyangtx:
Can you explain the reasons for the  problem?

core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
```diff
@@ -71,7 +71,9 @@ public void deleteFile(String path) {
       FileSystem fs = getFs(toDelete);
 
       try {
-        fs.delete(toDelete, false);
+        if (!fs.delete(toDelete, false)) {
+          throw new IOException("fileSystem delete " + path + " is failed");
+        }
```
wangtaohz, line 76: This modification will cause unit tests to fail on Windows according to @baiyangtx. Perhaps we should fix this problem in another PR, and this PR only fix the `OrphanFilesCleanService`.

Additionally, I check the Iceberg `HadoopFileIO`, it also doesn't throw an exception
```
  public void deleteFile(String path) {
    Path toDelete = new Path(path);
    FileSystem fs = Util.getFs(toDelete, hadoopConf.get());
    try {
      fs.delete(toDelete, false /* not recursive */);
    } catch (IOException e) {
      throw new RuntimeIOException(e, "Failed to delete file: %s", path);
    }
  }
```

xujiangfeng001:
> Can you explain the reasons for the problem?



> Can you explain the reasons for the problem?

Hi @baiyangtx The background of the problem is as follows:https://github.com/NetEase/arctic/issues/1500.
Next, let me explain the following reasons:
When using `TableFileUtils.getUriPath(fileStatus.getPath().toString())`, the obtained absolute file path does not include the `hdfs://` prefix.The `HDFS` cluster environment deployed in Arctic is not an HDFS cluster where Iceberg files are stored.
When calling `ArcticHadoopFileIO.delete` to delete orphaned Iceberg files, it actually tries to delete files in the HDFS cluster deployed by Arctic, rather than the actual Iceberg storage HDFS. As a result, the deletion of orphaned Iceberg files is affected.

xujiangfeng001:
> Can you explain the reasons for the problem?

Hi @baiyangtx  The background of the problem is as follows:https://github.com/NetEase/arctic/issues/1500.
Next, let me explain the following reasons:
When using `TableFileUtils.getUriPath(fileStatus.getPath().toString())`, the obtained absolute file path does not include the `hdfs://` prefix.The `HDFS` cluster environment deployed in Arctic is not an HDFS cluster where Iceberg files are stored.
When calling `ArcticHadoopFileIO.delete` to delete orphaned Iceberg files, it actually tries to delete files in the HDFS cluster deployed by Arctic, rather than the actual Iceberg storage HDFS. As a result, the deletion of orphaned Iceberg files is affected.

core/src/main/java/com/netease/arctic/io/ArcticHadoopFileIO.java
xujiangfeng001: Yes, here, if the hdfs file deletion fails, false will be returned

xujiangfeng001: I deleted this code and resubmitted it.

shidayang (approved):

shidayang:
@xujiangfeng001 Thanks for your contributions

shidayang:
LGTM

wangtaohz (approved):
LGTM

baiyangtx (approved):
LGTM

shidayang:
BTW, you can change "refer to: https://github.com/NetEase/arctic/issues/1500" to "fix: https://github.com/NetEase/arctic/issues/1500", so the issue will be closed with the PR

xujiangfeng001:
> BTW, you can change "refer to: #1500" to "fix: #1500", so the issue will be closed with the PR

Thank you very much. I will pay attention to this next time.

PR #1516, opened 2023-06-05
[ARCTIC-1489] The Optimizer reports the complete exception stack to AMS

### LINKED ISSUE #1489: [Improvement]: The Optimizer should report the complete exception stack to AMS
### Search before asking

When a task fails to execute, the optimizer only reported simple error message to AMS, which makes it inconvenient to troubleshoot as one needs to check the optimizer's logs.

### How should we improve?

We should report the complete exception stack to AMS.

### Subtasks

_No response_

### Code of Conduct

wangtaohz:

fix #1489 

## Brief change log

  - The Optimizer reports the complete exception stack to AMS

## How was this patch tested?

## Documentation

### DIFF
--- a/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkConsumer.java
+++ b/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkConsumer.java
@@ -55,7 +55,7 @@ public class FlinkConsumer extends RichParallelSourceFunction<TaskWrapper> {
         if (task != null) {
           sourceContext.collect(task);
         } else {
-          LOG.info("poll no task and wait for {} ms", POLL_INTERVAL);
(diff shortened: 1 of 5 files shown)


wangtaohz:
This PR should be cherry-picked to 0.3.x.

Aireed:
I suggest to add hostname to the exception like 'exception from xxxx'. In the production,  we always have a lot of optimizer,  the message can help us locate the exception from quickly.

wangtaohz:
> I suggest to add hostname to the exception like 'exception from xxxx'. In the production, we always have a lot of optimizer, the message can help us locate the exception from quickly.

It's a good suggestion, but I'm a little worried about whether the host is enough to locate where the exception exist, especially when multiple optimizer's threads are running on the same machine, so this may not be a general solution.

And also, the goal of this PR is to avoid users searching logs on the optimizer by reporting complete exception information.

wangtaohz:
> This PR should be cherry-picked to 0.3.x.

I have cherry-picked it to 0.3.x

PR #1479, opened 2023-05-26
[ARCTIC-1478][AMS] Adapting IcebergCatalogImpl: GlueCatalog

XBaith:
## Why are the changes needed?

fix #1478 

## How was this patch tested?
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request

## Documentation

  - Does this pull request introduces a new feature? (no)
  - If yes, how is the feature documented? (not documented)

PR #1530, opened 2023-06-07
[AMS] Move aws sdk dependency profile

XBaith:

Fix #1479

## How was this patch tested?

## Documentation


baiyangtx (approved):
LGTM

pom.xml
```diff
@@ -594,6 +600,49 @@
                 <hive.version>2.1.1</hive.version>
             </properties>
         </profile>
+        <profile>
+            <id>aws</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <dependencies>
```
baiyangtx, line 608: This should be dependencyManagement.

shidayang, line 608: This will cause all module has these dependencies

baiyangtx (changes requested):
The dependencies shoud be added to core/pom.xml and manager the version in /pom.xml via dependencyManagement.

shidayang (changes requested):

core/pom.xml
```diff
@@ -88,6 +88,11 @@
             <artifactId>iceberg-parquet</artifactId>
         </dependency>
 
+        <dependency>
```
shidayang, line 91: Remove this dependency

shidayang (changes requested):

core/pom.xml
XBaith: Don

baiyangtx (approved):

PR #1531, opened 2023-06-07
[Bug]: Unexpected data in Operations section of table details

### LINKED ISSUE #1453: [Bug]: Unexpected data in Operations section of table details
### What happened?

image
As shown in the figure, without performing any operation on the table, it can be seen that these Properties are repeatedly submitted in the Operation page, with repeated SET and UNSET operations.

### Affects Versions

0.4.x

AMS

### How to reproduce

1.Configure the following parameters on the Catalog configuration page
image
2. Create a new Arctic table, enable self-optimizing, and write data through Flink

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hameizi:
resolve #1453 

## Brief change log

When reporting properties, tracer needs to merge catalog properties.

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/trace/AmsTableTracer.java
+++ b/core/src/main/java/com/netease/arctic/trace/AmsTableTracer.java
@@ -179,7 +179,11 @@ public class AmsTableTracer implements TableTracer {
     if (this.properties != null && Constants.INNER_TABLE_BASE.equals(innerTable)) {
       try {
         Map<String, String> catalogProperties = client.getCatalog(table.id().getCatalog()).getCatalogProperties();
-        commitMeta.setProperties(CatalogUtil.mergeCatalogPropertiesToTable(this.properties, catalogProperties));
+        if (catalogProperties != null) {
+          commitMeta.setProperties(CatalogUtil.mergeCatalogPropertiesToTable(this.properties, catalogProperties));
+        } else {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/netease/arctic/trace/AmsTableTracer.java
```diff
@@ -175,8 +177,13 @@ public void commit() {
       update = true;
     }
     if (this.properties != null && Constants.INNER_TABLE_BASE.equals(innerTable)) {
-      commitMeta.setProperties(this.properties);
-      update = true;
+      try {
+        Map<String, String> catalogProperties = client.getCatalog(table.id().getCatalog()).getCatalogProperties();
+        commitMeta.setProperties(CatalogUtil.mergeCatalogPropertiesToTable(this.properties, catalogProperties));
```
Aireed, line 182: mergeCatalogPropertiesToTable may throw NPE  for catalogProperties may be null .

Aireed (changes requested):

core/src/main/java/com/netease/arctic/trace/AmsTableTracer.java
hameizi: fixed

Aireed (approved):
LGTM

PR #1540, opened 2023-06-08
[ARCTIC-1539][AMS] Make the `updateConfigInternal` method always execute

### LINKED ISSUE #1539: [Bug]: Table refresh not get newly config
### What happened?

A table with optimization disabled via `'self-optimizing.enabled'='false'` did not trigger the optimizer after resetting `'self-optimizing.enabled'='true'`

### Affects Versions

master

AMS

### How to reproduce

1. Start AMS
2. Via `ALTER TABLE` set `'self-optimizing.enabled'='false'`
3. Set `'self-optimizing.enabled'='true'` after the table have a new snapshot

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

zhongqishang:
Fix #1539 

## Brief change log

  - *Make the `updateConfigInternal` method always execute*
  - *Optimize `updateConfigInternal` logic*

## How was this patch tested?

## Documentation


ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
```diff
@@ -180,12 +180,12 @@ public void setPendingInput(OptimizingEvaluator.PendingInput pendingInput) {
...
-      TableConfiguration configuration = tableConfiguration;
-      if (refreshSnapshots(table) || updateConfigInternal(table.properties())) {
+      boolean configChanged = updateConfigInternal(table.properties());
+      if (refreshSnapshots(table) || configChanged) {
         persistUpdatingRuntime();
       }
-      if (configuration != tableConfiguration) {
-        tableHandler.handleTableChanged(this, configuration);
+      if (configChanged) {
+        tableHandler.handleTableChanged(this, tableConfiguration);
```
wangtaohz, line 188: I think we should use the configuration before changed, because the interface is 
```
public interface TableRuntimeHandler {
  ...
  void handleTableChanged(TableRuntime tableRuntime, OptimizingStatus originalStatus);

  void handleTableChanged(TableRuntime tableRuntime, TableConfiguration originalConfig);
 ...
}

```

And the `originalConfig` means configuration before changed.

zhongqishang: You are right.

wangtaohz (approved):
LGTM. 

This problem only occurs when `refreshSnapshots(table)` always return true, which means the table keeps writing data every minute.

shidayang (approved):

shidayang:
LGTM

PR #1550, opened 2023-06-12
[ARCTIC-1496] Additional exception handling should be added to the getTable method in Thrift

### LINKED ISSUE #1469: [Improvement]: Additional exception handling should be added to the getTable method in Thrift.
### Search before asking

The getTable method did not catch all possible exceptions, which may cause the Thrift method caller to throw an unhandled exception com.netease.arctic.shaded.org.apache.thrift.TApplicationException: Internal error processing getTable. 

ams-server log
image

Taskmananger log:
image

### How should we improve?

First  we should add extra exception handing in getTable  of  class ArcticTableMetastoreHandler to avoid such situations.
second ,check if there are similar situations in other methods of the class ArcticTableMetastoreHandler.
Third, thrift client should add some retry logic to handle the situation.

### Subtasks

_No response_

### Code of Conduct

hameizi:
resolve #1469

## Brief change log

Add the server exception message to the exception returned by the Thrift interface.
Error message like:
image

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
@@ -261,6 +261,7 @@ public class ArcticMetaStore {
     }
     residentThreads.forEach(Thread::interrupt);
     residentThreads.clear();
+    ServiceContainer.clearService();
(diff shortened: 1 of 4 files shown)


Aireed (approved):
LGTM

PR #1553, opened 2023-06-12
[ARCTIC-1552] After multiple high availability switches of AMS, the Expire thread pool is not working

### LINKED ISSUE #1552: [Bug]: After multiple high availability switches of AMS, the Expire thread pool is not working.
### What happened?

I deployed two AMS(AMS-1/AMS-2) process configured for high availability services, the leader was AMS-1, after the leader switched to AMS-2 and then swtiched back to AMS-1,  the Expire thread pool did not work. The exception is

image

### Affects Versions

0.3.x,0.4.x

AMS

### How to reproduce

1.deploy two AMS process, config high availability with zookeeper
2.AMS-1 as leader, disconnect the network between zookeeper and AMS-1, the AMS-2 is the leader now
3.recover the network between AMS-1 and zookeeper
4.kill AMS-2, then AMS-1 is the leader now
5.reproduced 

### Relevant log output

_No response_

### Anything else

This problem occur every time.

This issue is caused by `Table Expire Service` using a shutdown thread pool object.

### Code of Conduct

hameizi:
resolve #1552 

## Brief change log

clear ServiceContainer when stop server

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/ArcticMetaStore.java
@@ -260,6 +260,7 @@ public class ArcticMetaStore {
     }
     residentThreads.forEach(Thread::interrupt);
     residentThreads.clear();
+    ServiceContainer.clearService();
     ThreadPool.shutdown();
   }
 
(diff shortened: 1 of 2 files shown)


wangtaohz (commented):
I add some comments, please take a look. @hameizi 

@Aireed Do you think we need to add some test case for this?

PR #1567, opened 2023-06-14
[ARCTIC-1522][AMS][Bug]: The error log for starting the local optimizer process was not…

xujiangfeng001:
fix https://github.com/NetEase/arctic/issues/1522

## Brief change log

Adding failed log for start `localOptimize`.

## How was this patch tested?

## Documentation


ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
```diff
@@ -40,7 +44,8 @@ protected Map<String, String> doScaleOut(String startUpArgs) {
     try {
       String[] cmd = {"/bin/sh", "-c", startUpArgs};
       LOG.info("Starting local optimizer using command:" + startUpArgs);
-      runtime.exec(cmd);
+      Process process = runtime.exec(cmd);
```
shidayang, line 47: The process status should be checked first. You can use process.waitFor() to wait for the program to complete execution and obtain the exit status code.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
```diff
@@ -40,7 +44,8 @@ protected Map<String, String> doScaleOut(String startUpArgs) {
...
+      printProcessErrorStream(process.getErrorStream());
```
shidayang, line 48: Tools such as IOUtils.readLines(process.getErrorStream()).stream().collect(Collectors.joining("\n")) can be used to obtain information about the exceptions.

shidayang (changes requested):

shidayang:
The Flink optimizer also has a similar issue, you should fix it also

ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
xujiangfeng001: I have tried using `process.waitFor()` to retrieve the exit status code and check the process status, but I encountered the following two issues:

1. When `process.waitFor()` returns, I am unable to retrieve the information from the input stream in `process.getErrorStream()` because the input stream is destroyed by the time `process.waitFor()` returns.

2. The `process.waitFor()` method returns the operating system error code. For example, a status code of 127 indicates insufficient permissions to execute the file. However, if there is an internal error within the executed file, such as the inability to find the main class due to a missing lib, `process.waitFor()` will still return 0. This means that I cannot directly determine whether the process was successfully created based solely on the exit status code.

Considering these two issues, I believe that `process.waitFor()` is not suitable for the current scenario.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
xujiangfeng001: I feel that your suggestion looks very good.

xujiangfeng001:
Based on this issue, I have identified another issue:

When inserting resource into the database, it will not determine whether optimize is started. Here, I will consider how to modify it together.

https://github.com/NetEase/arctic/blame/a7ad48a758745dcc25ce75c1f41ec3f904af3e45/ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/OptimizerController.java#L207

xujiangfeng001:
> The Flink optimizer also has a similar issue, you should fix it also

Okay, I will solve this problem together.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
shidayang: You need read input in anther thread，The following code is from Hadoop, you can refer to it.
```
  public int run(List<String> command, List<String> output) {
    int retCode = 1;
    ProcessBuilder pb = new ProcessBuilder(command);
    try {
      Process p = pb.start();
      OutputBufferThread stdOut = new OutputBufferThread(p.getInputStream());
      OutputBufferThread stdErr = new OutputBufferThread(p.getErrorStream());
      stdOut.start();
      stdErr.start();
      retCode = p.waitFor();
      if (retCode != 0) {
        mojo.getLog().warn(command + " failed with error code " + retCode);
        for (String s : stdErr.getOutput()) {
          mojo.getLog().debug(s);
        }
      }
      stdOut.join();
      stdErr.join();
      output.addAll(stdOut.getOutput());
    } catch (Exception ex) {
      mojo.getLog().warn(command + " failed: " + ex.toString());
    }
    return retCode;
  }
```

xujiangfeng001: Thank you for your suggestion. I will refer to the Hadoop code.

xujiangfeng001:
CC @shidayang ,PTAL thx

dist/src/main/arctic-bin/bin/localOptimize.sh
```diff
@@ -18,16 +18,27 @@
...
 
-$JAVA_RUN -Dlog.home=${LOG_DIR} -Dlog.subdir=localOptimizer-$(date +%s) -Xmx$2m com.netease.arctic.optimizer.local.LocalOptimizer "$@" &
\ No newline at end of file
+if [ ! -f $STDERR_LOG ];then
+    touch $STDERR_LOG
+fi
+JAVA_OPTS="-Xmx$2m -Dlog.home=${LOG_DIR} -Dlog.subdir=localOptimizer-$(date +%s)"
+RUN_SERVER="com.netease.arctic.optimizer.local.LocalOptimizer"
+CMDS="$JAVA_RUN $JAVA_OPTS $RUN_SERVER $@"
+nohup ${CMDS} >/dev/null 2>${STDERR_LOG} &
```
shidayang, line 44: Maybe don't need this STDERR_LOG，AMS will print log

ams/optimizer/src/main/java/com/netease/arctic/optimizer/util/ExecUtil.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class ExecUtil {
+  private static final Logger LOG = LoggerFactory.getLogger(ExecUtil.class);
+
+  public static int exec(String[] command, List<String> output) {
+    int retCode = 1;
```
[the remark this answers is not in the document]
xujiangfeng001: Thank you for your suggestion. This looks great.

dist/src/main/arctic-bin/bin/localOptimize.sh
```diff
@@ -18,16 +18,27 @@
...
 export ARCTIC_HOME
 
 LIB_PATH=$ARCTIC_HOME/lib
 LOG_DIR=$ARCTIC_HOME/logs
-export CLASSPATH=$CLASSPATH:$ARCTIC_HOME/conf/optimize
-
+STDERR_LOG=${LOG_DIR}/localOptimize.log.err
+export CLASSPATH=$CLASSPATH:$(find $LIB_PATH/ -type f -name "*.jar" | paste -sd':' -):$ARCTIC_HOME/conf/optimize
+if [ -z $(find $LIB_PATH/ -type f -name "*.jar" | paste -sd':' -) ]; then
+  echo "启动 localOptimize 任务缺少相关 jar,请检查" >&2
```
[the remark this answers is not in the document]
xujiangfeng001: Thank you for your suggestion. This looks great.

dist/src/main/arctic-bin/bin/localOptimize.sh
xujiangfeng001: My understanding is that the logs printed by `STDERR_LOG` are standard error logs at startup, not logs in `AMS`. So, I think this is necessary.

shidayang (approved):

shidayang:
LGTM

PR #1581, opened 2023-06-19
Create COMMUNITY.md

wuqiao:
Create a new community role member introduction

## Brief change log

*(For example:)*
  - *Add the 'scan.startup.mode' configuration to the ArcticScanContext in the flink modules*
  - *The flink 1.12 and 1.14 versions are affected.*
  - *Add the documentation in the flink chapter*

## How was this patch tested?

## Documentation


zhoujinsong:
I'm wondering if it's appropriate to directly place this document in the root directory of the project repository. Is this a common practice? Or should we manage it as part of our official website and release it on the website?

zhoujinsong:
This document will be placed into the new official website project.

PR #1599, opened 2023-06-21
[ARCTIC-1378] Incorrect data synchronization caused the data of the arctic base table to be overwritten

### LINKED ISSUE #1378: [Bug]: Incorrect data synchronization caused the data of the arctic base table to be overwritten
### What happened?

The arctic base table was overwritten due to incorrect data synchronization after an empty hive table was upgraded to an arctic table. Because empty hive tables don't have full 'table.partition-properties' when upgraded, But we need 'hive-location' and 'transient-time' in 'table.partition-properties' to decide if we want to sync our data

### Affects Versions

0.4.0

_No response_

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

hameizi:
resolve #1378

## Brief change log

Set partition properties for all partitions even if there are no files under the partition when upgrading a Hive table to an Arctic table.

## How was this patch tested?

## Documentation

### DIFF
--- a/hive/src/main/java/com/netease/arctic/hive/utils/UpgradeHiveTableUtil.java
+++ b/hive/src/main/java/com/netease/arctic/hive/utils/UpgradeHiveTableUtil.java
@@ -18,18 +18,24 @@
 
 package com.netease.arctic.hive.utils;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.netease.arctic.hive.HMSClientPool;
 import com.netease.arctic.hive.HiveTableProperties;
 import com.netease.arctic.hive.catalog.ArcticHiveCatalog;
 import com.netease.arctic.hive.table.SupportHive;
+import com.netease.arctic.hive.table.UnkeyedHiveTable;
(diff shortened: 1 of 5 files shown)


hive/src/main/java/com/netease/arctic/hive/utils/UpgradeHiveTableUtil.java
```diff
@@ -136,6 +143,7 @@ private static void hiveDataMigration(
       }
     }
     HiveMetaSynchronizer.syncHiveDataToArctic(arcticTable, arcticHiveCatalog.getHMSClient());
+    fillPartitionProperties(arcticTable, arcticHiveCatalog, hiveTable);
```
zhoujinsong, line 146: May need to reload `hiveTable` after `HiveMetaSynchronizer#syncHiveDataToArctic` is called.

zhoujinsong:
@hameizi  Did you reproduce the issue locally and validate that it has indeed been fixed with the new code?

zhoujinsong (approved):
LGTM.

PR #1612, opened 2023-06-26
[optimizer]For flinkcontainer optimizer, When making a connection via ams in high availability mode, NoSuchMethodError is encountered

xieyi888:
issue: https://github.com/NetEase/arctic/issues/1611

For flinkcontainer optimizer, When making a connection via ams in high availability mode, NoSuchMethodError is encountered 

## Brief change log
in ams/optimizer module, shade com.netease.arctic.shaded.org.apache.zookeeper and com.netease.arctic.shaded.org.apache.curator, to avoid NoSuchMethodError for flink optimizer

## How was this patch tested?

## Documentation

  no
  no

### DIFF
--- a/ams/api/pom.xml
+++ b/ams/api/pom.xml
@@ -157,6 +157,48 @@
                     <generator>java</generator>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>create-shaded-jar</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <artifactSet>
+                                <includes>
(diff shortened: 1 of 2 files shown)


zhoujinsong:
@xieyi888  Thanks a lot for reporting this issue and try to fix it.

From the error, it seems that there is a conflict between the version of Zookeeper that the `ams-api` module depends on and other dependencies in the execution environment. Using shade is indeed a good workaround.

However, this conflict may not only occur in the flinkOptimizer but also in other modules, so I'm considering whether performing shade in the ams-api would be a better solution to completely address this issue.

xieyi888:
> @xieyi888 Thanks a lot for reporting this issue and try to fix it.
> 
> From the error, it seems that there is a conflict between the version of Zookeeper that the `ams-api` module depends on and other dependencies in the execution environment. Using shade is indeed a good workaround.
> 
> However, this conflict may not only occur in the flinkOptimizer but also in other modules, so I'm considering whether performing shade in the ams-api would be a better solution to completely address this issue.

Thanks, It's a good solution, I will try to shade Zookeeper in ams-api

PR #1635, opened 2023-06-30
[ARCTIC-1634] avoid using timestamp without default value for MySQL

### LINKED ISSUE #1634: [Improvement]: Fix MySQL timestamp type missing default value
### Search before asking

In MySQL, if a TIMESTAMP field is defined in a table without specifying a default value, it will automatically be set to '0000-00-00 00:00:00' when inserting a row into that field. However, if strict sql mode is enabled, MySQL will prevent the insertion of '0000-00-00 00:00:00' because it is an invalid datetime value. In strict mode, only valid datetime values can be inserted into TIMESTAMP fields, otherwise an error will be thrown.

### How should we improve?

To avoid using a TIMESTAMP without a default value.

### Subtasks

_No response_

### Code of Conduct

wangtaohz:

fix #1634 

## Brief change log

  - replace timestamp without default value with datetime default null

## How was this patch tested?

## Documentation


ams/server/src/main/resources/mysql/upgrade-0.4.1-to-0.5.0.sql
```diff
@@ -94,9 +94,9 @@ CREATE TABLE `table_runtime`
     `current_change_snapshotId`     bigint(20) DEFAULT NULL COMMENT 'Change table current snapshot id',
     `last_optimized_snapshotId`     bigint(20) NOT NULL DEFAULT '-1' COMMENT 'last optimized snapshot id',
     `last_optimized_change_snapshotId`     bigint(20) NOT NULL DEFAULT '-1' COMMENT 'last optimized change snapshot id',
-    `last_major_optimizing_time`    timestamp COMMENT 'Latest Major Optimize time for all partitions',
-    `last_minor_optimizing_time`    timestamp COMMENT 'Latest Minor Optimize time for all partitions',
-    `last_full_optimizing_time`     timestamp COMMENT 'Latest Full Optimize time for all partitions',
+    `last_major_optimizing_time`    datetime(3) DEFAULT NULL COMMENT 'Latest Major Optimize time for all partitions',
```
zhoujinsong, line 97: What is the reason for changing the type to datetime here? Now that both timestamp and datetime are used in the table structure, what is our standard for using the two types?

wangtaohz: I think we use `datetime` only in order to make `DEFAULT NULL` valid, because `timestamp DEFAULT NULL` is illegal, and we will get an exception
```
mysql> create table test_table (create_time timestamp default null);
ERROR 1067 (42000): Invalid default value for 'create_time'
```

zhoujinsong: As I know, the reason why you got the error is that field with timestamp type is not null by default, you will set the default null for timestamp type successfully by a statement like `create table test_table (create_time timestamp null default null);`

reference: read `[explicit_defaults_for_timestamp](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_explicit_defaults_for_timestamp)`

wangtaohz: You are right, it works! 

Maybe I should change all `datetime DEFAULT NULL` to `timestamp NULL DEFAULT NULL` to avoid the differences they dealing with time zones.

image

zhoujinsong: If there is no compelling reason to use `datetime`, I also recommend using a single type to store all timestamp information. Don't forget to also modify `uprade-0.4.1-to-0.5.0.sql`.

zhoujinsong (approved):

Issue #1600, opened 2023-06-23
[Bug]: When the partition field is null, the optimizer is always in a failure loop


XiaoGerGer:
### What happened?

If you insert a record with a null partition field and it looks like `/arctic-warehouse/my-catalog/ods.db/table-name/ds_day=null` in hdfs, the Optimizer will encounter an error. And record `fail_reason` is `Text 'null' could not be parsed at index 0` in the table_optimizing_process table, and continue to try to optimize the table, and then in a failed loop, Other tables fail to line up

![image](https://github.com/NetEase/arctic/assets/20039124/110759b3-727c-4243-9a8b-a7b190d9b919)
![image](https://github.com/NetEase/arctic/assets/20039124/fdb3162f-ab69-42ee-b3fe-60bff90f5999)

### Affects Versions

master-0.5.0

### What engines are you seeing the problem on?

Optimizer

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shidayang:
Thank you for reporting this bug. Which format are you using? Mixed-format or Iceberg-format?

zhoujinsong:
> Thank you for reporting this bug. Which format are you using? Mixed-format or Iceberg-format?

To my knowledge, the format used is mixed-iceberg format, but I think iceberg format may also have similar issues.

XiaoGerGer:
> Thank you for reporting this bug. Which format are you using? Mixed-format or Iceberg-format?

mixed-iceberg format

shidayang:
I constructed a Mixed-iceberg table and inserted data with a null partition field, but I was unable to reproduce this issue. 
<img width="822" alt="image" src="https://github.com/NetEase/arctic/assets/26699250/56554a37-68dc-487d-97a6-126d88351a7e">
<img width="1164" alt="image" src="https://github.com/NetEase/arctic/assets/26699250/6482a9e8-3e05-4bb6-a9f2-51639fffb5ef">
Does this need to be triggered under certain specific circumstances? Can you provide more details?

PR #1641, opened 2023-07-03
[ARCTIC-1600][Core]Fix Self-Optimizing error when partition value is null in Mixed-format

shidayang:
fix: #1600 

## Brief change log
Modify the handling of the string "null" in the com.netease.arctic.utils.ArcticDataFiles#data method.

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/utils/ArcticDataFiles.java
+++ b/core/src/main/java/com/netease/arctic/utils/ArcticDataFiles.java
@@ -114,6 +114,10 @@ public class ArcticDataFiles {
               field.name().equals(parts[0]),
           "Invalid partition: %s", partitions[i]);
 
+      if ("null".equals(parts[1])) {
+        parts[1] = null;
+      }
+
       data.set(i, ArcticDataFiles.fromPartitionString(field, spec.partitionType().fieldType(parts[0]), parts[1]));
     }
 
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/netease/arctic/utils/ArcticDataFiles.java
```diff
@@ -114,6 +114,10 @@ public static GenericRecord data(PartitionSpec spec, String partitionPath) {
               field.name().equals(parts[0]),
           "Invalid partition: %s", partitions[i]);
 
+      if ("null".equals(parts[1])) {
```
zhoujinsong, line 117: If the value of the partition is the string "null", will it cause misjudgment here?

shidayang: The PartitionSpec#partitionToPath method of Iceberg cannot distinguish between these two situations. If the existing pattern is used, then it is not possible to avoid this problem.

shidayang:
linked: https://github.com/NetEase/arctic/issues/1642

zhoujinsong (approved):

wangtaohz:
@shidayang Could you cherry-pick this PR to the 0.3.x and 0.4.x?

PR #1655, opened 2023-07-04
[ARCTIC-1648] Commit IDs in Files page are not accurate

### LINKED ISSUE #1648: [Bug]: `Commit IDs` in `Files` page are not accurate
### What happened?

Commit IDs have lost precision.

image

### Affects Versions

0.5.0

AMS

### How to reproduce

Find a non-empty table and go to the Table Files page.

### Relevant log output

_No response_

### Anything else

To convert a Java object to a JSON string, it is possible to lose precision of long data type. This is because the JSON specification does not provide a clear definition for long data type and instead uses the Number type in JavaScript to represent numerical values, which has a precision range of 53 bits.

### Code of Conduct

hameizi:
resolve #1648 

## Brief change log

 Use the String data type to represent snapshotId in the HTTP interface.

## How was this patch tested?

## Documentation

  - Does this pull request introduces a new feature? (yes /no)


wangtaohz (approved):
LGTM

PR #1657, opened 2023-07-04
[ARCTIC-1656][AMS] Make plan return when offered a task and do not evaluate when the table is optimizing

### LINKED ISSUE #1656: [Bug]: AMS poll task will wait until all the scheduled table plans finished
### What happened?

I'm doing a performance test for AMS.
I rewrote the optimizer so that it returns mock results almost immediately, but I still found that the 'minor' status lasted a long time:
image
So I'm finding why in the code:
POPO20230704-174930
I found that if plenty of tables are written in a short period of time, the size of `scheduledTables` could be very big, and there's no break for this 'for' loop. So `poll` action will wait until this plan is finished. This could be a long time and will cause performance problems.

We should add a break once a table plan is finished.

### Affects Versions

master

AMS

### How to reproduce

Make lots of tables and write data to all the tables at the same time.

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

HuangFru:
Fix #1656 and some related questions.

## Brief change log

  - *Make plan task breaks when offered tasks to the task queue.*
  - *Make the evaluator not evaluate when the table is processing in optimizing.*

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

wangtaohz (approved):
LGTM

PR #1667, opened 2023-07-07
[ARCTIC-1666] Support creating a catalog with an empty configuration file.

hameizi:
Support creating a catalog with an empty configuration file.

## Brief change log

Set default value `<configuration></configuration>` for catalog when the store config file for creating a new catalog is empty.

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
@@ -137,7 +137,7 @@ public class OptimizingQueue extends PersistentBase implements OptimizingService
   }
 
   public void removeOptimizer(String resourceId) {
-    authOptimizers.values().removeIf(op -> op.getResourceId().equals(resourceId));
+    authOptimizers.entrySet().removeIf(op -> op.getValue().getResourceId().equals(resourceId));
   }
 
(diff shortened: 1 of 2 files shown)


ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -137,7 +137,7 @@ public List<OptimizerInstance> getOptimizers() {
   }
 
   public void removeOptimizer(String resourceId) {
-    authOptimizers.values().removeIf(op -> op.getResourceId().equals(resourceId));
+    authOptimizers.entrySet().removeIf(op -> op.getValue().getResourceId().equals(resourceId));
```
zhoujinsong, line 140: What is the difference between iterating over entries or values here?

hameizi: `values().removeIf` only removes the value from the map but not the key, while `entrySet().removeIf` can remove both the key and value simultaneously.

zhoujinsong: got it.

zhoujinsong (approved):

PR #1670, opened 2023-07-08
[ARCTIC-1646][Bug]: Build failed when change hadoop version to 3.2.1

zhaomin1423:

close https://github.com/NetEase/arctic/issues/1646#issue-1787080167
## Brief change log

update org.apache.commons.lang.* to org.apache.commons.lang3.* in AMS module.

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/test/java/com/netease/arctic/server/optimizing/flow/view/MatchResult.java
+++ b/ams/server/src/test/java/com/netease/arctic/server/optimizing/flow/view/MatchResult.java
@@ -60,7 +60,8 @@ public class MatchResult {
 
   @Override
   public String toString() {
-    return MoreObjects.toStringHelper(this).add("notInView", notInView)
+    return MoreObjects.toStringHelper(this)
+        .add("notInView", notInView)
         .add("inViewButCountError", inViewButCountError)
         .add("inViewButMiss", inViewButMiss)
         .toString();
--- a/ams/server/src/test/java/com/netease/arctic/server/optimizing/flow/view/MatchResult.java
+++ b/ams/server/src/test/java/com/netease/arctic/server/optimizing/flow/view/MatchResult.java
@@ -18,7 +18,7 @@
 
 package com.netease.arctic.server.optimizing.flow.view;
 
-import com.google.common.base.Objects;
+import com.google.common.base.MoreObjects;
 import org.apache.commons.collections.CollectionUtils;
 import org.apache.iceberg.StructLike;
 
@@ -60,8 +60,7 @@ public class MatchResult {
 
   @Override
   public String toString() {
-    return Objects.toStringHelper(this)
-        .add("notInView", notInView)
+    return MoreObjects.toStringHelper(this).add("notInView", notInView)
         .add("inViewButCountError", inViewButCountError)
         .add("inViewButMiss", inViewButMiss)
         .toString();
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -116,6 +116,11 @@
             <classifier>nohive</classifier>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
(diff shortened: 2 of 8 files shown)


zhoujinsong:
Thanks a lot for your contribution!

Based on your code, I was able to build successfully, but when starting AMS, the following error occurred:
351687333241_ pic

It seems to be caused by a conflict in the Guava dependency. Can you also fix this issue?

zhaomin1423:
> Thanks a lot for your contribution!
> 
> Based on your code, I was able to build successfully, but when starting AMS, the following error occurred: 351687333241_ pic
> 
> It seems to be caused by a conflict in the Guava dependency. Can you also fix this issue?

Guava is used by hadoop common, hive-metastore and this project, Can we explicitly declare a guava dependency？

zhoujinsong:
> Can we explicitly declare a guava dependency？

Yes, although the Arctic project depends on the Guava package that the Iceberg project shade on, in order to solve the conflict between Hadoop and Hive's dependencies on Guava, specifying the Guava version that the project depends on may be a good solution.

However, we may still need to validate this solution by conducting actual deployment testing to ensure that it does not introduce any other issues.

zhaomin1423:
> > Can we explicitly declare a guava dependency？
> 
> Yes, although the Arctic project depends on the Guava package that the Iceberg project shade on, in order to solve the conflict between Hadoop and Hive's dependencies on Guava, specifying the Guava version that the project depends on may be a good solution.
> 
> However, we may still need to validate this solution by conducting actual deployment testing to ensure that it does not introduce any other issues.

Thanks, I build and start ams to do few simple tests.

ams/server/src/test/java/com/netease/arctic/server/optimizing/flow/view/MatchResult.java
```diff
@@ -60,7 +60,7 @@ public boolean isOk() {
 
   @Override
   public String toString() {
-    return Objects.toStringHelper(this)
+    return MoreObjects.toStringHelper(this)
```
zhoujinsong, line 63: In the Arctic project, I suggest using the version of Guava Shaded by the Iceberg project as a dependency. Therefore, the imported classes here can be changed to: `org.apache.iceberg.relocated.com.google.common.base.MoreObjects`

pom.xml
```diff
@@ -341,6 +342,42 @@
                 <version>${awssdk.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>com.google.guava</groupId>
+                <artifactId>guava</artifactId>
+                <version>${guava.version}</version>
+                <exclusions>
```
zhoujinsong, line 349: Is there any reason or basis for the content excluded here?

core/pom.xml
```diff
@@ -116,6 +116,11 @@
             <classifier>nohive</classifier>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
```
zhoujinsong, line 122: Now the Flink/Trino module also has its own declared Guava dependency. If the dependency is declared in the core, it seems that the dependencies of other modules can be deleted.

zhoujinsong (commented):
@zhaomin1423 
Here is the code I previously attempted to solve this problem, which can be used as a reference: https://github.com/NetEase/arctic/compare/master...zhoujinsong:arctic:adapt-commons-dependency

zhaomin1423:
> @zhaomin1423 Here is the code I previously attempted to solve this problem, which can be used as a reference: [master...zhoujinsong:arctic:adapt-commons-dependency](https://github.com/NetEase/arctic/compare/master...zhoujinsong:arctic:adapt-commons-dependency)

Maybe adding a hadoop-3 profile to declare hadoop-client-api as dependency is more suitable for hadoop3.

zhoujinsong:
fixed by 1728.

Issue #1650, opened 2023-07-04
[Improvement]: Reduce the size of the serialization of rewritten files


XBaith:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

For tables with many rewritten files, it is more challenging for DB to read and write records in a single transaction.

### How should we improve?

Try to reduce the size of the serialization of rewritten files first

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

XBaith:
```
2023-07-04 06:57:20,651 ERROR [pool-10-thread-18] [com.netease.arctic.server.optimizing.OptimizingQueue] [] - catalog.db.table(tableId=37) plan failed, continue
com.netease.arctic.server.exception.PersistenceException: org.apache.ibatis.exceptions.PersistenceException: 
### Error committing transaction.  Cause: java.sql.SQLException: message:"raft entry is too large, region 40146788, entry size 39380841" raft_entry_too_large:<region_id:40146788 entry_size:39380841 > 
### Cause: java.sql.SQLException: message:"raft entry is too large, region 40146788, entry size 39380841" raft_entry_too_large:<region_id:40146788 entry_size:39380841 > 
        at com.netease.arctic.server.exception.ArcticRuntimeException.wrap(ArcticRuntimeException.java:123) ~[arctic-ams-server-0.5.0-SNAPSHOT.jar:?]
        at com.netease.arctic.server.persistence.PersistentBase.doAsTransaction(PersistentBase.java:66) ~[arctic-ams-server-0.5.0-SNAPSHOT.jar:?]
        at com.netease.arctic.server.optimizing.OptimizingQueue.access$600(OptimizingQueue.java:52) ~[arctic-ams-server-0.5.0-SNAPSHOT.jar:?]
        at com.netease.arctic.server.optimizing.OptimizingQueue$TableOptimizingProcess.beginAndPersistProcess(OptimizingQueue.java:540) ~[arctic-ams-server-0.5.0-SNAPSHOT.jar:?]
        at com.netease.arctic.server.optimizing.OptimizingQueue$TableOptimizingProcess.<init>(OptimizingQueue.java:334) ~[arctic-ams-server-0.5.0-SNAPSHOT.jar:?]
[...]

Issue #1624, opened 2023-06-28
[Bug]: Rewrite_output MySQL storage length is too short for the tables have a mount of files


XBaith:
### What happened?

After rewriting table, the rewrite output file status will persit in MySQL DB. However, for some tables(especially those with delete file), the original column "rewrite_output" is very short in length.

### Affects Versions

master

### How to reproduce

1. Write a lot of samll data files and delete files
2. Enable self-optimizing
3. After the optimizer rework, errors may occur.

### Relevant log output

```shell
root@arctic-meta-service-6b9c5b984c-rmcjr:/usr/local/ams/arctic# cat logs/history/ams-info-2023-06-26-1.log | grep -B 1 "processId:1687773891937, taskId:116"
2023-06-26 10:06:03,717 INFO [pool-10-thread-15] [com.netease.arctic.server.optimizing.OptimizingQueue] [] - bts_iceberg.pda.callanalyzer_result_metrics_v2_table(tableId=66) plan new task OptimizingTaskId(processId:1687773891937, taskId:115), summary MetricsSummary{newFileSize=0, newFileCnt=0, rewriteDataSize=10260719, rewritePosDataSize=0, equalityDeleteSize=0, positionalDeleteSize=0, rewriteDataFileCnt=4, reRowDeletedDataFileCnt=0, eqDeleteFileCnt=0, posDeleteFileCnt=0}
2023-06-26 10:06:03,717 INFO [pool-10-thread-15] [com.netease.arctic.server.optimizing.OptimizingQueue] [] - bts_iceberg.pda.callanalyzer_result_metrics_v2_table(tableId=66) plan new task OptimizingTaskId(processId:1687773891937, taskId:116), summary MetricsSummary{newFileSize=0, newFileCnt=0, rewriteDataSize=12219285, rewritePosDataSize=0, equalityDeleteSize=0, positionalDeleteSize=0, rewriteDataFileCnt=2, reRowDeletedDataFileCnt=0, eqDeleteFileCnt=0, posDeleteFileCnt=0}
--
2023-06-26 10:37:57,717 INFO [ASYNC-TableRuntimeRefreshExecutor-5] [org.apache.iceberg.BaseMetastoreTableOperations] [] - Refreshing table metadata from new version: hdfs://nameservice1/user/hive/warehouse/mats_streaming.db/test_p7_lab/metadata/00001-b64a98c5-288b-4011-8b31-ac054e9583e0.metadata.json
[...]

PR #1688, opened 2023-07-11
[ARCTIC-1687] Exclude kryo dependency which in ams-core from ams-optimizer module

### LINKED ISSUE #1687: [Bug]: Flink optimizer may throw LinkageError when load KryoInstantiator
### What happened?

flink-streaming-java_2.12 contains a kryo dependency, and the optimizer module depends on the submodule ams-core, which also has a kryo dependency, so there may be a dependency conflict when running Flink Optimizer.

### Affects Versions

master

Optimizer

### How to reproduce

1. enable spillt disk map when starting flink optimizer
2. enable self-optimizing

### Relevant log output

```shell
Caused by: java.lang.LinkageError: loader constraint violation: when resolving method "com.esotericsoftware.kryo.Kryo$DefaultInstantiatorStrategy.setFallbackInstantiatorStrategy(Lorg/objenesis/strategy/InstantiatorStrategy;)V" the class loader (instance of org/apache/flink/util/ChildFirstClassLoader) of the current class, com/netease/arctic/utils/SerializationUtil$KryoInstantiator, and the class loader (instance of sun/misc/Launcher$AppClassLoader) for the method's defining class, com/esotericsoftware/kryo/Kryo$DefaultInstantiatorStrategy, have different Class objects for the type org/objenesis/strategy/InstantiatorStrategy used in the signature
```

### Anything else

fix some other bugs

### Code of Conduct

XBaith:

fix #1687 
enhance #1650 and #1624 

## Brief change log

1. resolve kryo dependency conflict
2. catch exception when unzip rewrite input tasks from db
3. increase the storage length of rewrit_input in derby

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/utils/CompressUtil.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/utils/CompressUtil.java
@@ -18,13 +18,19 @@
 
 package com.netease.arctic.server.utils;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.GZIPOutputStream;
+import java.util.zip.ZipException;
 
 public class CompressUtil {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CompressUtil.class);
 
   /**
    * Compress the given data using gzip.
@@ -58,6 +64,9 @@ public class CompressUtil {
       while ((len = gzipInputStream.read(buffer)) > 0) {
         byteArrayOutputStream.write(buffer, 0, len);
       }
+    } catch (ZipException e) {
+      LOG.warn("Fail to unzip, return original bytes", e);
+      return bytes;
     } catch (IOException e) {
       throw new RuntimeException(e);
     }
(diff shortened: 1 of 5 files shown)


ams/server/src/main/resources/derby/ams-derby-init.sql
```diff
@@ -119,7 +119,7 @@ CREATE TABLE table_optimizing_process (
     plan_time           TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     end_time            TIMESTAMP DEFAULT NULL,
     fail_reason         VARCHAR(4096),
-    rewrite_input       BLOB(64m),
+    rewrite_input       BLOB,
```
shidayang, line 122: What's the difference between BLOB(64m) and BLOB

ams/server/src/main/java/com/netease/arctic/server/utils/CompressUtil.java
```diff
@@ -58,6 +64,9 @@ public static byte[] unGzip(byte[] bytes) {
       while ((len = gzipInputStream.read(buffer)) > 0) {
         byteArrayOutputStream.write(buffer, 0, len);
       }
+    } catch (ZipException e) {
+      LOG.warn("Fail to unzip, return original bytes", e);
+      return bytes;
```
shidayang, line 69: I guess this was done to be compatible with data that's not compressed with Gzip. I think such compatibility would be more appropriate to be placed in the Object2ByteArrayConvert.

XBaith: This method is used by Object2ByteArrayConvert and TaskFilesPersistence#loadTaskInputs. Should I add try-catch to both of them?

shidayang: yeah

shidayang (approved):

shidayang (approved):

PR #1689, opened 2023-07-11
[ARCTIC-1686] Update status start time after optimizing status changed to pending

### LINKED ISSUE #1686: [Bug]: `Duration` is not updated after the table optimizing status change from `idle` to `pending`
### What happened?

The `Duration` is not updated after the table changed from `idle` to `pending`.
image

### Affects Versions

0.5.0

AMS

### How to reproduce

Without starting the optimizer, when keep inserting data into a table, the table's optimizing status will become `pending`, but the duration will not change.

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

wangtaohz:

fix #1686 

## Brief change log
  - update currentStatusStartTime after change to pending

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
@@ -170,6 +170,7 @@ public class TableRuntime extends StatedPersistentBase {
       this.pendingInput = pendingInput;
       if (optimizingStatus == OptimizingStatus.IDLE) {
         this.currentChangeSnapshotId = System.currentTimeMillis();
+        this.currentStatusStartTime = System.currentTimeMillis();
         this.optimizingStatus = OptimizingStatus.PENDING;
         persistUpdatingRuntime();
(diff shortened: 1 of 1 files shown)


ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
```diff
@@ -171,6 +171,7 @@ public void setPendingInput(OptimizingEvaluator.PendingInput pendingInput) {
       this.pendingInput = pendingInput;
       if (optimizingStatus == OptimizingStatus.IDLE) {
         this.currentChangeSnapshotId = System.currentTimeMillis();
+        this.currentStatusStartTime = System.currentTimeMillis();
```
zhoujinsong, line 174: If `currentStatusStartTime` and `optimizingStatus` always change together, would it be better to abstract a method to encapsulate their changes?

wangtaohz: Yes, it is. I will extract a method to handle the change.

zhoujinsong (approved):
LGTM.

Issue #1733, opened 2023-07-24
[Feature]: Support optimize tables with many eq-delete files through the partitioning process.


XBaith:
### Description

1. Support the number of tasks per process according to the current parallelism of the optimizer packaged , so that AMS can maximize the use of the optimizer to optimize more tables. For example, a table with a large number of delete file to do optimization takes a lot of time and the number of task exceeds  total optimizer parallism, it will take a long time to occupy all the resources of the optimizer, other tables can not be optimized in a timely manner, even if the table requires very little resources.
2. The optimization process of a table with too many eq-delete files can be partitioned, dividing a huge optimization process into several smaller processes, so that they can be merged in time to reduce the risk of failure and retrying.
3. Split processes and tasks can be executed sequentially according to certain sorting rules, such as file size, file count, sequence-number.

### Use case/motivation

![image](https://github.com/NetEase/arctic/assets/54210725/c73292fe-59ac-4bd5-b6fd-472466974199)
The above situation will happen if the table has a lot of delete file to merge.We should split the process in order to complete the optimization as soon as possible.

### Describe the solution

Split process partitioned and iterate them sequential.
```
                                        process-group-1 (partition 1) -> sort tasks   
splitted tasks -> group by partition -> process-group-2 (partition 2) -> sort tasks -> sort process -> group-1 complete
     ^                                  process-group-3 (partition 3) -> sort tasks                           |
     |                                                                                                        v
     |                                                                                                 group-2 complete
     |                                                                                                        |
     |                                                                                                        v
split task <---- planed tasks <----------------------- [Planer] <-------- wait for next plan <-------- group-3 complete
```

### Related issues

- #1729

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1734, opened 2023-07-24
[ARCTIC-1733] Implement splitable and ordered processes

XBaith:

implement #1733 

## Brief change log

1. Split large processes into smaller ones allows you to complete the table optimization and query enhancements to take effect as soon as possible.
2. Try to avoid retries caused by exceptions during the optimization process, such retries may waste resources
3. Order task and process 
    - file size asc: release optimizer resources earier
    - sequence number desc: optimize the latest updated files

## How was this patch tested?

## Documentation


XBaith:
cc @wangtaohz @Aireed

ams/server/src/main/java/com/netease/arctic/server/optimizing/UnKeyedTableCommit.java
```diff
@@ -186,14 +189,18 @@ protected void replaceFiles(
       if (CollectionUtils.isNotEmpty(removedDataFiles) ||
           CollectionUtils.isNotEmpty(addedDataFiles)) {
         RewriteFiles dataFileRewrite = transaction.newRewrite();
+        // original snapshot might be expired
+        Snapshot targetSnapshot = table.asUnkeyedTable().snapshot(targetSnapshotId) == null ?
+            table.asUnkeyedTable().currentSnapshot() :
+            table.asUnkeyedTable().snapshot(targetSnapshotId);
```
wangtaohz, line 195: The `targetSnapshot` should not be the current snapshot, which cannot achieve the goal of checking the commit conflicts.

Maybe we should open another PR to fix the commit failure issue caused by the expiration of the target snapshot.

XBaith: make sense

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -140,19 +132,16 @@ public void removeOptimizer(String resourceId) {
...
-
   @Override
   public void ping() {
   }
 
   @Override
   public void touch(String authToken) {
     OptimizerInstance optimizer = getAuthenticatedOptimizer(authToken).touch();
-    LOG.debug("Optimizer {} touch time: {}", optimizer.getToken(), optimizer.getTouchTime());
+    if (LOG.isDebugEnabled()) {
```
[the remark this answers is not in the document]
XBaith: I will chery pick this commit

shidayang:
@majin1102 cc

wangtaohz:
I have to make sure when a single plan generates many processes, how should these processes be displayed on the Web UI?

wangtaohz:
Another idea is that **generating only one process per plan** can be seen as a special case of **generating multiple processes per plan**. When users do not mind the cost of planning and want to plan every time to obtain the latest process, this approach can be considered.

XBaith:
> I have to make sure when a single plan generates many processes, how should these processes be displayed on the Web UI?

The mechanism for generating multiple processes is through lazy loading and only executes the next optimizing process until the current process completes. So will not display total processes immediately

XBaith:
Yes, getting a process after planning is a case of getting multiple processes during this implementation

wangtaohz:
> > I have to make sure when a single plan generates many processes, how should these processes be displayed on the Web UI?
> 
> The mechanism for generating multiple processes is through lazy loading and only executes the next optimizing process until the current process completes. So will not display total processes immediately

I think it would be more user-friendly to be able to see all the running processes on the Web UI, which better matches the original design concept. cc @majin1102 @zhoujinsong

XBaith:
> I think it would be more user-friendly to be able to see all the running processes on the Web UI, which better matches the original design concept. cc @majin1102 @zhoujinsong

There is only one **running** process for each table being optimized, other processes will be in the pending state, and a new process will not be created until the previous process is completed, so I think this should meet your original design?

PR #1738, opened 2023-07-25
[ARCTIC-1736] Fix SnapshotsExpiringExecutor will expire the snapshot which table used to plan

### LINKED ISSUE #1736: [Bug]: Optimizing's target snapshot should not be expired
### What happened?

image
The `fetchOptimizingSnapshotTime` in `SnapshotsExpiringExecutor` will get the table's snapshot that it planned, and does not expire it. But currently, it gets the current snapshot which may be changed in pending status.
This will cause snapshots that should not be expired to be expired.

It is the correct operation to obtain the target snapshot id of the optimizing process.

### Affects Versions

master

AMS

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

HuangFru:
Fix #1736.

## Brief change log

  - *Fetch optimizing snapshot time when the table is processing.*
  - *Fetch the target snapshot id in the optimizing process instead of the current snapshot id.*
  - *Modify the UT test.*

## How was this patch tested?

## Documentation


wangtaohz (approved):
LGTM

XBaith (approved):
LGTM

PR #821, opened 2022-12-02
[ARCTIC-820] Remove table configuration `self-optimizing.max-task-file-size`

wangtaohz:
## Why are the changes needed?

fix #820 

## Brief change log

  - remove `self-optimizing.max-task-file-size`
  - `BinPacking` of full-optimize/major-optimize use `self-optimizing.target-size` now

## How was this patch tested?
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] Run test locally before making a pull request

## Documentation

  - Does this pull request introduces a new feature? (no)
  - If yes, how is the feature documented? (not documented)

PR #1740, opened 2023-07-25
[ARCTIC-1729] IcebergPartitionPlan impl buildTaskSplitter by Bin-Packing

### LINKED ISSUE #1729: [Bug]: Optimizing task split has not been implemented
### What happened?

 Optimizing task split has not been implemented, so for the iceberg format, there will only one task per partition when doing optimizing.

### Affects Versions

0.5.0

AMS

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

huyuanfeng2018:

fix #1729 
IcebergPartitionPlan should implement buildTaskSplitter

## Brief change log

1. move MixedHivePartitionPlan Bin-Packing to AbstractPartitionPlan
2. IcebergPartitionPlan use Bin-Packing 

## How was this patch tested?

*before*:
image
*now*
image

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
@@ -356,8 +356,8 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
       // collect
       List<SplitTask> results = Lists.newArrayList();
       for (List<FileTask> fileTasks : packed) {
-        Map<IcebergDataFile, List<IcebergContentFile<?>>> fragmentFiles = com.google.common.collect.Maps.newHashMap();
-        Map<IcebergDataFile, List<IcebergContentFile<?>>> segmentFiles = com.google.common.collect.Maps.newHashMap();
+        Map<IcebergDataFile, List<IcebergContentFile<?>>> fragmentFiles = Maps.newHashMap();
+        Map<IcebergDataFile, List<IcebergContentFile<?>>> segmentFiles = Maps.newHashMap();
         fileTasks.stream().filter(FileTask::isFragment)
             .forEach(f -> fragmentFiles.put(f.getFile(), f.getDeleteFiles()));
         fileTasks.stream().filter(FileTask::isSegment)
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
@@ -27,8 +27,10 @@ import com.netease.arctic.server.optimizing.OptimizingType;
 import com.netease.arctic.server.table.TableRuntime;
 import com.netease.arctic.table.ArcticTable;
 import org.apache.iceberg.FileContent;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
+import org.apache.iceberg.util.BinPacking;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
```diff
@@ -302,4 +304,67 @@ public TaskDescriptor buildTask(OptimizingInputProperties properties) {
...
+      Long sum = allDataFiles.stream().map(f -> f.getFile().fileSizeInBytes()).reduce(0L, Long::sum);
+      int taskCnt = (int) (sum / taskSize) + 1;
+      List<List<FileTask>> packed = new BinPacking.ListPacker<FileTask>(taskSize, taskCnt, true)
+          .pack(allDataFiles, f -> f.getFile().fileSizeInBytes());
+
+      // collect
+      List<SplitTask> results = Lists.newArrayList();
+      for (List<FileTask> fileTasks : packed) {
+        Map<IcebergDataFile, List<IcebergContentFile<?>>> fragmentFiles = com.google.common.collect.Maps.newHashMap();
+        Map<IcebergDataFile, List<IcebergContentFile<?>>> segmentFiles = com.google.common.collect.Maps.newHashMap();
```
wangtaohz, line 360: We can use `org.apache.iceberg.relocated.com.google.common.collect.Maps` instead of `com.google.common.collect.Maps`.

shidayang:
Thanks for your contribution. BTW, You can associate the original issue using the syntax "fix: #1729 ".

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
```diff
@@ -302,4 +304,67 @@ public TaskDescriptor buildTask(OptimizingInputProperties properties) {
...
+
+    public boolean isSegment() {
+      return !isFragment;
+    }
+  }
+
+  protected class BinPackingTaskSplitter implements TaskSplitter {
+
+    @Override
+    public List<SplitTask> splitTasks(int targetTaskCount) {
```
shidayang, line 342: I think we should use the "targetTaskCount" parameter to control how to split. If we split according to "targetFileSize", the merged file size will definitely be smaller than or equal to "targetFileSize", which is not what we expected.

huyuanfeng2018:
> Thanks for your contribution. BTW, You can associate the original issue using the syntax "fix: #1729 ".

sry, I forgot

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
huyuanfeng2018: @wangtaohz  Can you reply to this question?

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
huyuanfeng2018: I will fix

wangtaohz:
> *before*:
> image
>*now*
> image

After the code modification, the average file size after optimizing seems to have decreased . I'm concerned that the tasks may have been split too small. Could you provide the value of this table property `self-optimizing.target-size`(which decides the split task size)? @huyuanfeng2018

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
wangtaohz: Using `targetFileSize` to split tasks has this flaw, which is an old issue inherited from 0.4.1 (PR #821). I believe that making the split task size `n` times larger than `targetFileSize` can solve this problem.

As for `targetTaskCount`, currently this parameter cannot guarantee that tasks are split into reasonable sizes, more work needs to be done in the generation of `targetTaskCount` to ensure that it is used reasonably for splitting.

huyuanfeng2018:
> > _before_:
> > image
> > _now_
> > image
> 
> After the code modification, the average file size after optimizing seems to have decreased . I'm concerned that the tasks may have been split too small. Could you provide the value of this table property `self-optimizing.target-size`(which decides the split task size)? @huyuanfeng2018

128mb

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
huyuanfeng2018: solved

zhoujinsong (approved):
LGTM and thanks a lot for your contribution!

wangtaohz (approved):

PR #1746, opened 2023-07-27
[ARCTIC-1745] fix the calculation of `actualInputSize` when skip some partitions for optimizing

### LINKED ISSUE #1745: [Bug]: The limit of the number of partitions processed in one Optimizing process doesn't work
### What happened?

According to the design, an optimizing process should not handle too much data at once to avoid resources being monopolized by a single optimizing process for a table. By default, this limit is set to 500MB. 

Under this limit, for partitioned tables, an optimizing process is allowed to process only a subset of partitions at a time. 

However, it was found during testing that an optimizing process processed multiple partitions and the data volume far exceeded 500MB, like

image

### Affects Versions

0.5.0

_No response_

### How to reproduce

1.create an partitioned iceberg format table
2.write to 3 partitions of the table, and write more than 500M data files to each partition
3.enable the optimizing of table
4.we could find an optimizing process handling data files for all partitions

### Relevant log output

_No response_

### Anything else

Bug in the calculation of actualInputSize.
image

### Code of Conduct

wangtaohz:

resolve #1745

## Brief change log

  - fix the calculation of `actualInputSize` when skip some partitions for optimizing

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM

Issue #1194, opened 2023-03-06
[Bug]: RocksDB JNI version doesn't support Apple Silicon/M1 chip


XBaith:
### What happened?

Cannot run rocksdb map locally on M1 chip(arm64)

### Affects Versions

master

### What engines are you seeing the problem on?

Core, AMS

### How to reproduce

mvn test

### Relevant log output

> [ERROR] Errors: 
[ERROR]   TaskReaderTest.testReadNegateWithSpillMap:259 » NoClassDefFound Could not initialize class org.rocksdb.DBOptions
[ERROR]   TaskReaderTest.testReadWithSpillMap:100 » UnsatisfiedLink /private/var/folders/kg/hplq6w3n2mz69s42bnpl_lq40000gn/T/librocksdbjni6521898922061163736.jnilib: dlopen(/private/var/folders/kg/hplq6w3n2mz69s42bnpl_lq40000gn/T/librocksdbjni6521898922061163736.jnilib, 0x0001): tried: '/private/var/folders/kg/hplq6w3n2mz69s42bnpl_lq40000gn/T/librocksdbjni6521898922061163736.jnilib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64')), '/System/Volumes/Preboot/Cryptexes/OS/private/var/folders/kg/hplq6w3n2mz69s42bnpl_lq40000gn/T/librocksdbjni6521898922061163736.jnilib' (no such file), '/private/var/folders/kg/hplq6w3n2mz69s42bnpl_lq40000gn/T/librocksdbjni6521898922061163736.jnilib' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))
[ERROR]   TestRocksDBBackend.testAddAndDropColumnFamily:25 » NoClassDefFound Could not initialize class org.rocksdb.DBOptions
[ERROR]   TestRocksDBBackend.testClose:90 » NoClassDefFound Could not initialize class org.rocksdb.DBOptions
[ERROR]   TestRocksDBBackend.testIterator:62 » NoClassDefFound Could not initialize class org.rocksdb.DBOptions
[ERROR]   TestRocksDBBackend.testPutGetDelete:35 » NoClassDefFound Could not initialize class org.rocksdb.DBOptions
[ERROR]   TestSimpleSpillableMap.testSpillableMap:44->testMap:58 » NoClassDefFound Could not initialize class org.rocksdb.DBOptions
[...]

PR #1752, opened 2023-07-28
[ARCTIC-1751] Update the memory map in spillable map when new value was existed for the key

### LINKED ISSUE #1751: [Bug]: Find duplicate records when enable rocksdb map
### What happened?

When user write to Iceberg table in flink upsert mode, and spillable map was enabled in Optimizer, it is possible that the equailty-delete ids do not take effect after self-optimization, meaning that duplicates may be generated

### Affects Versions

master/0.4.x

Optimizer

### How to reproduce

1. enable spillable map
2. start self-optimizing if reach the threshold for disk spillable 

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

XBaith:

fix #1751 and #1194 

## Brief change log

get corrent eq-delete file and the sequence number to make sure duplicate records are filtered

## How was this patch tested?

## Documentation

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -33,7 +33,7 @@
     <url>https://arctic.netease.com</url>
 
     <properties>
-        <rocksdb.version>5.17.2</rocksdb.version>
+        <rocksdb.version>7.10.2</rocksdb.version>
         <kryo.version>2.24.0</kryo.version>
         <jol.version>0.16</jol.version>
     </properties>
--- a/core/src/main/java/com/netease/arctic/utils/map/SimpleSpillableMap.java
+++ b/core/src/main/java/com/netease/arctic/utils/map/SimpleSpillableMap.java
@@ -111,6 +111,11 @@ public class SimpleSpillableMap<K, T> implements SimpleMap<K, T> {
       this.currentInMemoryMapSize = this.memoryMap.size() * this.estimatedPayloadSize;
     }
 
+    if (memoryMap.containsKey(key)) {
+      memoryMap.put(key, value);
+      return;
+    }
+
     if (this.currentInMemoryMapSize < maxInMemorySizeInBytes) {
       if (memoryMap.put(key, value) == null) {
         currentInMemoryMapSize += this.estimatedPayloadSize;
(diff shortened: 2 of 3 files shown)


core/pom.xml
```diff
@@ -33,7 +33,7 @@
     <url>https://arctic.netease.com</url>
 
     <properties>
-        <rocksdb.version>5.17.2</rocksdb.version>
+        <rocksdb.version>7.10.2</rocksdb.version>
```
wangtaohz, line 36: I test it locally on M1 chip(arm64), and it works. 👍

shidayang (approved):

zhongqishang:
Thanks for your great work! 

Regarding the issue of consistency, I think there may also be a problem in a scenario. After we delete some data, the `currentInMemoryMapSize` will become smaller, which will cause the data in the `diskBasedMap` to continue to be written to the `memoryMap`. When the deletion is triggered again, only the `memoryMap` will be deleted. The data in the `diskBasedMap` will continue to exist.

@shidayang @XBaith  WDYT.

shidayang:
> 

You are right,  So delete need to delete memory and disk

core/pom.xml
zhoujinsong: Why should we change the rocksdb version?
PR #1558  changes the version to `6.20.3-ververica-1.0`, I am not sure dose this will affect this PR.

core/src/main/java/com/netease/arctic/utils/map/SimpleSpillableMap.java
```diff
@@ -111,6 +111,11 @@ public void put(K key, T value) {
       this.currentInMemoryMapSize = this.memoryMap.size() * this.estimatedPayloadSize;
     }
 
+    if (memoryMap.containsKey(key)) {
```
zhongqishang, line 114: Maybe we can put the new logic in the else of the following condition (this.currentInMemoryMapSize < maxInMemorySizeInBytes) to reduce calculations.

XBaith:
> Regarding the issue of consistency, I think there may also be a problem in a scenario. After we delete some data, the `currentInMemoryMapSize` will become smaller, which will cause the data in the `diskBasedMap` to continue to be written to the `memoryMap`. When the deletion is triggered again, only the `memoryMap` will be deleted. The data in the `diskBasedMap` will continue to exist.

Make sense

core/pom.xml
XBaith: Getting rocksdb to work on M1 chip, which I mentioned in #1194

XBaith:
> I think this PR needs to be backport to 0.4.x. WDYT @shidayang @XBaith

+1, I will pick this commit to branch 0.4.x later

PR #1755, opened 2023-07-30
[ARCTIC-1754] self-optimizing.enabled change true, trigger tryEvaluatingPendingInput

### LINKED ISSUE #1754: [Bug]: When iceberg table self-optimizing.enabled change true, need to trigger tryEvaluatingPendingInput
### What happened?

Change the properties self-optimizing.enabled of a table from false to true, and if there is no new commit for this table, optimizing will not be triggered.

### Affects Versions

master

_No response_

### How to reproduce

Any time

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

huyuanfeng2018:

fix #1754 

## Brief change log
- Solve the issue of repeating execution of tasks on the table when the table configuration change.
- Change the condition of `evaluate` to enable the optimizing of the new enabled table.

## How was this patch tested?

## Documentation


ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java
```diff
@@ -51,6 +53,17 @@ public void handleStatusChanged(TableRuntime tableRuntime, OptimizingStatus orig
     tryEvaluatingPendingInput(tableRuntime, loadTable(tableRuntime));
   }
 
+  @Override
+  public void handleConfigChanged(TableRuntime tableRuntime, TableConfiguration originalConfig) {
```
huyuanfeng2018, line 57: This can solve the problem that Optimizing cannot be triggered when self-optimizing.enabled changes from false to true, but I think this part should be abstracted. There should be only one method call for tryEvaluatingPendingInput? A state should be maintained when handleConfigChanged is triggered, and it will be found in the next TableRuntimeRefresh And trigger, but unfortunately, the current TableRuntime cannot support this judgment

ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java
```diff
@@ -51,6 +53,17 @@ public void handleStatusChanged(TableRuntime tableRuntime, OptimizingStatus orig
...
+    OptimizingConfig originalOptimizingConfig = originalConfig.getOptimizingConfig();
+    OptimizingConfig newOptimizingConfig = tableRuntime.getOptimizingConfig();
+    if (!originalOptimizingConfig.equals(newOptimizingConfig) &&
+        newOptimizingConfig.isEnabled()) {
```
shidayang, line 61: Would it be better to use
```
if (!originalOptimizingConfig.isEnable &&
        newOptimizingConfig.isEnabled())
```?

File: ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java (hunk 1 above)
Comment (XBaith) [inline ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: This function will be called in `BaseTableExecutor` when optimizing configs are changed , but actually `TableRuntimeRefreshExecutor#execute` is the final operation. So if you evaluate pending input in `handleConfigChanged`, then the table will have a duplicate scan plan.

File: ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java (hunk 2 above)
Comment (huyuanfeng2018) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:61]: If it is simply to see whether it is turned on or off, it is enough to write this way, but I don’t know if other configuration changes under other originalOptimizingConfig will affect subsequent operations

File: ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java (hunk 1 above)
Comment (huyuanfeng2018) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: Yes, but I didn't find that there is a persistent variable in the `RuntimeTable` that can carry the ConfigChanged event to the `execute` method triggered by the next `TableRuntimeRefreshExecutor`

Comment (XBaith) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: Maybe check the optimizing state in `execute` function is better?

Comment (huyuanfeng2018) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: > Maybe check the optimizing state in `execute` function is better?

+1, But the key point is how to know in `execute` that the configuration update of the current table has occurred, so as to trigger the `tryEvaluatingPendingInput` method, maybe a persistent field or other solutions are required?

Comment (XBaith) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: before the tableRuntime refresh, we can get the original self-optimizing status

Comment (huyuanfeng2018) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: > before the tableRuntime refresh, we can get the original self-optimizing status

no, The essence of `handleConfigChanged` is triggered from `refresh`,

File: ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java (hunk 2 above)
Comment (wangtaohz) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:61]: I believe it's better to compare all optimizing configurations, because most configurations will affect the behavior of the plan.

**Comment by wangtaohz** [comment]
I have found another defect, but may not be related to this modification.

That is, when `isExecutable` always returns true, a periodically running task will be generated every time `scheduleIfNecessary` is called, resulting in an increasingly frequent check frequency. Unfortunately, it happens when `TableRuntimeRefreshExecutor#handleConfigChanged` is called.

So, I think we should fix it in this PR.

File: ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java (hunk 1 above)
Comment (zhoujinsong) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: I agree that we should limit the method `tryEvaluatingPendingInput` only be called in the executor thread pool and I found that `TableRuntimeRefreshExecutor#handleStatusChanged` had called the method already now which should be fixed too.

`BaseTableExecutor#scheduleIfNecessary` is used when you want to trigger a function on a table proactively and I think we should call it in `handleStatusChanged` and `handleConfigChanged`.

The problem is that when the current snapshot of the table is not changed, `tryEvaluatingPendingInput` will be skipped, and we may need to force the evaluating process in some way, like setting the `currentSnapshotId` to -1 or add a non-persistent variable to store this state.

Comment (baiyangtx) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: I think setting `currentSnapshotId` to -1 is not a good option. 

We can introduce an additional field, `lastOptimizedSnapshotId`, to record the snapshotId when optimizing was last executed.

**Comment by huyuanfeng2018** [comment]
>

File: ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java (hunk 1 above)
Comment (majin1102) [reply ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java:57]: 1. I agree that refresh operation should only happen in execute() method
2. handleStatusChanged() or handleConfigChanged() could only call scheduleIfNecessary()
3. enabled() is the key for RefreshingExecutor to determine whether a table should be refreshed. I don't think current impl is a good idea:
`  
@Override
  protected boolean enabled(TableRuntime tableRuntime) {
    return  loadTable(tableRuntime).originalTable() instanceof ArcticTable;
  }
`

File: ams/server/src/main/java/com/netease/arctic/server/table/executor/TableRuntimeRefreshExecutor.java [hunk 3]
```diff
@@ -48,7 +48,9 @@ protected long getNextExecutingTime(TableRuntime tableRuntime) {
 
   @Override
   public void handleStatusChanged(TableRuntime tableRuntime, OptimizingStatus originalStatus) {
-    tryEvaluatingPendingInput(tableRuntime, loadTable(tableRuntime));
+    if (tableRuntime.getOptimizingStatus().equals(OptimizingStatus.IDLE)) {
```
XBaith, line 51: The parameter `originalStatus` doesn't work here? Should we follow the definition of the interface and compare the original status with the current status?

I think AMS will call this method in the following case:
PENDING -> IDLE(when clean up pending inputs): there is no need for a processing task, but we have to scan the file again in `OptimizingEvaluator`?
Do I understand correctly?

zhoujinsong: Yes, I think you are right.

We can limit the evaluation process only if the status changed from processing to idle with codes like :
```
if (originalStatus != null && originalStatus.isProcessing() && tableRuntime.getOptimizingStatus().equals(OptimizingStatus.IDLE)) {

```

majin1102: agree with up floors

majin1102: It seems that status transformation is quite enumerable, and in my opinion, any status changed to `IDLE` could trigger evaluating, because `IDLE` status is the only status previous to evaluating(may be we shall define a status of evaluating in the future) and the condition is necessary and sufficient.

so I think the code is proper

majin1102 (approved):
LGTM

PR #1763, opened 2023-08-01
[ARCTIC-1762] Close the empty process when restarting AMS

### LINKED ISSUE #1762: [Bug]: Throw NPE during restarting AMS
### What happened?

If SQL session had closed when execute multiple SQL operations in one transaction, the process id may only persistence in databse `table_runtime`, but not in others(table_optimizing_process, task_runtime). The status of optimzing will block in `RUNNING`, so we have to clear them in the database and restart AMS. but AMS will not load any task based on the process id that exists only in the table meta, so an exception will occur.

### Affects Versions

master

AMS

### How to reproduce

1. only persistence process id in table_runtime and update optimzing status as Running
2. restart AMS

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

XBaith:

fix #1762 

## Brief change log

Enhance robustness

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
@@ -86,9 +86,10 @@ public class OptimizingQueue extends PersistentBase implements OptimizingService
 
   private void initTableRuntime(TableRuntimeMeta tableRuntimeMeta) {
     TableRuntime tableRuntime = tableRuntimeMeta.getTableRuntime();
+    TableOptimizingProcess process = new TableOptimizingProcess(tableRuntimeMeta);
     if (tableRuntime.getOptimizingStatus().isProcessing() &&
-        tableRuntimeMeta.getOptimizingProcessId() != 0) {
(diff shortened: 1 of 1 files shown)


ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -86,9 +86,10 @@ public OptimizingQueue(
 
   private void initTableRuntime(TableRuntimeMeta tableRuntimeMeta) {
     TableRuntime tableRuntime = tableRuntimeMeta.getTableRuntime();
+    TableOptimizingProcess process = new TableOptimizingProcess(tableRuntimeMeta);
```
shidayang, line 89: The new TableOptimizingProcess(tableRuntimeMeta) method will automatically call tableRuntime.recover(process) inside, so does this if statement still make sense?

XBaith: I have deleted it in class `TableOptimizingProcess `

PR #1764, opened 2023-08-01
[ARCTIC-1745] Only skip partitions when the maximum size limit is exceeded

### LINKED ISSUE #1745: [Bug]: The limit of the number of partitions processed in one Optimizing process doesn't work

wangtaohz:

fix #1745 , if the first partition's size > maxSize, it should not be skipped

## Brief change log

  - fix skip all partitions when the first partition's size > maxSize

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
@@ -9,6 +9,7 @@ import com.netease.arctic.ams.api.OptimizingService;
 import com.netease.arctic.ams.api.OptimizingTask;
 import com.netease.arctic.ams.api.OptimizingTaskId;
 import com.netease.arctic.ams.api.OptimizingTaskResult;
+import com.netease.arctic.ams.api.resource.Resource;
 import com.netease.arctic.ams.api.resource.ResourceGroup;
 import com.netease.arctic.optimizing.RewriteFilesInput;
 import com.netease.arctic.server.ArcticServiceConstants;
@@ -268,7 +269,7 @@ public class OptimizingQueue extends PersistentBase implements OptimizingService
       try {
         ArcticTable table = tableManager.loadTable(tableRuntime.getTableIdentifier());
         OptimizingPlanner planner = new OptimizingPlanner(tableRuntime.refresh(table), table,
-            getAvailableCore(tableRuntime));
+            getAvailableCore());
         if (tableRuntime.isBlocked(BlockableOperation.OPTIMIZE)) {
           LOG.info("{} optimize is blocked, continue", tableRuntime.getTableIdentifier());
           continue;
@@ -288,8 +289,8 @@ public class OptimizingQueue extends PersistentBase implements OptimizingService
     }
   }
 
-  private double getAvailableCore(TableRuntime tableRuntime) {
-    return tableRuntime.getOptimizingConfig().getTargetQuota();
+  private double getAvailableCore() {
+    return authOptimizers.values().stream().map(Resource::getThreadCount).reduce(Integer::sum).orElse(0);
   }
 
   @VisibleForTesting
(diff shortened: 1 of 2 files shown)


ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/OptimizingPlanner.java
```diff
@@ -40,7 +40,7 @@
 public class OptimizingPlanner extends OptimizingEvaluator {
   private static final Logger LOG = LoggerFactory.getLogger(OptimizingPlanner.class);
 
-  private static final long MAX_INPUT_FILE_SIZE_PER_THREAD = 5L * 1024 * 1024 * 1024;
+  private static final long MAX_INPUT_FILE_SIZE_PER_THREAD = 512 * 1024 * 1024; // 500MB
```
XBaith, line 43: 512MB?

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -288,8 +289,8 @@ private void planTasks() {
     }
   }
 
-  private double getAvailableCore(TableRuntime tableRuntime) {
-    return tableRuntime.getOptimizingConfig().getTargetQuota();
+  private double getAvailableCore() {
+    return authOptimizers.values().stream().map(Resource::getThreadCount).reduce(Integer::sum).orElse(0);
```
zhongqishang, line 293: Maybe we should match table group ?

wangtaohz: Do you mean the `Optimizer Group`? All the `authOptimizers` have the same `Optimizer Group` in the object of `OptimizingQueue`.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/OptimizingPlanner.java
wangtaohz: Yes, I will fix it.

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
zhongqishang: Yes, you are right. My fault.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/OptimizingPlanner.java
```diff
@@ -123,11 +123,11 @@ public List<TaskDescriptor> planTasks() {
     List<PartitionEvaluator> inputPartitions = Lists.newArrayList();
```
XBaith, line 123:
```
    double maxInputSize = MAX_INPUT_FILE_SIZE_PER_THREAD * availableCore;
```
`availableCore` should not be zero

wangtaohz: Okay, I think setting the minimum value of availableCore to 1 is reasonable.

XBaith: nerver mind, 0 `maxInputSize` doesn't really affect anything for planing

zhoujinsong (approved):
LGTM.

Issue #1783, opened 2023-08-02
[Improvement]: When the primary key definition is different from the schema fields order, the primary key serialization exception occurs in Lookup Join


YesOrNo828:
### Search before asking

- [X] I have searched in the [issues](https://github.com/NetEase/arctic/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Case :
table schema:
id int,
grade string,
num int

primary keys defined in the table:
grade, id

### How should we improve?

Avoid primary keys serializing failure during the initialization process;
Avoid NPE in the writing-rocksDB-threads

### Are you willing to submit PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #1784, opened 2023-08-02
[ARCTIC-1783][lookup] The primary key serialization exception occurs when the primary key definition is different from the schema fields order

YesOrNo828:

fixed: #1783 

## Brief change log

  - *Adjust the order of the primary keys and join keys when creating unique and join keys mapping index*
  - *The Flink 1.15 and 1.14 versions are affected.*
  - *Avoid NPE if the initialization process fails*
  - *Add some tests*

## How was this patch tested?

## Documentation

### DIFF
--- a/flink/v1.14/flink/src/main/java/com/netease/arctic/flink/lookup/KVTableFactory.java
+++ b/flink/v1.14/flink/src/main/java/com/netease/arctic/flink/lookup/KVTableFactory.java
@@ -21,6 +21,7 @@ package com.netease.arctic.flink.lookup;
 import org.apache.flink.configuration.Configuration;
 import org.apache.flink.table.data.RowData;
 import org.apache.iceberg.Schema;
+import org.apache.iceberg.types.Types;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -29,6 +30,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import static com.netease.arctic.flink.util.LookupUtil.convertLookupOptions;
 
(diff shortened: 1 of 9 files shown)


flink/v1.14/flink/src/main/java/com/netease/arctic/flink/lookup/RocksDBCacheState.java
```diff
@@ -71,6 +71,7 @@ public abstract class RocksDBCacheState<V> {
   protected BinaryRowDataSerializerWrapper valueSerializer;
   private ExecutorService writeRocksDBService;
   private final AtomicBoolean initialized = new AtomicBoolean(false);
+  private final AtomicBoolean running = new AtomicBoolean(true);
```
zhoujinsong, line 74: Is the variable name `closed` better here?
I do not think the class is runnable.

flink/v1.14/flink/src/main/java/com/netease/arctic/flink/lookup/KVTableFactory.java
```diff
@@ -46,6 +48,12 @@ public KVTable<RowData> create(
       Predicate<RowData> rowDataPredicate) {
     Set<String> joinKeySet = new HashSet<>(joinKeys);
     Set<String> primaryKeySet = new HashSet<>(primaryKeys);
+    // keep the primary keys order with projected schema fields.
+    primaryKeys = projectSchema.asStruct().fields().stream()
```
zhoujinsong, line 52: Should the order of `joinKeys` be adjust too here?

flink/v1.14/flink/src/main/java/com/netease/arctic/flink/lookup/RocksDBCacheState.java
YesOrNo828: Make sense to me.

flink/v1.14/flink/src/main/java/com/netease/arctic/flink/lookup/KVTableFactory.java
YesOrNo828: I add a test about this case.

zhoujinsong (approved):
LGTM.

PR #1793, opened 2023-08-03
[ARCTIC-1783][lookup] Ensure that projection schemas are in the same order

YesOrNo828:

fixed: #1783 
## Brief change log

- Adjust the order of the projected fields when creating projected schema from base schema;
- The Flink 1.15 and 1.14 versions are affected.

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

PR #1803, opened 2023-08-04
[AMORO-1789][Improvement]: modify the parameter transfer method for flink optimize

xujiangfeng001:

FIX: https://github.com/NetEase/amoro/issues/1789
FIX: https://github.com/NetEase/amoro/issues/1777

## Brief change log

When starting `flink optimize`, use `-yD <property=value>` to support more flink configuration parameters, such as `yarn.application.queue`.

Users can use the following methods:

1. jobmanager.memory / taskmanager.memory
2. jobmanager.memory.process.size / taskmanager.memory.process.size
3. default configuration in flink-conf.yaml

To configure memory resources.Priority from high to low.

## How was this patch tested?

## Documentation

### Use flink-conf.yaml setting
amoro setting:

WechatIMG49

flink-conf.yaml setting:

WechatIMG50

JobManager:

WechatIMG51

TaskManager:

WechatIMG52

### Use Old Parameter
amoro setting:

WechatIMG53

JobManager:

WechatIMG54

TaskManager:

WechatIMG55

### Use Flink Parameter

amoro setting

WechatIMG56

yarn queue:

WechatIMG57

JobManager:

WechatIMG58

TaskManager:

WechatIMG59


ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
```diff
@@ -72,15 +82,105 @@ protected Map<String, String> doScaleOut(String startUpArgs) {
...
+    Long taskManagerMemory = getMemorySizeValue(properties, flinkConfig, TASK_MANAGER_MEMORY_PROPERTY,
+        TASK_MANAGER_TOTAL_PROCESS_MEMORY);
     String jobPath = getAMSHome() + "/plugin/optimize/OptimizeJob.jar";
-    long memory = Long.parseLong(jobManagerMemory) + Long.parseLong(taskManagerMemory) * resource.getThreadCount();
-    return String.format("%s/bin/flink run -m yarn-cluster -ytm %s -yjm %s -c %s %s -m %s %s",
-        getFlinkHome(), taskManagerMemory, jobManagerMemory,
-        FlinkOptimizer.class.getName(), jobPath, memory,
-        super.buildOptimizerStartupArgsString(resource));
+    long memory = jobManagerMemory + taskManagerMemory * resource.getThreadCount();
+    return String.format("%s/bin/flink run -m yarn-cluster -ytm %s -yjm %s %s -c %s %s -m %s %s",
```
zhongqishang, line 95: This may not comply with the setting of flink's submit cmd.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
```diff
@@ -72,15 +82,105 @@ protected Map<String, String> doScaleOut(String startUpArgs) {
...
+          getFlinkHome(), taskManagerMemory, jobManagerMemory, buildFlinkArgs(properties),
+          FlinkOptimizer.class.getName(), jobPath, memory,
+          super.buildOptimizerStartupArgsString(resource));
+  }
+
+  private Map<String, String> loadFlinkConfig() {
+    try {
+      return new Yaml().load(Files.newInputStream(Paths.get(getFlinkHome() + FLINK_CONFIG_PATH)));
```
zhongqishang, line 103: The way we can load configuration using flink `org.apache.flink.configuration.GlobalConfiguration#loadConfiguration(java.lang.String)`.

zhongqishang (commented):
@xujiangfeng001 Thanks for your contribution.

I think we change `-m yarn-cluster` to `-t yarn-per-job`, so that we can unify the group optimizer configuration and flink config to be consistent. Such as :
```
./bin/flink run -t yarn-per-job --detached -Dtaskmanager.memory.process.size=1024m -Dyarn.application.name=Demo ./examples/streaming/TopSpeedWindowing.jar
```

The group optimizer has no configuration parameters and uses the parameters of flink-conf.yaml by default.
If configured we via `-D` to configure the optimizer.

At optimizer side we via env to get memory config and report to AMS.

WDYT?

wangtaohz:
👍 Thanks for your contribution. I will review it ASAP.

wangtaohz:
> I think we change `-m yarn-cluster` to `-t yarn-per-job`

@zhongqishang I found that the cmd of `-t yarn-per-job` was deprecated from flink 1.15[[flink doc](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/deployment/resource-providers/yarn/#per-job-mode-deprecated)], which may bring problems working with higher flink versions.

zhongqishang:
> > I think we change `-m yarn-cluster` to `-t yarn-per-job`
> 
> @zhongqishang I found that the cmd of `-t yarn-per-job` was deprecated from flink 1.15[[flink doc](https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/deployment/resource-providers/yarn/#per-job-mode-deprecated)], which may bring problems working with higher flink versions.

@wangtaohz Yes, The best way should be to use `Application mode`.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
xujiangfeng001: > The way we can load configuration using flink `org.apache.flink.configuration.GlobalConfiguration#loadConfiguration(java.lang.String)`.
> 
> such as :
> 
> ```
>     Configuration flinkConfig = GlobalConfiguration.loadConfiguration(getFlinkHome() + FLINK_CONFIG_PATH);;
>     MemorySize jobMemorySize = flinkConfig.get(JobManagerOptions.TOTAL_PROCESS_MEMORY);
>     MemorySize taskMemorySize = flinkConfig.get(TaskManagerOptions.TOTAL_PROCESS_MEMORY);
> ```

Thank you very much for your suggestion. I originally thought of using the `flink api` to complete these operations, but later I looked at `pom.xml` and found that all flink related dependencies are `<scope>provided</scope>`. Therefore, if I use the `flink api` to do so, it will result in missing package dependencies.

I have considered the requirements for this module flink jars come from the local environment of flink, so I don't want to add flink dependencies to `pom.xml` to cause the package to become larger or have dependency conflicts, so I chose to implement the relevant functions myself.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
xujiangfeng001: > This may not comply with the setting of flink's submit cmd.

Why does it not meet the requirements， can you provide a detailed explanation? Thank you。

xujiangfeng001:
Hi @wangtaohz @zhongqishang ，I don't think we should discuss the issue of running mode here. We can create a new feature to discuss whether it is necessary to enable `flink optimize` to support multiple running modes, such as `yarn application`, `per job`, or `K8s`.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
zhongqishang: Sorry for this, it's my missread.

wangtaohz:
> Hi @wangtaohz @zhongqishang ，I don't think we should discuss the issue of running mode here. We can create a new feature to discuss whether it is necessary to enable `flink optimize` to support multiple running modes, such as `yarn application`, `per job`, or `K8s`.

I agree with that.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
shidayang, line 95: Why don't we directly reuse Flink's parameter configuration for memory-related parameters? Why do we need to recalculate them?

xujiangfeng001: Firstly, the purpose of calculating memory in this area is to display resource usage on the front-end.

Secondly, due to the possibility of users using old parameters in history:` jobmanager.memory`，`taskmanager.memory`, in order to be compatible these users did not directly use the flink parameter to calculate the memory situation.

This is also a point that needs to be discussed, whether it is necessary to be compatible with the parameters configured by old users: `jobmanager.memory`、`taskmanager.memory`

wangtaohz (approved):

wangtaohz (approved):
LGTM.

shidayang (approved):

xujiangfeng001:
> Thanks a lot for your geat work! The codes look good to me totally.
> 
> But can you help to change the docs to help users understand your improvement? You can find related doc in `/docs/admin-guides/managing-optimizers.md`.

Hi @zhoujinsong , I have added feature introduction in `/docs/admin-guides/managing-optimizers.md`.

docs/admin-guides/managing-optimizers.md
```diff
@@ -80,16 +80,16 @@ The following configuration needs to be filled in:
...
 | scheduling-policy   | All | No | quota | The scheduler group scheduling policy, the default value is `quota`, it will be scheduled according to the quota resources configured for each table, the larger the table quota is, the more optimizer resources it can take. There is also a configuration `balanced` that will balance the scheduling of each table, the longer the table has not been optimized, the higher the scheduling priority will be. |
-| taskmanager.memory   | flink | Yes | N/A | The memory size of Flink TaskManager. |
-| jobmanager.memory    | flink | Yes | N/A | The memory size of Flink JobManager.  |
-| memory   | local | Yes | N/A | The memory size of the local optimizer Java process. |
+| taskmanager.memory   | flink | No | N/A | The memory size of Flink TaskManager. Deprecated, suggested use `taskmanager.memory.process.size`. If not set, the parameters in `flink-conf.yaml` will be used by default.                                                                                                                                                                                                                                      |
```
zhoujinsong, line 90: As property `taskmanager.memory` and `jobmanager.memory` is deprecated, I suggest removing them from the supported properties.

We can describe the supported properties for flink container like this:
```
| flink-conf.* | flink | No | N/A | Any configuration for `flink on yarn` mode, like `flink-conf.taskmanager.memory.process.size` or `flink-conf.jobmanager.memory.process.size`. The value in `conf/flink-conf.yaml` will be used if not setted here. You can find more supported property in [Flink Configuration](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/config/)
```

zhoujinsong: When I try to write the document, I found we may add a prefix for the flink properties like `flink-conf`.

If we want to support some properties not used by Flink,  we can identify them then.

What do you think? @xujiangfeng001

xujiangfeng001: > https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/config/

This suggestion looks good, I'll modify the logic.

xujiangfeng001: > As property `taskmanager.memory` and `jobmanager.memory` is deprecated, I suggest removing them from the supported properties.
> 
> We can describe the supported properties for flink container like this:
> 
> ```
> | flink-conf.* | flink | No | N/A | Any configuration for `flink on yarn` mode, like `flink-conf.taskmanager.memory.process.size` or `flink-conf.jobmanager.memory.process.size`. The value in `conf/flink-conf.yaml` will be used if not setted here. You can find more supported property in [Flink Configuration](https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/config/)
> ```

Do we need to be compatible with `taskmanager. memory` and `jobmanager. memory` in the code here.

zhoujinsong: Yes, we can keep the compatible code for one major version and remove it from 0.7.0.
You do better use a `@Deprecated` annotation to tell that in code.

xujiangfeng001: OKey.

xujiangfeng001: Hi @zhoujinsong , PTAL.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
```diff
@@ -72,15 +96,98 @@ protected Map<String, String> doScaleOut(String startUpArgs) {
...
+                .map((v) -> parseMemorySize(v))
+                .orElse(0L)));
+  }
+
+  /**
+   * build user configured flink native parameters
+   */
+  protected String buildFlinkArgs(Map<String, String> properties) {
+    final Set<String> keysToFilter = new HashSet<>(Arrays.asList(
+        JOB_MANAGER_MEMORY_PROPERTY,
```
zhoujinsong, line 149: It seems `JOB_MANAGER_MEMORY_PROPERTY` and `TASK_MANAGER_MEMORY_PROPERTY` are no longer needed to be filtered.

And should we use a static final variable to store the filter set?

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
```diff
@@ -72,15 +96,98 @@ protected Map<String, String> doScaleOut(String startUpArgs) {
...
+  public long parseMemorySize(String memoryStr) {
+    if (memoryStr == null || memoryStr.isEmpty()) {
+      return 0;
+    }
+    memoryStr = memoryStr.toLowerCase().trim().replaceAll("\\s", "");
+    Matcher matcher = Pattern.compile("(\\d+)([mg])").matcher(memoryStr);
+    if (matcher.matches()) {
+      long size = Long.parseLong(matcher.group(1));
+      String unit = matcher.group(2);
+      switch (unit) {
```
zhongqishang, line 174: I'm a little worried about compatibility issues. You can refer to the analysis of flink conf [1], of course, this has satisfied the most commonly used unit (m, g).

[1]: https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/configuration/MemorySize.java#L352-L356

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
```diff
@@ -25,29 +25,53 @@
...
   public static final String YARN_APPLICATION_ID_PROPERTY = "yarn-application-id";
 
   private static final Pattern APPLICATION_ID_PATTERN = Pattern.compile("(.*)application_(\\d+)_(\\d+)");
   private static final int MAX_READ_APP_ID_TIME = 600000; //10 min
 
 
   public String getFlinkHome() {
-    return PropertyUtil.checkAndGetProperty(getContainerProperties(), FLINK_HOME_PROPERTY);
+    String flinkHome = PropertyUtil.checkAndGetProperty(getContainerProperties(), FLINK_HOME_PROPERTY);
+    return  flinkHome != null ? flinkHome.replaceAll("/$", "") : null;
```
zhongqishang, line 74: Need format code.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
xujiangfeng001: > I'm a little worried about compatibility issues. You can refer to the parse of flink conf [1](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/configuration/MemorySize.java#L352-L356). Of course, this has satisfied the most commonly used unit (m, g).

At present, it is indeed only supported for `M` and `G`, because I believe there is no need to be compatible with units as small as `KB`. I believe that even testing such small resources is not enough, let alone production environments.

So I only support `M` and `G`.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
xujiangfeng001: There is no need for `JOB_MANAGER_MEMORY_PROPERTY` and `TASK_MANAGER_MEMORY_PROPERTY ` here, which is my negligence. But I don't think it's necessary to use static final variables for storage, it feels meaningless.

xujiangfeng001: > It seems `JOB_MANAGER_MEMORY_PROPERTY` and `TASK_MANAGER_MEMORY_PROPERTY` are no longer needed to be filtered.
> 
> And should we use a static final variable to store the filter set?

There is no need for `JOB_MANAGER_MEMORY_PROPERTY` and `TASK_MANAGER_MEMORY_PROPERTY`  here, which is my negligence. But I don't think it's necessary to use static final variables for storage, it feels meaningless.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
xujiangfeng001: > Need format code.

Well, I have modified the code style.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/FlinkOptimizerContainer.java
zhoujinsong: My original idea was to add a static final member variable in the Class to store the filter set, in order to avoid constructing a temporary variable every time it is called. However, since this method itself is called frequently, keeping the current implementation is also okay. What do you think?

xujiangfeng001: > My original idea was to add a static final member variable in the Class to store the filter set, in order to avoid constructing a temporary variable every time it is called. However, since this method itself is called frequently, keeping the current implementation is also okay. What do you think?

Thank you for your explanation. I think this is a good idea. I'll modify the code.

zhoujinsong (approved):
LGTM.

zhongqishang (approved):
LGTM.

PR #1811, opened 2023-08-07
[AMORO-1807] When I upgrade a hive table and it fails, it deletes the original hive table

### LINKED ISSUE #1807: [Bug]: When I upgrade a hive table and it fails, it deletes the original hive table.
### What happened?

First I upgraded a hive table to a mixed hive table.
image
image
You can tell I failed at this step.
image
However when I refreshed the page to load the table it no longer existed, later I went to hms to check that the table had been deleted before.
image

### Affects Versions

master

Core, AMS

### How to reproduce

1. Create a hive table and upgrade it, and let it fail.
2. Check the hive table

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

nicochen:

fix #1807 

## Brief change log
Added a logic that deletes the hive table, but does not delete the amoro table

## How was this patch tested?

## Documentation

### DIFF
--- a/hive/src/main/java/com/netease/arctic/hive/catalog/MixedHiveTables.java
+++ b/hive/src/main/java/com/netease/arctic/hive/catalog/MixedHiveTables.java
@@ -234,21 +234,36 @@ public class MixedHiveTables extends MixedTables {
     super.dropTableByMeta(tableMeta, purge);
     // drop hive table operation will only delete hive table metadata
     // delete data files operation will use BasicArcticCatalog
-    try {
-      hiveClientPool.run(client -> {
-        client.dropTable(tableMeta.getTableIdentifier().getDatabase(),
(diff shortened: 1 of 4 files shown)


hive/src/main/java/com/netease/arctic/hive/catalog/MixedHiveTables.java
```diff
@@ -247,6 +247,10 @@ public void dropTableByMeta(TableMeta tableMeta, boolean purge) {
     }
   }
 
+  public void dropTableMetaButNotDropHiveTable(TableMeta tableMeta) {
```
zhoujinsong, line 250: Does this method appear to be necessary?
You can just call `MixedTables#dropTableByMeta` directly.

zhoujinsong (approved):
LGTM. Thanks a lot for your contribution.

PR #1821, opened 2023-08-08
[AMORO-1820]: Fix spark drop namespace error in rest catalog services.

### LINKED ISSUE #1820: [Bug]: terminal can't drop database
### What happened?

drop database under some  iceberg catalog
image

image

### Affects Versions

master

Spark

### How to reproduce

drop database under some  iceberg catalog

### Relevant log output

```shell
drop database under some  iceberg catalog
```

### Anything else

_No response_

### Code of Conduct

baiyangtx:

fix #1820 

## Brief change log

- Return empty list when listNamespace passes a database name

## How was this patch tested?

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/IcebergRestCatalogService.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/IcebergRestCatalogService.java
@@ -38,6 +38,7 @@ import org.apache.iceberg.exceptions.NoSuchNamespaceException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.io.FileIO;
 import org.apache.iceberg.relocated.com.google.common.base.Preconditions;
+import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.rest.RESTResponse;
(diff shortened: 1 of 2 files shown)


baiyangtx:
image

checked

ams/server/src/main/java/com/netease/arctic/server/IcebergRestCatalogService.java
```diff
@@ -137,10 +137,11 @@ public void handleException(Exception e, Context ctx) {
     jsonResponse(ctx, response);
     if (code.code >= 500) {
       LOG.warn("InternalServer Error", e);
+    } else {
```
zhoujinsong, line 140: Why do we need to reduce the logging level for errors in which the code is smaller than 500 here?

baiyangtx: because those errors happened when the client-side passed arguments with problems. 
The service side usually does not need to log those errors.

zhoujinsong: Got it.
Maybe add some comments here to tell that.

zhoujinsong (approved):
Validate in my local envrionment.

PR #1825, opened 2023-08-09
[ARCTIC-1628] Rename `Independent` delete files to `Dangling` delete files

### LINKED ISSUE #1628: [Feature]: Remove the Independent `DeleteFiles` for the Iceberg Format Table
### Description

For the Iceberg Format table, there are some scenarios that produce some DeleteFiles that are not related to any DataFile , which are called Independent DeleteFiles. These scenarios include: 
- when upsert mode is enabled, a record which not exist will be deleted before inserting it
- delete files that are not automatically cleaned up after rewriting files
- ...

Although these Independent DeleteFiles cannot be found when scanning files and do not affect read performance, they can cause file accumulation and put pressure on the file system.

We should remove these Independent `DeleteFiles` periodically.

### Use case/motivation

In our scenario, after several days of testing, we were able to find thousands of Independent DeleteFiles, including both equ-delete Files and pos-delete Files, even far more than the number of Data Files. Removing these files can greatly reduce the total number of files on the disk.

### Describe the solution

I suggest removing Independent DeleteFiles during Orphan files cleanup because Independent DeleteFiles can be considered a special type of Orphan Files. 

Of course, a table property, like `clean-independent-delete-files.enabled` (default true), should be introduced to control whether to clean Independent DeleteFiles for the table.

We can further discuss whether this implementation is appropriate.

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

wangtaohz:

resolve #1628 

## Brief change log

  - rename `Independent` delete files to `Dangling` delete files
  - rename `arctic` to `amoro` for derby location

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/table/executor/OrphanFilesCleaningExecutor.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/table/executor/OrphanFilesCleaningExecutor.java
@@ -113,15 +113,15 @@ public class OrphanFilesCleaningExecutor extends BaseTableExecutor {
       // clear metadata files
       cleanMetadata(arcticTable, System.currentTimeMillis() - keepTime);
 
-      boolean needIndependentClean = CompatiblePropertyUtil.propertyAsBoolean(arcticTable.properties(),
-          TableProperties.ENABLE_INDEPENDENT_CLEAN,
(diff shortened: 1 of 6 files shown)


zhoujinsong:
Change the configuration in docs too.

zhoujinsong (approved):
LGTM.

PR #1850, opened 2023-08-16
[AMORO-1845][Optimizer] Add calculate optimizer resource allocation on optimizer side

### LINKED ISSUE #1845: [Improvement]: Show correct resource occupation for external optimizer
### Search before asking

Now the resources reported by external optimizer are configured through the -m parameter. If you modify the resource without modifying the `-m` parameter, it will cause AMS to display wrong resource occupation.

### How should we improve?

Calculate resource utilization in optimizer, override `-m` parameter.

### Subtasks

_No response_

### Code of Conduct

zhongqishang:

fix #1845 

## Brief change log

  - *Add calculate optimizer resource allocation on optimizer side*

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
+++ b/ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
@@ -53,7 +53,7 @@ public class LocalOptimizerContainer extends AbstractResourceContainer {
     long memoryPerThread = Long.parseLong(PropertyUtil.checkAndGetProperty(resource.getProperties(),
         JOB_MEMORY_PROPERTY));
     long memory = memoryPerThread * resource.getThreadCount();
-    return String.format("%s/bin/localOptimize.sh %s %s", getAMSHome(), memory,
+    return String.format("%s/bin/localOptimize.sh -m %s %s", getAMSHome(), memory,
         super.buildOptimizerStartupArgsString(resource));
   }
 
--- a/ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
+++ b/ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
@@ -38,7 +38,6 @@ public class LocalOptimizerContainer extends AbstractResourceContainer {
 
   @Override
   protected Map<String, String> doScaleOut(String startUpArgs) {
-    Runtime runtime = Runtime.getRuntime();
     try {
       String[] cmd = {"/bin/sh", "-c", startUpArgs};
       LOG.info("Starting local optimizer using command : {}", startUpArgs);
--- a/ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
+++ b/ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
@@ -53,7 +53,7 @@ public class LocalOptimizerContainer extends AbstractResourceContainer {
     long memoryPerThread = Long.parseLong(PropertyUtil.checkAndGetProperty(resource.getProperties(),
         JOB_MEMORY_PROPERTY));
     long memory = memoryPerThread * resource.getThreadCount();
-    return String.format("%s/bin/localOptimize.sh -m %s %s", getAMSHome(), memory,
+    return String.format("%s/bin/localOptimize.sh %s %s", getAMSHome(), memory,
         super.buildOptimizerStartupArgsString(resource));
   }
 
--- a/ams/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
+++ b/ams/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
@@ -4,6 +4,10 @@ import com.netease.arctic.optimizer.Optimizer;
 import com.netease.arctic.optimizer.OptimizerConfig;
 import org.apache.flink.api.common.typeinfo.Types;
 import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.JobManagerOptions;
+import org.apache.flink.configuration.MemorySize;
+import org.apache.flink.configuration.ReadableConfig;
+import org.apache.flink.configuration.TaskManagerOptions;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.streaming.api.functions.sink.DiscardingSink;
 import org.kohsuke.args4j.CmdLineException;
@@ -20,6 +24,10 @@ public class FlinkOptimizer {
   public static void main(String[] args) throws CmdLineException {
     StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment(new Configuration());
     OptimizerConfig optimizerConfig = new OptimizerConfig(args);
+
+    // calculate optimizer memory allocation
+    calcOptimizerMemory(optimizerConfig, env);
+
     Optimizer optimizer = new Optimizer(optimizerConfig);
     env.addSource(new FlinkToucher(optimizer.getToucher()))
         .setParallelism(1)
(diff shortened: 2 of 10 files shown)


wangtaohz (commented):
Thanks for your contribution! 👍
This PR also made me realize that the previous calculation of memory was wrong, in the case of TaskManager with slot>1, which is
```
    long memory = jobManagerMemory + taskManagerMemory * resource.getThreadCount();
```.

I think we should remove this erroneous code to avoid causing confusion for other developers

**Comment by wangtaohz** [comment]
Furthermore, `LocalOptimizer` may also adopt a similar approach to return memory usage during the optimizer's runtime, in order to unify the mechanism for calculating memory usage for optimizers.

How do you think about it? @zhongqishang

**Comment by zhongqishang** [comment]
> I think we should remove this erroneous code to avoid causing confusion for other developers

If this logic is removed, how should we set the value of `-m`?

> Furthermore, LocalOptimizer may also adopt a similar approach to return memory usage during the optimizer's runtime, in order to unify the mechanism for calculating memory usage for optimizers.

I agree. The user may submit the LocalOptimizer themselves and set an incorrect value.

**Comment by wangtaohz** [comment]
> If this logic is removed, how should we set the value of `-m`?

I think we can set it to 0 for `FlinkOptimizer`, and this `-m` config will be deprecated if `LocalOptimizer` no longer depends on it.

File: ams/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java [hunk 1]
```diff
@@ -36,4 +44,25 @@ public static void main(String[] args) throws CmdLineException {
...
+    ReadableConfig configuration = env.getConfiguration();
+    MemorySize jobMemorySize = configuration.get(JobManagerOptions.TOTAL_PROCESS_MEMORY);
+    MemorySize taskMemorySize = configuration.get(TaskManagerOptions.TOTAL_PROCESS_MEMORY);
+    if (jobMemorySize == null || taskMemorySize == null) {
+      // Running locally will not get the configuration.
+      return;
+    }
+    int parallelism = config.getExecutionParallel();
+    int numberOfTaskSlots = configuration.get(TaskManagerOptions.NUM_TASK_SLOTS);
+    int memorySize = jobMemorySize.getMebiBytes() + (parallelism / numberOfTaskSlots) * taskMemorySize.getMebiBytes();
```
shidayang, line 58: The TaskManager's count cannot be simply calculated using parallelism/numberOfTaskSlots. The capacity of the TaskManager is related to many configuration parameters in Flink. Can the calculation method used in Flink be reused here?

wangtaohz: In the following code, when calculating the TaskManager's count, it takes into account rounding up the result of `parallelism/numberOfTaskSlots`.

Can you provide more details of mechanisms which may affect the count of TaskManagers? @shidayang

zhongqishang:
> this -m config will be deprecated if LocalOptimizer no longer depends on it.

`localOptimize.sh` depends on the -m parameter setting `-Xmx`, I think it is still needed.

wangtaohz:
> localOptimize.sh depends on the -m parameter setting -Xmx, I think it is still needed.

I checked the `localOptimize.sh`, 
```
JAVA_OPTS="-Xmx$2m -Dlog.home=${LOG_DIR} -Dlog.subdir=localOptimizer-$(date +%s)"
```
we could see that the `-Xmx` is set by the secondary parameter, which happens to be the memory after `-m`, but it has nothing to do with `-m`.

```
    return String.format("%s/bin/localOptimize.sh -m %s %s", getAMSHome(), memory,
        super.buildOptimizerStartupArgsString(resource));
```

So the `-m` can be deprecated, and after taht, we can directly use the `-Xmx` to set the memory. Of course, some modifications need to be made to `localOptimize.sh`.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
shidayang: The Flink code has been changed significantly and I couldn't find it，so this calculation method is also okay.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/local/LocalOptimizer.java
```diff
@@ -12,6 +12,11 @@ public class LocalOptimizer {
   public static void main(String[] args) throws CmdLineException {
     OptimizerConfig optimizerConfig = new OptimizerConfig(args);
     Optimizer optimizer = new Optimizer(optimizerConfig);
+
+    // calculate optimizer memory allocation
+    long memorySize = Runtime.getRuntime().maxMemory() / 1024 / 1024;
+    optimizerConfig.setMemorySize((int) memorySize);
+
```
wangtaohz, line 19: I tested it locally.
image
Although there are some differences compared to -Xmx, I think the results are acceptable. 👍

zhongqishang: Yes, my local test also has a little gap. And the gap is not big, it feels acceptable.

wangtaohz:
We would appreciate it if you could provide some test results(like screenshots) to demonstrate the correctness of the logic for computing Flink optimizer memory, as it cannot be covered by unit tests. Could you help with this? @zhongqishang

ams/optimizer/src/main/java/com/netease/arctic/optimizer/LocalOptimizerContainer.java
```diff
@@ -54,7 +53,7 @@ protected String buildOptimizerStartupArgsString(Resource resource) {
     long memoryPerThread = Long.parseLong(PropertyUtil.checkAndGetProperty(resource.getProperties(),
         JOB_MEMORY_PROPERTY));
     long memory = memoryPerThread * resource.getThreadCount();
-    return String.format("%s/bin/localOptimize.sh -m %s %s", getAMSHome(), memory,
+    return String.format("%s/bin/localOptimize.sh %s %s", getAMSHome(), memory,
```
wangtaohz, line 56: After remove `-m`, The unit test cases in `TestOptimizerConfig` also needs adjustment.

zhongqishang: Done.

zhongqishang:
The external optimizer parameters are as follows:
```
-a thrift://172.16.58.151:1261 -p 5 -g flinkGroup -id abcbc38d-9774-46c7-8b31-8dda48d1cf0c -hb 60000 -eds -msz 2048 -dsp /tmp
```
Memory
image

Optimizer
image

wangtaohz (approved):
LGTM

ams/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
```diff
@@ -36,4 +44,25 @@ public static void main(String[] args) throws CmdLineException {
       LOG.error("Execute flink optimizer failed", e);
     }
   }
+
+  private static void calcOptimizerMemory(OptimizerConfig config, StreamExecutionEnvironment env) {
+    ReadableConfig configuration = env.getConfiguration();
+    MemorySize jobMemorySize = configuration.get(JobManagerOptions.TOTAL_PROCESS_MEMORY);
+    MemorySize taskMemorySize = configuration.get(TaskManagerOptions.TOTAL_PROCESS_MEMORY);
+    if (jobMemorySize == null || taskMemorySize == null) {
+      // Running locally will not get the configuration.
```
zhoujinsong, line 53: Is `Running locally` means deploying the flink job in the standalone cluster?

Besides, how can we calculate the memory usage in this situation if we have deprecated the `-m` configuration?

docs/admin-guides/managing-optimizers.md
```diff
@@ -135,4 +135,4 @@ The description of the relevant parameters is shown in the following table:
 | -a AMS_THRIFT_SERVER_URL | The address of the AMS thrift service, for example: thrift://127.0.0.1:1261, can be obtained from the config.yaml configuration. |
 | -g OPTIMIZE_GROUP_NAME | Group name created in advance under external container. |
 | -p EXECUTOR_PARALLELISM | Optimizer parallelism usage. |
-| -m EXECUTOR_MEMORY | Optimizer memory usage. |
\ No newline at end of file
+| -m EXECUTOR_MEMORY | @Deprecated Optimizer memory usage. |
```
zhoujinsong, line 138: We may remove this configuration from docs if we have decided to remove it in the future.

zhoujinsong (commented):
@zhongqishang This PR seems good to me, but I still have some question, please take another look when you have time.

ams/optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkOptimizer.java
zhongqishang: > Is `Running locally` means deploying the flink job in the standalone cluster?

`Running locally` means running the optimizer in the IDEA. This method is supported in all deployment modes.

> Besides, how can we calculate the memory usage in this situation if we have deprecated the `-m` configuration?

In this situation, there is indeed no coverage. We can also obtain the usage of jvm in the same way as the local optimizer. Do you think it is necessary?

zhoujinsong: Ok, I don't think we will run the flink optimizer in the IDEA.

If we do, we still can use an argument to specify the memory usage.

zhoujinsong (approved):

PR #1859, opened 2023-08-20
[WIP] [AMORO-1771] Add two line charts to the Table Transactions page, showing the count of table records and the count of table files.

### LINKED ISSUE #1771: [Subtask]: Implement displaying metrics on the front end of AMS
### Search before asking

Implement displaying metrics on the front end of AMS

### Parent issue

#1528 

### Code of Conduct

minteliuwm:

resolve #1771 

## Brief change log

- Add two line charts to the Table Transactions page, showing the count of table records and the count of table files.

## How was this patch tested?

## Documentation


shidayang:
Thanks for your contributions, Ccould you provide a screenshot of the modified style?

minteliuwm:
> Thanks for your contributions, Ccould you provide a screenshot of the modified style?

Of course, I posted the screenshot in the comments, it's implemented as [the designed page](https://modao.cc/proto/STt1Oi0iryn5xuAHPPghVf/sharing?view_mode=device&screen=rbpTlYTQTSKqw8Dzp&canvasId=rcTlZ5icwip4cCnq), a little different is the x-axis shows the full date and time, if you think something is not implemented correctly please feel free to suggest it!
Snipaste_2023-08-21_22-14-30

wangtaohz:
If I understand correctly, each point in the graph corresponds to a Transaction, right? @minteliuwm 

I'm concerned that displaying these points uniformly in the x direction may cause distortion, because the time intervals between Transactions are not fixed.

minteliuwm:
> Transaction

@wangtaohz 

Yes, currently each point on the graph represents the commit time of a transaction.
I assume you want to use the x-axis as a timeline over a period of time and then show the corresponding data at specific points in time?
I'm not quite sure of the business logic here, if you want to see where the data is going on a time trend through this chart, I think your idea is fine, but there are just two areas that might need a bit of attention:
1. Currently our data is in milliseconds, do we need to be accurate to milliseconds here? If it is not accurate to milliseconds, there may be multiple data within the same time, is it necessary to consolidate the data in this case?
2. The time range of the return may be relatively large, resulting in some of the close time points being crowded on the graph, is this acceptable behaviour?

hameizi:
> Thank you for your prompt reply!
> 
> As for the two questions you mentioned, for the first one, I think accuracy to the second is sufficient because based on my experience, it is unlikely that multiple Transactions would be committed within one second.
> 
> As for the second question, I think it is acceptable.
> 
> @hameizi Could you provide some suggestions on how these two issues were considered during design?

We have discussed the display logic for this section before, and there is a theoretical possibility of submitting a large number of snapshots in a short time. Therefore, the current display logic is: only the snapshot of the current page is displayed, and the distance between the snapshots in the chart is equal.

wangtaohz:
> We have discussed the display logic for this section before, and there is a theoretical possibility of submitting a large number of snapshots in a short time. Therefore, the current display logic is: only the snapshot of the current page is displayed, and the distance between the snapshots in the chart is equal.

Make sense to me. 👍

In the scenario where only Transactions from the current page are displayed, I agree with the current implementation.

zhoujinsong (approved):
Validate in my local environment.

PR #1867, opened 2023-08-22
[AMORO-1866] Replace binpack task strategy from "fragment file first" with "segment file first".

### LINKED ISSUE #1866: [Bug]: Segment file first binpack strategy leads to inability to optimize SplitTasks
### What happened?

image
OptimizeEvalutor found some partitions that need to optimize, but Planer didn't get any tasks.

### Affects Versions

master

AMS

### How to reproduce

1. enable self-optimizing for the Iceberg table which has a certain number of segment files
2. write some fragment files

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

XBaith:

Fix #1866 

## Brief change log

Putting a "fragment file first" binpack strategy will prioritize packing fragment files, and then only try to find the segement file once the fragment file is almost packed, which will try to avoid having only one fragment file in a bin, which can lead to certain tasks being filtered.

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
@@ -341,20 +341,18 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
     @Override
     public List<SplitTask> splitTasks(int targetTaskCount) {
       // bin-packing
-      List<FileTask> allDataFiles = Lists.newArrayList();
-      segmentFiles.forEach((dataFile, deleteFiles) ->
-          allDataFiles.add(new FileTask(dataFile, deleteFiles, false)));
+      List<FileTask> allDataFiles = Lists.newLinkedList();
       fragmentFiles.forEach((dataFile, deleteFiles) ->
           allDataFiles.add(new FileTask(dataFile, deleteFiles, true)));
+      segmentFiles.forEach((dataFile, deleteFiles) ->
+          allDataFiles.add(new FileTask(dataFile, deleteFiles, false)));
 
-      long taskSize = config.getTargetSize();
-      Long sum = allDataFiles.stream().map(f -> f.getFile().fileSizeInBytes()).reduce(0L, Long::sum);
-      int taskCnt = (int) (sum / taskSize) + 1;
-      List<List<FileTask>> packed = new BinPacking.ListPacker<FileTask>(taskSize, taskCnt, true)
+      List<List<FileTask>> packed = new BinPacking.ListPacker<FileTask>(
+          config.getTargetSize(), Integer.MAX_VALUE, false)
           .pack(allDataFiles, f -> f.getFile().fileSizeInBytes());
 
       // collect
-      List<SplitTask> results = Lists.newArrayList();
+      List<SplitTask> results = Lists.newArrayListWithCapacity(packed.size());
       for (List<FileTask> fileTasks : packed) {
         Map<IcebergDataFile, List<IcebergContentFile<?>>> fragmentFiles = Maps.newHashMap();
         Map<IcebergDataFile, List<IcebergContentFile<?>>> segmentFiles = Maps.newHashMap();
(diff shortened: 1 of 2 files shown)


ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/OptimizingPlanner.java
```diff
@@ -117,7 +117,6 @@ public List<TaskDescriptor> planTasks() {
 
     List<PartitionEvaluator> evaluators = new ArrayList<>(partitionPlanMap.values());
     evaluators.sort(Comparator.comparing(PartitionEvaluator::getWeight));
-    Collections.reverse(evaluators);
```
wangtaohz, line 120: What is the reason for changing the sorting from descending to ascending by Weight?

XBaith: I think prioritizing partitions with low cost will free up the optimizer's resources as soon as possible, plus it will try to process as many partitions as possible in one more process.

wangtaohz: To my knowledge, we prioritize the partitions with high cost to avoid starvation. 
This is because the cost of a partition is expected to decrease after optimizing it. Therefore, if we prioritize partitions with low cost, it may cause high cost partitions to never optimizing.

wangtaohz: Considering this risk, I suggest keeping the original sorting order. What do you think?

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
```diff
@@ -341,20 +341,18 @@ protected class BinPackingTaskSplitter implements TaskSplitter {
...
           allDataFiles.add(new FileTask(dataFile, deleteFiles, true)));
+      segmentFiles.forEach((dataFile, deleteFiles) ->
+          allDataFiles.add(new FileTask(dataFile, deleteFiles, false)));
 
-      long taskSize = config.getTargetSize();
-      Long sum = allDataFiles.stream().map(f -> f.getFile().fileSizeInBytes()).reduce(0L, Long::sum);
-      int taskCnt = (int) (sum / taskSize) + 1;
-      List<List<FileTask>> packed = new BinPacking.ListPacker<FileTask>(taskSize, taskCnt, true)
+      List<List<FileTask>> packed = new BinPacking.ListPacker<FileTask>(
+          config.getTargetSize(), Integer.MAX_VALUE, false)
```
shidayang, line 351: Why lookback set to Max

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/OptimizingPlanner.java
XBaith: Make sense

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
XBaith: Setting it to the maximum value ensures that the maximum number of bin sizes can be packed to match the target size.

XBaith: I think in most of cases, there are no difference between using `Integer.MAX_VALUE` or `taskCnt`.WDYT

shidayang: I am worried that Integer.MAX_VALUE may cause performance issues with binpacking when there are a large number of files. We can test it and if there are no problems, I think it's okay.

XBaith: Okay, do you think setting the lookback to `1` is acceptable?

XBaith: Okay, let's keep the original lookback size

wangtaohz: I agree with using `Integer.MAX_VALUE`, because `Integer.MAX_VALUE` and `taskCnt` are essentially the same, and `taskCnt` worked well before.

wangtaohz (approved):

PR #1879, opened 2023-08-23
[AMORO-1870]: Support table filters to make AMS ignore tables that are not needed

### LINKED ISSUE #1870: [Feature]: Support table filters to make AMS ignore tables that are not needed
### Description

I want table filter ,  but now catalog  database tables is full show

### Use case/motivation

I want table filter ,  but now catalog  database tables is full show,so add table filter 

### Describe the solution

add a table filter to list  I want see the table  

### Subtasks

_No response_

### Related issues

no

### Code of Conduct

celltobig:

Resolve #1870.
AMS has a database filter, but not a table filter.

## Brief change log

  - Add table filter configuration for catalog

## How was this patch tested?

## Documentation

  - Does this pull request introduce a new feature? (yes )
  - If yes, how is the feature documented? (documented)


wangtaohz:
Thanks for your work! You can add resolve #1879 to relate to the issue.

celltobig:
image
image
image

wangtaohz:
The code looks good to me, but there are some check style issues to fix.
You could run `mvn checkstyle:check` locally to check the code before committing. @celltobig

.gitignore
```diff
@@ -64,3 +64,4 @@ conf/
 *unit_test_base_tmp/
 
 */.gitignore
+local/conf/
```
wangtaohz, line 67: I tested locally and found that this directory can be ignored by the configuration on line 61, so there is no need to add additional information to the `.gitignore`.

```
➜  arctic git:(master) git check-ignore -v local/conf
.gitignore:61:conf/     local/conf
```

which means that the directory named `conf` in the root directory and all its subdirectories will be ignored.

baiyangtx:
Thank you for submitting your Pull Request.

I'm considering using two different configuration items to decide the database and table filters separately, is it a good style? Can we use one configuration item to solve this problem?
For example, using a package name filter instead of a regular expression, such as ods_*.user_*

@zhoujinsong What's your point of view?

zhoujinsong:
Yes, there are indeed configuration complexities and limitations when using two expressions to filter databases and tables.

My idea is to make the configuration `table.filter-regular-expression` apply to both the database name and table name, like:

- `db1/..+` means tables under database `db1`
- `a/.A|b/.B` means `a.A` and `b.B`

f designed in this way, ·table.filter-regular-expression· can replace the existing ·database.filter-regular-expression·, but we can keep a major version to be compatible with the old configuration.

HDYT? @wangtaohz @baiyangtx @celltobig

wangtaohz:
I think it's a very feasible way to do it. However, there are two small issues:
1. The `/.` may be `\.` ?  And it should be like
- `db1\..+` means tables under database `db1`
- `a\.A|b\.B` means `a.A` and `b.B`

2.How about changing the `table.filter-regular-expression` to `database.table.filter-regular-expression`?  
It may be clearer to indicate matching both the database name and table name

zhoujinsong:
> 1. The `/.` may be `\.` ?  And it should be like
Sure, I have changed it in the comment.

> 2.How about changing the `table.filter-regular-expression` to `database.table.filter-regular-expression`? It may be clearer to indicate matching both the database name and table name
I do not like the configuration name `database.table.filter-regular-expression` and it may be clear enough if we expose the only configuration `table.filter-regular-expression` in docs and add some explanation.

docs/admin-guides/managing-catalogs.md
```diff
@@ -49,6 +49,7 @@ Common properties include:
 - warehouse: Warehouse **must be configured**, as it determines where our database and table files should be placed
 - catalog-impl: when the metastore is **Custom**, an additional catalog-impl must be defined, and the user must put the jar package for the custom catalog implementation into the **{ARCTIC_HOME}/lib** directory, **and the service must be restarted to take effect**
 - table.*: If you want to add the same table configuration to all tables under a catalog, you can add `table.` before the configuration key to indicate that it is a table-level configuration. For example, `table.self-optimizing.group`
+- table.filter-regular-expression: If you want to filter the same table under a database, you can add `table.filter-regular-expression` before the configuration key to indicate that it is a table-level configuration. For example, `table.filter-regular-expression`='(A\.a)|(B\.b)'
```
wangtaohz, line 52: You can modify the docs to:

- table.filter-regular-expression: Configure a regular expression to filter tables under databases. The matching will be done in the format of `db.table`. For example, if it is set to '(A\.a)|(B\.b)', it will ignore all tables except for table `a` in database `A` and table `b` in database `B`.

wangtaohz (approved):
LGTM

docs/admin-guides/managing-catalogs.md
```diff
@@ -49,6 +49,7 @@ Common properties include:
 - warehouse: Warehouse **must be configured**, as it determines where our database and table files should be placed
 - catalog-impl: when the metastore is **Custom**, an additional catalog-impl must be defined, and the user must put the jar package for the custom catalog implementation into the **{ARCTIC_HOME}/lib** directory, **and the service must be restarted to take effect**
 - table.*: If you want to add the same table configuration to all tables under a catalog, you can add `table.` before the configuration key to indicate that it is a table-level configuration. For example, `table.self-optimizing.group`
+- table.filter-regular-expression: Configure a regular expression to filter tables under databases. The matching will be done in the format of `db.table`. For example, if it is set to `(A\.a)|(B\.b)`, it will ignore all tables except for table `a` in database `A` and table `b` in database `B`
```
zhoujinsong, line 52: It seems the `table.filter-regular-expression` name may be confused with the default configuration for tables.
Would `table-filter` be better as the configuration name?

And I recommend moving it upper to `table.*` description.

ams/api/src/main/java/com/netease/arctic/ams/api/properties/CatalogMetaProperties.java
```diff
@@ -41,7 +41,10 @@ public class CatalogMetaProperties {
   @Deprecated
   public static final String KEY_WAREHOUSE_DIR = "warehouse.dir";
   public static final String KEY_WAREHOUSE = "warehouse";
+  //Deprecated from version v0.7.0, use KEY_TABLE_FILTER
```
zhoujinsong, line 44:
```suggestion
/**
   * @deprecated since 0.6.0, will be removed in 0.7.0; use {@link CatalogMetaProperties#KEY_TABLE_FILTER} instead.
   */
```

docs/admin-guides/managing-catalogs.md
```diff
@@ -48,6 +48,7 @@ You can create a catalog in the AMS frontend:
 Common properties include:
 - warehouse: Warehouse **must be configured**, as it determines where our database and table files should be placed
 - catalog-impl: when the metastore is **Custom**, an additional catalog-impl must be defined, and the user must put the jar package for the custom catalog implementation into the **{ARCTIC_HOME}/lib** directory, **and the service must be restarted to take effect**
+- table-filter: Configure a regular expression to filter tables under databases. The matching will be done in the format of `db.table`. For example, if it is set to `(A\.a)|(B\.b)`, it will ignore all tables except for table `a` in database `A` and table `b` in database `B`
```
zhoujinsong, line 51:
```suggestion
- table-filter: Configure a regular expression to filter tables in the catalog. The matching will be done in the format of `database.table`. For example, if it is set to `(A\.a)|(B\.b)`, it will ignore all tables except for table `a` in database `A` and table `b` in database `B`
```

PR #1881, opened 2023-08-23
 [AMORO-1880] Mixed Hive Table can't Sync Hive data properly

### LINKED ISSUE #1880: [Bug]: Mixed Hive Table can't Sync Hive data properly
### What happened?

image
As you can see ArcticTableFlag will be set to true when a Hive partition has previously written any data through Amoro.
But when I delete the data from this partition, write it again with Hive and try to synchronize it to the Mixed Hive table, the files cannot be added to the Mixed Hive table with this “if” logic, because there is no data in this partition of the Mixed Hive table so filesMap.get(partitionData) == null at the same time 
ArcticTableFlag exists because the Hive partition has not been deleted and data has been written to it.
So I think there is a problem with this logic.

### Affects Versions

master

Core

### How to reproduce

1. Create a Mixed Hive Table with partition
2. Insert overwrite some data
3. Delete the data insert overwrite before
4. Insert into the same data with Hive 
5. Use HiveDataSync to sync step4's data to Mixed Hive Table

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

nicochen:

fix #1880 

## Brief change log

The files in the Hive partition need to be synchronized into the Mixed Hive Table anyway.

## How was this patch tested?

## Documentation

### DIFF
--- a/hive/src/main/java/com/netease/arctic/hive/utils/HiveMetaSynchronizer.java
+++ b/hive/src/main/java/com/netease/arctic/hive/utils/HiveMetaSynchronizer.java
@@ -197,12 +197,9 @@ public class HiveMetaSynchronizer {
                 hivePartition.getSd().getLocation());
             if (filesMap.get(partitionData) != null) {
               filesToDelete.addAll(filesMap.get(partitionData));
-              filesToAdd.addAll(hiveDataFiles);
               // make sure new partition is not created by arctic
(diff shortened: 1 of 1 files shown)


czy006:
@nicochen The code has been modified because it has not been updated for a long time. If this problem persists, try using version 0.7. Or you can turn ISSUSE feedback back on.

PR #1886, opened 2023-08-24
[AMORO-1883] Perform file filtering when adding files to `PartitionEvaluator` instead of during splitting tasks

### LINKED ISSUE #1883: [Improvement]: Perform file filtering as early as possible when during optimizing plan process
### Search before asking

Currently, the optimizing plan evaluates all files first, then splits tasks, and performs file filtering after task splitting. This is not intuitive and can cause several issues, such as:

- The evaluating result is different from that after split tasks, which causes the pending data to be different from the actual optimizing data
- Bin-packing split task is inaccurate because some files will be filtered out later
- Verifying which DataFile files were involved in the optimizing requires recalculating after splitting tasks, which results in performance issues

### How should we improve?

We should filter files as early as possible: filtering should be done before splitting tasks and not be done again after.

### Subtasks

_No response_

### Code of Conduct

wangtaohz:

fix #1883 

## Brief change log

  - filter files when adding files to `PartitionEvaluator` and `AbstractPartitionPlan`
  - split tasks based on rewriteFiles and rewritePosFiles instead of fragment files and segment files
  - remove `taskNeedExecute` after splitting tasks
  - remove `addPartitionProperties` from `PartitionEvaluator`, add it to the constructor

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
@@ -107,6 +107,8 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
       rewriteDataFiles.put(dataFile, deletes);
     } else if (evaluator().segmentFileShouldRewritePos(dataFile, deletes)) {
       rewritePosDataFiles.put(dataFile, deletes);
+    } else {
+      return false;
     }
     return true;
   }
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
@@ -52,7 +52,7 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
 
   protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> rewriteDataFiles = Maps.newHashMap();
   protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> rewritePosDataFiles = Maps.newHashMap();
-  // protected Delete files are Delete files related to Data files not optimized in this plan
+  // protected Delete files are Delete files which are related to Data files not optimized in this plan
   protected final Set<String> protectedDeleteFiles = Sets.newHashSet();
 
   public AbstractPartitionPlan(TableRuntime tableRuntime,
@@ -99,7 +99,7 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
   public boolean addFile(IcebergDataFile dataFile, List<IcebergContentFile<?>> deletes) {
     boolean added = evaluator().addFile(dataFile, deletes);
     if (!added) {
-      // if the Data file is not added, it's Delete files should be not be removed from iceberg
+      // if the Data file is not added, it's Delete files should not be removed from iceberg
       deletes.stream().map(delete -> delete.path().toString()).forEach(protectedDeleteFiles::add);
       return false;
     }
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
@@ -26,6 +26,7 @@ import com.netease.arctic.server.optimizing.OptimizingConfig;
 import com.netease.arctic.server.optimizing.OptimizingType;
 import com.netease.arctic.server.table.TableRuntime;
 import com.netease.arctic.table.ArcticTable;
+import com.netease.arctic.utils.TablePropertyUtil;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
@@ -49,6 +50,7 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
   private long fromSequence = INVALID_SEQUENCE;
   private long toSequence = INVALID_SEQUENCE;
   protected final long planTime;
+  protected final Map<String, String> partitionProperties;
 
   protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> rewriteDataFiles = Maps.newHashMap();
   protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> rewritePosDataFiles = Maps.newHashMap();
@@ -62,6 +64,7 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
     this.config = tableRuntime.getOptimizingConfig();
     this.tableRuntime = tableRuntime;
     this.planTime = planTime;
+    this.partitionProperties = TablePropertyUtil.getPartitionProperties(table, partition);
   }
 
   @Override
@@ -77,7 +80,7 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
   }
 
   protected CommonPartitionEvaluator buildEvaluator() {
-    return new CommonPartitionEvaluator(tableRuntime, partition, planTime);
+    return new CommonPartitionEvaluator(tableRuntime, partition, partitionProperties, planTime);
   }
 
   @Override
@@ -111,11 +114,6 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
       return false;
     }
     return true;
-  }
-
-  @Override
-  public void addPartitionProperties(Map<String, String> properties) {
-    evaluator().addPartitionProperties(properties);
   }
 
   public List<TaskDescriptor> splitTasks(int targetTaskCount) {
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
@@ -26,7 +26,6 @@ import com.netease.arctic.server.optimizing.OptimizingConfig;
 import com.netease.arctic.server.optimizing.OptimizingType;
 import com.netease.arctic.server.table.TableRuntime;
 import com.netease.arctic.table.ArcticTable;
-import org.apache.iceberg.FileContent;
 import org.apache.iceberg.relocated.com.google.common.collect.Lists;
 import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.iceberg.relocated.com.google.common.collect.Sets;
@@ -51,12 +50,10 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
   private long toSequence = INVALID_SEQUENCE;
   protected final long planTime;
 
-  protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> fragmentFiles = Maps.newHashMap();
-  protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> segmentFiles = Maps.newHashMap();
-  protected final Map<String, Set<IcebergDataFile>> equalityDeleteFileMap = Maps.newHashMap();
-  protected final Map<String, Set<IcebergDataFile>> posDeleteFileMap = Maps.newHashMap();
-  
-  private List<SplitTask> splitTasks;
+  protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> rewriteDataFiles = Maps.newHashMap();
+  protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> rewritePosDataFiles = Maps.newHashMap();
+  // protected Delete files are Delete files related to Data files not optimized in this plan
+  protected final Set<String> protectedDeleteFiles = Sets.newHashSet();
 
   public AbstractPartitionPlan(TableRuntime tableRuntime,
                                ArcticTable table, String partition, long planTime) {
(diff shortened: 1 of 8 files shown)


ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
```diff
@@ -51,12 +50,10 @@ public abstract class AbstractPartitionPlan implements PartitionEvaluator {
...
-  protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> fragmentFiles = Maps.newHashMap();
-  protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> segmentFiles = Maps.newHashMap();
-  protected final Map<String, Set<IcebergDataFile>> equalityDeleteFileMap = Maps.newHashMap();
-  protected final Map<String, Set<IcebergDataFile>> posDeleteFileMap = Maps.newHashMap();
-  
-  private List<SplitTask> splitTasks;
+  protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> rewriteDataFiles = Maps.newHashMap();
+  protected final Map<IcebergDataFile, List<IcebergContentFile<?>>> rewritePosDataFiles = Maps.newHashMap();
+  // protected Delete files are Delete files related to Data files not optimized in this plan
+  protected final Set<String> protectedDeleteFiles = Sets.newHashSet();
```
zhoujinsong, line 56: The name `protectedDeleteFiles` is somewhat ambiguous. Would `keepedDeleteFiles` or `reservedDeleteFiles` be more appropriate?

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
```diff
@@ -99,24 +96,19 @@ public long getCost() {
...
-      } else {
-        equalityDeleteFileMap
-            .computeIfAbsent(deleteFile.path().toString(), delete -> Sets.newHashSet())
-            .add(dataFile);
-      }
+    if (evaluator().fileShouldRewrite(dataFile, deletes)) {
+      rewriteDataFiles.put(dataFile, deletes);
+    } else if (evaluator().segmentFileShouldRewritePos(dataFile, deletes)) {
+      rewritePosDataFiles.put(dataFile, deletes);
     }
```
zhoujinsong, line 110: It seems it is possible a data file can not match both conditions here, do we should handle this situation?

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/PartitionEvaluator.java
```diff
@@ -36,7 +36,7 @@ interface Weight extends Comparable<Weight> {
 
   String getPartition();
 
-  void addFile(IcebergDataFile dataFile, List<IcebergContentFile<?>> deletes);
+  boolean addFile(IcebergDataFile dataFile, List<IcebergContentFile<?>> deletes);
```
zhoujinsong, line 39: Could you help to add Java doc for this interface?

zhoujinsong (commented):
I left some comments, PTAL.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/PartitionEvaluator.java
wangtaohz: Sure, I will add come docs.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
wangtaohz: Considering the implementation of the Evaluator, this scenario should not occur, but it is quite unclear. 

We have two ways to handle this situation, throwing an exception or returning false. I prefer the latter, what do you think?

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
wangtaohz: OK, I prefer `reservedDeleteFiles`.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -213,8 +248,11 @@ protected boolean reachMinorInterval() {
   }
 
   protected boolean reachFullInterval() {
-    return config.getFullTriggerInterval() >= 0 &&
-        planTime - tableRuntime.getLastFullOptimizingTime() > config.getFullTriggerInterval();
+    if (reachFullInterval == null) {
+      reachFullInterval = config.getFullTriggerInterval() >= 0 &&
+          planTime - tableRuntime.getLastFullOptimizingTime() > config.getFullTriggerInterval();
```
zhongqishang, line 253: How about putting it in the constructor?

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
zhoujinsong: LGTM.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
zhoujinsong: Returning false seems good, but we should put the delete files path into the `reservedDeleteFiles` too.

wangtaohz: OK, it will.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -105,36 +108,66 @@ private boolean isDuplicateDelete(IcebergContentFile<?> delete) {
...
+  public boolean segmentFileShouldRewritePos(IcebergDataFile dataFile, List<IcebergContentFile<?>> deletes) {
+    if (isFullOptimizing()) {
+      return false;
+    }
+    if (isFragmentFile(dataFile)) {
+      return false;
+    }
     if (deletes.stream().anyMatch(delete -> delete.content() == FileContent.EQUALITY_DELETES)) {
       return true;
     } else if (deletes.stream().filter(delete -> delete.content() == FileContent.POSITION_DELETES).count() >= 2) {
```
XBaith, line 173: Why we need at least 2 position deletes?

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/PartitionEvaluator.java
zhoujinsong: Can you add Java doc for other methods and the class too?

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -105,36 +108,66 @@ private boolean isDuplicateDelete(IcebergContentFile<?> delete) {
...
+    }
+    // if a file is related any delete files or is not big enough, it should full optimizing
+    return !deleteFiles.isEmpty() || dataFile.fileSizeInBytes() < config.getTargetSize() * 0.9;
+  }
+
+  public boolean fileShouldRewrite(IcebergDataFile dataFile, List<IcebergContentFile<?>> deletes) {
+    if (isFullOptimizing()) {
+      return fileShouldFullOptimizing(dataFile, deletes);
+    }
+    if (isFragmentFile(dataFile)) {
```
XBaith, line 158: Sometimes there are some empty files(0 bytes), I think we should skip them, WDYT

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -143,6 +176,10 @@ public boolean shouldRewritePosForSegmentFile(IcebergDataFile dataFile, List<Ice
       return false;
     }
   }
+  
+  protected boolean isFullOptimizing() {
```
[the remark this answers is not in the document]
wangtaohz: Full optimizing is enabled no not is is determined by the trigger.interval >=0, which has already been checked here.

```
this.reachFullInterval = config.getFullTriggerInterval() >= 0 &&
        planTime - tableRuntime.getLastFullOptimizingTime() > config.getFullTriggerInterval();
```

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
wangtaohz: The purpose here is to keep the segment file unchanged and generate a pos-delete for it. If there is only one pos- delete file, there is no need to regenerate it.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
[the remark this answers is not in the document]
XBaith: If the user updates the full optimizating trigger interval to a number greater than 0, there may still be a risk of misjudgment

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
wangtaohz: Self-optimizing has not yet considered how to deal with empty files, so we can open a new issue to discuss and address it.
My current view is that an empty file should be included as a fragment file for optimizing, to avoid this file lingering in the metadata indefinitely.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -105,36 +108,66 @@ private boolean isDuplicateDelete(IcebergContentFile<?> delete) {
...
+    }
     return getRecordCount(deletes) > dataFile.recordCount() * config.getMajorDuplicateRatio();
   }
 
-  public boolean shouldRewritePosForSegmentFile(IcebergDataFile dataFile, List<IcebergContentFile<?>> deletes) {
+  public boolean segmentFileShouldRewritePos(IcebergDataFile dataFile, List<IcebergContentFile<?>> deletes) {
+    if (isFullOptimizing()) {
+      return false;
+    }
+    if (isFragmentFile(dataFile)) {
```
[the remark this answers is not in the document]
wangtaohz: As above.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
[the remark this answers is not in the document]
XBaith: If we new this Class after check optimizing enabled, it should no risk. Maybe we could consider introducing additional preconditions to help mitigate the risk of inadvertent calls.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
XBaith: OKay

XBaith:
Link to issue #1866

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/OptimizingEvaluator.java
```diff
@@ -94,19 +94,20 @@ protected TableFileScanHelper.PartitionFilter getPartitionFilter() {
 
   private void initPartitionPlans(TableFileScanHelper tableFileScanHelper) {
     PartitionSpec partitionSpec = arcticTable.spec();
-    for (TableFileScanHelper.FileScanResult fileScanResult : tableFileScanHelper.scan()) {
-      StructLike partition = fileScanResult.file().partition();
-      String partitionPath = partitionSpec.partitionToPath(partition);
-      PartitionEvaluator evaluator = partitionPlanMap.computeIfAbsent(partitionPath, this::buildEvaluator);
-      evaluator.addFile(fileScanResult.file(), fileScanResult.deleteFiles());
-    }
+    // add partition properties first, then add files, because files may use these properties
```
zhongqishang, line 97: The evaluator is generated in the following logic,
```
partitionPlanMap.computeIfAbsent(partitionPath, this::buildEvaluator);
```
so the evaluator is always null here, which will not make the properties take effect ?

wangtaohz: Yes, you are right.👍
My negligence, I will fix it immediately.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/PartitionEvaluator.java
wangtaohz: Sure, of course.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/OptimizingEvaluator.java
wangtaohz: I moved adding partition properties to the constructor, to avoid adding files before adding properties.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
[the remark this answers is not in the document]
wangtaohz: Okay, I will add such a check: If optimization is not enabled, ignore any files added to it.

zhongqishang (approved):
LGTM.

zhongqishang (approved):

zhoujinsong (approved):
LGTM.

PR #1895, opened 2023-08-29
[Improvement] Polish optimizing executor code

shidayang:
There are some flaws in the code.

## Brief change log

- Remove useless code
- Add annotation to some interfaces

## How was this patch tested?

## Documentation

- If yes, how is the feature documented? (not applicable / docs / **JavaDocs** / not documented)


core/src/main/java/com/netease/arctic/optimizing/OptimizingExecutor.java
```diff
@@ -20,6 +20,14 @@
...
 
+/**
+ * Self-optimizing executor for file merging, such as merging small files and converting equality-delete to
+ * position-delete.
+ */
 public interface OptimizingExecutor<O extends TableOptimizing.OptimizingOutput> extends Serializable {
+
+  /**
+   * Execute compaction for {@link com.netease.arctic.optimizing.TableOptimizing.OptimizingInput}
+   */
```
zhoujinsong, line 31: Would it be better to extend `java.io.Closeable`?

core/src/main/java/com/netease/arctic/optimizing/OptimizingDataReader.java
```diff
@@ -21,10 +21,23 @@
...
   CloseableIterable<Record> readData();
 
+  /**
+   * Reading data that needs to be deleted during MOR. If there is no delete file, then the result is empty.
+   */
   CloseableIterable<Record> readDeletedData();
 
+  /**
+   * Close the reader
+   */
```
zhoujinsong, line 41: Would it be better to extend `java.io.Closeable`?

shidayang: yeah

wangtaohz (approved):

PR #1901, opened 2023-08-30
[AMORO-1758] Support for table data expiration

### LINKED ISSUE #1758: [Feature]: Support for expiring partitions in Iceberg tables
### Description

After the user configures the time parameter for partition expiration, Arctic can automatically expire the data in these partitions and clean up the corresponding files.

### Use case/motivation

Many users partition their data by time and usually only keep data within a certain time frame. Data that exceeds this time frame needs to be expired and cleaned up to free up corresponding storage.

### Describe the solution

_No response_

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

XBaith:

Close #1758 .

## Brief change log

- Automatically remove expired partitions or files

## How was this patch tested?

## Documentation


XBaith:
Design document: 
- Chinese: https://docs.google.com/document/d/1Enkpmzu8xIKDTVb4-AUFePg3orbjRrQY4FXIS1uzAJ4/edit#heading=h.qpxw9thhgbj0

zhoujinsong:
@XBaith You may want to add a `[WIP]` prefix in this PR title as you have not finished the unit test and document part.

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
```diff
@@ -0,0 +1,449 @@
...
+                e -> willNotRetain(e, expirationConfig, partitionFreshness))
+            .forEach(expiredFiles::addFile);
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+      expireFiles(unkeyedTable, snapshot.snapshotId(), expiredFiles, expireTimestamp);
+    }
+  }
+
+  private static Expression getPartitionExpression(DataExpirationConfig expirationConfig, long expireTimestamp) {
```
huyuanfeng2018, line 224: How about we generate Expression through `org.apache.iceberg.expressions.Literal`, which has a lot of conversion logic built in. WDYT?

huyuanfeng2018 (commented):
PTAL

XBaith:
> How about we generate Expression through org.apache.iceberg.expressions.Literal, which has a lot of conversion logic built in. WDYT?

Thanks for your review! There are some good comments, I will fix them soon. And can you explain more details about how to generate Expression through `org.apache.iceberg.expressions.Literal`

huyuanfeng2018:
> > How about we generate Expression through org.apache.iceberg.expressions.Literal, which has a lot of conversion logic built in. WDYT?
> 
> Thanks for your review! There are some good comments, I will fix them soon. And can you explain more details about how to generate Expression through `org.apache.iceberg.expressions.Literal`

比如我想删除24小时之前的数据，并且对应字段是timestamp我可以这样

huyuanfeng2018:
> > How about we generate Expression through org.apache.iceberg.expressions.Literal, which has a lot of conversion logic built in. WDYT?
> 
> Thanks for your review! There are some good comments, I will fix them soon. And can you explain more details about how to generate Expression through `org.apache.iceberg.expressions.Literal`

If you need to process the timestamp type field, you can do this 

Expiring data older than a week

```
    //timestampz
    long expireTimestampz = (long) Literal.of(OffsetDateTime.now().minusHours(168).toString())
        .to(Types.TimestampType.withZone()).value();
    //timestamp
    long expireTimestamp = (long) Literal.of(OffsetDateTime.now().minusHours(168)
            .atZoneSameInstant(ZoneOffset.UTC).toLocalDateTime().toString())
        .to(Types.TimestampType.withoutZone()).value();
```

XBaith:
cc @zhoujinsong @shidayang

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
```diff
@@ -0,0 +1,468 @@
...
+  private static void expireFiles(UnkeyedTable table, long snapshotId, ExpireFiles expiredFiles, long expireTimestamp) {
+    List<DataFile> dataFiles = expiredFiles.dataFiles;
+    List<DeleteFile> deleteFiles = expiredFiles.deleteFiles;
+    if (dataFiles.isEmpty() && deleteFiles.isEmpty()) {
+      return;
+    }
+    // expire data files
+    DeleteFiles delete = table.newDelete();
+    dataFiles.forEach(delete::deleteFile);
+    delete.set(SnapshotSummary.SNAPSHOT_PRODUCER, "Amoro data expiration");
```
baiyangtx, line 264: "Amoro data expiration" defined in `SnapshotSummary` would be better ? It's better to define it as an enumeration or a single word, rather than a value with spaces.

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
```diff
@@ -0,0 +1,468 @@
...
+    delete.commit();
+    // expire delete files
+    if (!deleteFiles.isEmpty()) {
```
baiyangtx, line 267: Is it better to apply both operations, Delete datafile and DeleteFile rewrite, in one transaction?

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
```diff
@@ -0,0 +1,468 @@
...
+              if (e.isChange) {
+                changeExpiredFiles.addFile(e);
+              } else {
+                baseExpiredFiles.addFile(e);
+              }
+            });
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+      expireFiles(changeTable, changeSnapshot.snapshotId(), changeExpiredFiles, expireTimestamp);
```
baiyangtx, line 221: Is it necessary to set an expiry on the change store for Mixed format, or can it be resolved through the retention time of the change store itself? @zhoujinsong

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
XBaith: Do you mean replace thoses operation with RewriteFiles interface?

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
XBaith: Makes sense, but I'm more confused `CommitMetaProducer` already defines certain enums, should I add new enums to it? And how does this affect thrift connection? Or defining new operation constants in `SnapshotSummary`?

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
```diff
@@ -0,0 +1,468 @@
...
+    } else if (type.typeId().equals(Type.TypeID.STRING)) {
+      literal = Literal.of(
+          LocalDate.parse(upperBound.toString(), formatter)
+              .atStartOfDay()
+              .atZone(getDefaultZoneId(field)).toInstant().toEpochMilli());
+    }
+    return literal;
+  }
+
+  protected static class DataExpirationConfig {
```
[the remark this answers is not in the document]
XBaith: I think it would be better to split it because `ExpiringDataConfig` will be persisted in the database and `DataExpirationConfig` has multiple fields that are not easy to serialize.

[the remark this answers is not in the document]
shidayang: Why ExpiringDataConfig will   be persisted in the database

shidayang:
I think setting retentionTime will require a limitation, for example, if the partition is set according to day(), setting retentionTime to one hour will cause the data in the partition to be deleted before it is fully collected.

XBaith:
> I think setting retentionTime will require a limitation, for example, if the partition is set according to day(), setting retentionTime to one hour will cause the data in the partition to be deleted before it is fully collected.

I'm sure this situation is predictable for users. If user set retentionTime that less than partition unit, we suggest `FILE` expiration level is better chose. And it's not easy to handle for other non-time partitioned

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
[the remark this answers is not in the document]
XBaith:
```
void updateTableRuntime(@Param("runtime") TableRuntime runtime);
```
`TableConfiguration` will be serialized and persisted to sysdb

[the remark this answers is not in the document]
shidayang: ok

shidayang:
> > I think setting retentionTime will require a limitation, for example, if the partition is set according to day(), setting retentionTime to one hour will cause the data in the partition to be deleted before it is fully collected.
> 
> I'm sure this situation is predictable for users. If user set retentionTime that less than partition unit, we suggest `FILE` expiration level is better chose. And it's not easy to handle for other non-time partitioned

makes sense

ams/server/src/main/java/com/netease/arctic/server/table/ExpiringDataConfig.java
```diff
@@ -0,0 +1,108 @@
+package com.netease.arctic.server.table;
```
shidayang, line 1: Lisence

shidayang (changes requested):

shidayang (approved):

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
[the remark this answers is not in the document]
zhoujinsong: When I first saw it, I also wondered if it could be unioned with `ExpiringDataConfig`.
Can we mark those parts that are not suitable for persistence as transient and reinitialize them when needed?

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
```diff
@@ -0,0 +1,479 @@
...
+            .forEach(expiredFiles::addFile);
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+      expireFiles(unkeyedTable, snapshot.snapshotId(), expiredFiles, expireTimestamp);
+    }
+  }
+
+  private static Expression getPartitionExpression(DataExpirationConfig expirationConfig, long expireTimestamp) {
+    if (expirationConfig.expirationLevel.equals(DataExpirationConfig.ExpireLevel.PARTITION)) {
```
zhoujinsong, line 224: Can you add some comment here to tell why the partition level expiration do not need the partition filter?

zhoujinsong: BTW, It seems to be a data filter rather than partition filter.

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
```diff
@@ -0,0 +1,479 @@
...
+
+  protected static ZoneId getDefaultZoneId(Types.NestedField expireField) {
+    Type type = expireField.type();
+    if (type.typeId() == Type.TypeID.STRING) {
+      return ZoneId.systemDefault();
+    }
+    return ZoneOffset.UTC;
+  }
+
+  private static TableEntriesScan fileEntriesScan(UnkeyedTable table, Expression partitionFilter, Snapshot snapshot) {
```
zhoujinsong, line 153: I recommend not using `TableEntriesScan` anymore, as it has significant performance issues. Can we consider using `TableScan` directly to get all files, or try the `org.apache.iceberg.FindFiles` instead?

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
zhoujinsong: As far as I understand, ChangeStore already has a TTL mechanism based on commit time, but this data expiration feature is a more flexible expiration strategy based on business fields. Both can coexist.

zhoujinsong (commented):
@XBaith Thanks a lot for your work!
I left some comments, please task another look.

ams/server/src/main/java/com/netease/arctic/server/table/executor/DataExpiringExecutor.java
zhoujinsong: Maybe we can define a static variable in `SnapshotSummry` like `DATA_EXPIRATION` ?

XBaith: +1

PR #1905, opened 2023-08-31
[AMORO-1898] 0.4.x fix move files to stale hive location for Mixed Hive Table

### LINKED ISSUE #1898: [Bug]: If the optimizing of a Mixed Hive table fails to alter the Hive location, the files will be moved to the old Hive location.
### What happened?

When optimizing the Mixed Hive Format table but failing to alter the hive location,  the following full-optimizing will move files to the old Hive location. This brings two problems
- If the asynchronous thread successfully alters the Hive location, users will see a decrease in the data in Hive, similar to phantom reads.
- The commit of the following full-optimizing will fail, because the old hive location contains a mixture of files from new and old snapshots. The exception will be like

image

### Affects Versions

0.5.x,0.4.x,0.3.x

Core

### How to reproduce

This issue occurs only when failing to alter the Hive location.

To reproduce this scenario
1. We can deliberately change the hive location back to its original location after the full-optimizing committing. 
2. Then insert some data, and wait for the next full-optimizing move files to the old hive location.
3. Then trigger another full-optimizing, it will fail when committing.

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

wangtaohz:

Close #1898 .

## Brief change log

- move files to the location stored in iceberg properties instead of to the hive location
- support check and fix hive location before committing

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BasicOptimizeCommit.java
+++ b/ams/ams-server/src/main/java/com/netease/arctic/ams/server/optimize/BasicOptimizeCommit.java
@@ -26,6 +26,8 @@ import com.netease.arctic.ams.server.model.TableOptimizeRuntime;
 import com.netease.arctic.ams.server.utils.UnKeyedTableUtil;
 import com.netease.arctic.data.DataTreeNode;
 import com.netease.arctic.data.file.FileNameGenerator;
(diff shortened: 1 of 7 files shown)


wangtaohz:
a cherry-pick of #1911

PR #1909, opened 2023-09-03
[AMORO-1335]Amoro AMS Support On K8S

### LINKED ISSUE #1335: [Feature]: Support Kubernetes deployment for AMS
### Description

Support deploying AMS on Kubernetes.

### Use case/motivation

Kubernetes is one of the most popular and widely used container orchestration tools, supporting Kubernetes deployment can facilitate the use of AMS in cloud-native environments.

### Describe the solution

_No response_

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

czy006:

AMS Run On K8S

Close #1335.

## Brief change log

- add helm chart for amoro
- ams running on k8s

## How was this patch tested?

- [x] ams log4j2 and config.xml can be rewrite
- [x] ams server and web ui on k8s open normal
- [x] Run test locally on k8s before making a pull request

## Documentation

### DIFF
--- a/charts/amoro/templates/amoro-configmap.yaml
+++ b/charts/amoro/templates/amoro-configmap.yaml
@@ -1,0 +1,132 @@
+{{/*
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/}}
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}
+  labels:
+    {{- include "amoro.labels" . | nindent 4 }}
+data:
+  ## Helm chart provided Amoro configurations
+  config.yaml: |
+    ams:
+      admin-username: admin
+      admin-password: admin
+      server-bind-host: "0.0.0.0"
+      server-expose-host: "127.0.0.1"
+
+      thrift-server:
+        max-message-size: 104857600 # 100MB
+        selector-thread-count: 2
+        selector-queue-size: 4
+        table-service:
+          bind-port: {{ .Values.server.table.port }}
+          worker-thread-count: 20
+        optimizing-service:
+          bind-port: {{ .Values.server.optimizing.port }}
+
+      http-server:
+        bind-port: {{ .Values.server.rest.port }}
+
+      refresh-external-catalogs:
+        interval: 180000 # 3min
+
+      refresh-tables:
+        thread-count: 10
+        interval: 60000 # 1min
+
+      self-optimizing:
+        commit-thread-count: 10
+
+      optimizer:
+        heart-beat-timeout: 60000 # 1min
+        task-ack-timeout: 30000 # 30s
+
+      blocker:
+        timeout: 60000 # 1min
+
+      # optional features
+      expire-snapshots:
+        enabled: true
+        thread-count: 10
+
+      clean-orphan-files:
+        enabled: true
+        thread-count: 10
+
+      sync-hive-tables:
+        enabled: true
+        thread-count: 10
+
+      database:
+        type: {{ .Values.amoroConf.database.type }}
+        url: {{ .Values.amoroConf.database.url }}
+        jdbc-driver-class: {{ .Values.amoroConf.database.driver | quote }}
+        {{- if eq .Values.amoroConf.database.type "mysql" }}
+        username: {{ .Values.amoroConf.database.username | quote }}
+        password: {{ .Values.amoroConf.database.password | quote }}
+        {{- end }}
+
+      #  MySQL database configuration.
+      #  database:
+      #    type: mysql
+      #    jdbc-driver-class: com.mysql.cj.jdbc.Driver
+      #    url: jdbc:mysql://127.0.0.1:3306/db?useUnicode=true&characterEncoding=UTF8&autoReconnect=true&useAffectedRows=true&useSSL=false
+      #    username: root
+      #    password: root
+
+      terminal:
+        backend: local
+        local.spark.sql.iceberg.handle-timestamp-without-timezone: false
+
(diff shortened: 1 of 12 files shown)


majin1102:
Thank you for your contribution. Your PR is very valuable to Amoro community.

Review request @baiyangtx bai @HuangFru

baiyangtx:
Thanks for your contribution, Your PR is very valuable to me. I will try this as soon as possible.

charts/amoro/values.yaml
```diff
@@ -0,0 +1,163 @@
...
+    enabled: true
+    port: 1261
+    service:
+      type: ClusterIP
+      port: "{{ .Values.server.optimizing.port }}"
+      nodePort: ~
+      annotations: { }
+
+# $AMS_CONF_DIR directory
+amoroConfDir: /usr/local/ams/amoro-0.5.0/conf
```
HuangFru, line 78: Same thing here as I comment in 'Chart.yaml', set a value to present the Amoro version and use the value instead of the number, like:
'amoroConfDir: /usr/local/ams/amoro-{{ .Values.appVersion }}/conf'.
But I'm not sure if this works fine. Appreciate it if you could verify it. This will save some upgrade work and avoid oversights.

charts/amoro/templates/amoro-configmap.yaml
```diff
@@ -0,0 +1,128 @@
...
+      database:
+        type: derby
+        jdbc-driver-class: org.apache.derby.jdbc.EmbeddedDriver
+        url: jdbc:derby:/tmp/amoro/derby;create=true
+
+      #  MySQL database configuration.
+      #  database:
+      #    type: mysql
+      #    jdbc-driver-class: com.mysql.cj.jdbc.Driver
+      #    url: jdbc:mysql://127.0.0.1:3306/db?useUnicode=true&characterEncoding=UTF8&autoReconnect=true&useAffectedRows=true&useSSL=false
```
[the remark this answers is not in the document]
XBaith: In product environments, it is usually not possible to display accounts/passwords directly due to security issues, can this be put in sercert?

charts/amoro/templates/amoro-configmap.yaml
```diff
@@ -0,0 +1,128 @@
...
+  labels:
+    {{- include "amoro.labels" . | nindent 4 }}
+data:
+  ## Helm chart provided Amoro configurations
+  config.yaml: |
+    ams:
+      admin-username: admin
+      admin-password: admin
+      server-bind-host: "0.0.0.0"
+      server-expose-host: "127.0.0.1"
```
baiyangtx, line 31: Using 127.0.0.1 will cause the RestCatalog URI generated by AMS to be accessible only from within the container. It may be necessary to expose the access address of AMS. However, this is not well adapted to non-IP address situations.

charts/amoro/templates/amoro-configmap.yaml
```diff
@@ -0,0 +1,128 @@
...
+    #  terminal:
+    #    backend: kyuubi
+    #    kyuubi.jdbc.url: jdbc:hive2://127.0.0.1:10009/
+
+
+    #  High availability configuration.
+    #  ha:
+    #    enabled: true
+    #    cluster-name: default
+    #    zookeeper-address: 127.0.0.1:2181,127.0.0.1:2182,127.0.0.1:2183
```
baiyangtx, line 103: High availability is recommended in production environments, so it is recommended to open this part of the configuration to users.

charts/amoro/templates/amoro-configmap.yaml
```diff
@@ -0,0 +1,128 @@
...
+        thread-count: 10
+
+      sync-hive-tables:
+        enabled: true
+        thread-count: 10
+
+      database:
+        type: derby
+        jdbc-driver-class: org.apache.derby.jdbc.EmbeddedDriver
+        url: jdbc:derby:/tmp/amoro/derby;create=true
```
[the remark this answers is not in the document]
czy006: sure,Can I assume here that derby database is only for testing and not suitable for the tips that most users use in production environments ?

charts/amoro/templates/amoro-configmap.yaml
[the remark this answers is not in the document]
czy006: When derby is currently used by default, should this part of our configuration be exposed to value for the user to configure cc @Kyofin 

Yeah, it's safer if you put it in secert, I'll change that later cc @XBaith

charts/amoro/templates/amoro-configmap.yaml
czy006: Can I assume this is follow up done in other issuse? , currently we can use the **port-forward** command to access

baiyangtx: OK, I will follow this.

charts/amoro/templates/amoro-configmap.yaml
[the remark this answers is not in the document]
czy006: After offline discussion, this part is out of this [issuse #1945](https://github.com/NetEase/amoro/issues/1945) access, I have opened new issuse, this feature will be done there

cc @Kyofin @XBaith

charts/amoro/values.yaml
```diff
@@ -0,0 +1,162 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+# AMS server numbers
+replicaCount: 1
```
baiyangtx, line 20: think it's more appropriate to specify this configuration in the HA section.

baiyangtx (commented):
I tested it on Minikube. It's easy to install the AMS service using Helm. However, I don't think we should currently support multi-replica deployment for the following reasons:

1. Amoro is not a stateless service and caches a lot of state in memory. The current HA design is an active/standby architecture, and when the standby node becomes the master, the memory state needs to be reinitialized.

2. In the current design, the standby node has no accessible ports for Thrift or REST, so the pod liveness probe will have problems.

I suggest that we only support single-replica deployment in the first Helm version and support multi-replica deployment after resolving the pod liveness probe problems.

charts/amoro/values.yaml
czy006: The first initial release did not support multi-copy deployment, I suggest that this part currently needs to mask the HA multi-copy configuration, which would be outside the scope of this issue if code support was required.I will temporarily block this option and default to 1 cc @baiyangtx

baiyangtx (approved):
LGTM, Thanks a lot for your contribution!

HuangFru:
I've tried this in Rancher Desktop with K8s on my local machine, works fine.
LGTM. Thanks for your contribution!

HuangFru (approved):

XBaith (approved):

PR #1911, opened 2023-09-04
[AMORO-1898] fix Mixed Hive Table's self-optimizing moving files to the stale hive table/partition location

### LINKED ISSUE #1898: [Bug]: If the optimizing of a Mixed Hive table fails to alter the Hive location, the files will be moved to the old Hive location.

wangtaohz:

Close #1898.

## Brief change log

- move files to the location stored in iceberg properties instead of to the hive location
- support check and fix hive location before committing

## How was this patch tested?

## Documentation


hive/src/main/java/com/netease/arctic/hive/utils/HiveMetaSynchronizer.java
```diff
@@ -55,13 +58,16 @@
...
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
- * Synchronize the metadata of the hive table to arctic table
+ * Utils for sync the metadata(schema/data) of the hive table to arctic table or sync data of arctic table to hive table
```
zhoujinsong, line 70:
```suggestion
 * Utils for syncing the metadata between the hive table and the arctic table.
```

ams/server/src/main/java/com/netease/arctic/server/optimizing/UnKeyedTableCommit.java
```diff
@@ -105,41 +109,60 @@ protected List<DataFile> moveFile2HiveIfNeed() {
...
+      String hiveSubdirectory = table.isKeyedTable() ?
+          HiveTableUtil.newHiveSubdirectory(maxTransactionId) : HiveTableUtil.newHiveSubdirectory();
+
+      Partition p = HivePartitionUtil.getPartition(hiveClient, table, partitionValues);
+      if (p == null) {
+        return HiveTableUtil.newHiveDataLocation(((SupportHive) table).hiveLocation(),
+            table.spec(), targetFile.partition(), hiveSubdirectory);
+      } else {
+        return p.getSd().getLocation();
+      }
```
baiyangtx, line 150: Does it seem like there might be a performance issue if HMSClient.getPartition() needs to be called for each DataFile?

hive/src/main/java/com/netease/arctic/hive/utils/HiveMetaSynchronizer.java
```diff
@@ -222,6 +228,207 @@ public static void syncHiveDataToArctic(SupportHive table, HMSClientPool hiveCli
     }
   }
 
+  /**
+   * Synchronize the data change of the arctic table to hive table
+   * @param table support hive table
+   */
+  public static void syncArcticDataToHive(SupportHive table) {
+    UnkeyedTable baseStore;
+    if (table.isKeyedTable()) {
```
baiyangtx, line 237: Should this method be in HiveDataSynchronizer ?

Is there some unit test for this method?

ams/server/src/main/java/com/netease/arctic/server/optimizing/UnKeyedTableCommit.java
wangtaohz: The `computeIfAbsent` is trying to avoid the performance issue.
```
String partitionPath = partitionPathMap.computeIfAbsent(taskRuntime.getPartition(),
            key -> getPartitionPath(hiveClient, maxTransactionId, targetFile, partitionSchema));
```

hive/src/main/java/com/netease/arctic/hive/utils/HiveMetaSynchronizer.java
wangtaohz: I don't think we have a class named `HiveDataSynchronizer` now, and the `HiveMetaSynchronizer` contains both methods `syncHiveSchemaToArctic` and `syncHiveDataToArctic`, which means `Schema` and `Data` are both kinds of `Meta`.

The unit test class is in the module `ams/server` which is `TestHiveCommitSync`, I will move it to module `hive`.

baiyangtx: Is the integration test under the ams/server module? Should additional unit tests be added specifically for this method?

wangtaohz: Not the integration test, it‘s a specialized test class for syncing Amoro data to Hive.

baiyangtx (approved):
LGTM, Thanks a lot for your contribution.

PR #1284, opened 2023-03-27
[ARCTIC-1242][ams] Optimizing plan thread should only plan tables in pending status

wangtaohz:
## Why are the changes needed?

fix, #1242 

## Brief change log

  - optimizing-check thread pool skips `pending` tables and only changes optimizing status from `idle` to `pending`
  - optimizing-planner thread skips `idle` tables and only plans tables in `pending` 

## How was this patch tested?
- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request

## Documentation

  - Does this pull request introduces a new feature? (yes / **no**)
  - If yes, how is the feature documented? (not applicable / docs / JavaDocs / **not documented**)

PR #1928, opened 2023-09-07
[AMORO-1242] Fix refreshing a large number of tables during optimizing plan

### LINKED ISSUE #1242: [Improvement]: Improve resource utilization of parallel Optimizer
### Search before asking

The Optimizer resource utilization is very lower(~10%) when I have an Optimizing Group with 100+ iceberg format tables and 20 parallelisms, and many tables are `Pending`.

The bottleneck may be the planning process since there is only one thread planning for all the tables.

The plan cost was as follows:

image

### How should we improve?

We should improve the plan performance or plan in parallel.

### Subtasks

_No response_

### Code of Conduct

wangtaohz:

Close #1242 . 

 This PR (#1284) has achieved only planning the pending tables to improve plan efficiency, but it still refreshes the idle tables. When there are many tables, this can become a bottleneck.

In the production scenario, we observed more than 1,00 tables were refreshed for each plan, which cost 10-20s. 
However, plan for each table only cost 1-2s.

## Brief change log

- check pending status before refresh table
- modify some logs for plan cost of each table

## How was this patch tested?

## Documentation


nicochen (approved):

shendanfengg (commented):
LGTM.

PR #1936, opened 2023-09-07
[AMORO-1935] 0.4.x introduce a table property for task spliting size of self-optimizing

### LINKED ISSUE #1935: [Improvement]: Introduce a table property to set the task splitting size of self-optimizing
### Search before asking

Currently, the size of the optimizing task is implicitly determined by the table property  `self-optimizing.target-size`, which is not very clear. Also, the size of the task does not need to be equal to the target size of the file.

A reasonable task size should satisfy two conditions: 1) it should not be smaller than the file target size, and 2) it should not cause the task execution time to be too long (>5~10 minutes).

### How should we improve?

Therefore, we should introduce a table property to determine the size of task splitting of the table. 1GB maybe a reasonable default value.

### Subtasks

_No response_

### Code of Conduct

wangtaohz:

Close #1935 .

## Brief change log

- add table property self-optimizing.max-task-file-size-bytes
- add to docs

## How was this patch tested?

## Documentation


wangtaohz:
t should also be supported in the master.

PR #1969, opened 2023-09-13
[AMORO-1968] Enable using JAVA8 API to handle timestamp in the Terminal

### LINKED ISSUE #1968: [Bug]: Terminal has time zone problem with timestamp
### What happened?

When setting the time zone in the Terminal with `spark.sql.session.timeZone` like this :
image

And then execute the 'select' statement to read from a table, The config did not take effect.

Spark does not support timestamps without the time zone. So when reading a timestamp without a time zone type written by Flink, Spark will add a local time zone and will cause a time difference. So we need to set `spark.sql.session.timeZone` to 'UTC' to avoid this situation. I have already verified this in the Spark environment.

However, this config does not take effect in the AMS terminal when reading data. According to https://stackoverflow.com/questions/49644232/how-to-set-timezone-to-utc-in-apache-spark, 
There may be implicit conversions in the process of executing Java code and returning results in the terminal.

### Affects Versions

master

AMS

### How to reproduce

Create a mix-iceberg table in the Flink with a time stamp filed(without time zone).

Write some data.

Read in the Terminal.

### Relevant log output

```shell
Write:
1	aaa	2015-10-11 00:00:00.0

Read in terminal:
1	aaa	2015-10-11 08:00:00.0
```

### Anything else

_No response_

### Code of Conduct

HuangFru:

Close #1968.

## Brief change log
- Set default time zone in Java when the Spark time zone is set

## How was this patch tested?

## Documentation


baiyangtx:
Is it necessary to set Timezone for this?

HuangFru:
> Is it necessary to set Timezone for this?

If not, the timestamp will still get UTC+8.

baiyangtx:
Cloud you please add some screenshot ?

Screenshot before your fix, and config.yml

Screenshot after your fix, and config.yml

HuangFru:
> Cloud you please add some screenshot ?
> 
> Screenshot before your fix, and config.yml
> 
> Screenshot after your fix, and config.yml

BEFORE:
image

image

AFTER:
image

image

HuangFru:
According to an article from the databricks : https://www.databricks.com/blog/2020/07/22/a-comprehensive-look-at-dates-and-timestamps-in-apache-spark-3-0.html
I find a better solution and works fine:
image

PR #1978, opened 2023-09-15
[AMORO-1945]AMS Support Database Secret On K8S

### LINKED ISSUE #1945: [Feature]: AMS Support Database Secert On K8S
### Description

Currently, the database password is written to the configuration file in plain text, which may pose certain security risks. In the cloud-native environment, this configuration will be saved through K8S Secert, which will improve password security.

### Use case/motivation

After discussing with [pull request](https://github.com/NetEase/amoro/pull/1909), we decided to separate and complete this part of the function, which will improve the security of database passwords

### Describe the solution

The Secert mount function only supports running on K8S

- Mount the Secert or get the password by reading the environment variable, and decode it through base64
- The environment variable name is: DATABASE_PASSWORD
- Read Priority: Environment Variables > Configuration Files
  - If the deployment environment is K8S, the variable will be read from the environment variable and the base64 deserialized password will be performed. The user only needs to set the password in the configuration file, and it will be automatically set to Env in Secert mode, which is read by the program, and there is no need to care about this part of the implementation.
  - If the deployment environment is a non-K8S environment, first read from the environment variable, if it is empty, then read directly from the configuration file, this part will not have any encryption behavior

### Subtasks

_No response_

### Related issues

None

### Code of Conduct

czy006:

AMS Helm Chart Support Database Secret On K8S
When the user fills in the mysql password in the helm template, use secret to protect it.

Close #1945 .

## Brief change log

- Add System Env Method Filter start with "ams.*" properties
- Database MySQL Type use secret to write env and read it

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
@@ -20,6 +20,7 @@ package com.netease.arctic.server;
 
 import com.alibaba.fastjson.JSONArray;
 import com.alibaba.fastjson.JSONObject;
+import com.google.common.collect.Maps;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.netease.arctic.ams.api.ArcticTableMetastore;
 import com.netease.arctic.ams.api.Constants;
@@ -41,13 +42,13 @@ import com.netease.arctic.server.table.TableService;
 import com.netease.arctic.server.table.executor.AsyncTableExecutors;
 import com.netease.arctic.server.terminal.TerminalManager;
 import com.netease.arctic.server.utils.ConfigOption;
+import com.netease.arctic.server.utils.ConfigurationUtil;
 import com.netease.arctic.server.utils.Configurations;
 import com.netease.arctic.server.utils.ThriftServiceProxy;
 import io.javalin.Javalin;
 import io.javalin.http.HttpCode;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
-import org.apache.iceberg.relocated.com.google.common.collect.Maps;
 import org.apache.thrift.TMultiplexedProcessor;
 import org.apache.thrift.TProcessor;
 import org.apache.thrift.protocol.TBinaryProtocol;
@@ -301,12 +302,13 @@ public class ArcticServiceContainer {
     private JSONObject yamlConfig;
 
     public void init() throws IOException {
-      initServiceConfig();
+      Map<String, Object> envConfig = initEnvConfig();
+      initServiceConfig(envConfig);
       initContainerConfig();
     }
 
     @SuppressWarnings("unchecked")
-    private void initServiceConfig() throws IOException {
+    private void initServiceConfig(Map<String, Object> envConfig) throws IOException {
       LOG.info("initializing service configuration...");
       String configPath = Environments.getConfigPath() + "/" + SERVER_CONFIG_FILENAME;
       LOG.info("load config from path: {}", configPath);
@@ -314,9 +316,17 @@ public class ArcticServiceContainer {
       JSONObject systemConfig = yamlConfig.getJSONObject(ArcticManagementConf.SYSTEM_CONFIG);
       Map<String, Object> expandedConfigurationMap = Maps.newHashMap();
       expandConfigMap(systemConfig, "", expandedConfigurationMap);
+      // If same configurations in files and environment variables, environment variables have higher priority.
+      expandedConfigurationMap.putAll(envConfig);
       validateConfig(expandedConfigurationMap);
       serviceConfig = Configurations.fromObjectMap(expandedConfigurationMap);
       SqlSessionFactoryProvider.getInstance().init(serviceConfig);
+    }
+
+    private Map<String, Object> initEnvConfig() {
+      LOG.info("initializing system env configuration...");
+      String prefix = ArcticManagementConf.SYSTEM_CONFIG.toUpperCase();
+      return ConfigurationUtil.convertConfigurationKeys(prefix, System.getenv());
     }
 
     private void validateConfig(Map<String, Object> systemConfig) {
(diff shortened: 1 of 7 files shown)


czy006:
The password has been injected into the environment variable and the read configuration is correct

image

ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
```diff
@@ -192,7 +192,7 @@ private void startThriftServer(TServer server, String threadName) {
 
   private void initHttpService() {
     DashboardServer dashboardServer = new DashboardServer(
-        serviceConfig, tableService, optimizingService, terminalManager);
+            serviceConfig, tableService, optimizingService, terminalManager);
```
baiyangtx, line 195: Please use 2 spaces for indentation.

ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
```diff
@@ -301,57 +302,72 @@ private class ConfigurationHelper {
...
+      validateConfig(envConfig);
+      serviceConfig = Configurations.fromObjectMap(envConfig);
       SqlSessionFactoryProvider.getInstance().init(serviceConfig);
     }
 
+    private Map<String, Object> initEnvConfig() {
+      LOG.info("initializing system env configuration...");
+      Map<String, String> envConfigMap = System.getenv().entrySet()
+              .stream()
+              .filter(entry -> entry.getKey().startsWith(ArcticManagementConf.SYSTEM_CONFIG))
```
baiyangtx, line 327: The value of `SYSTEM_CONFIG` is `ams`. Should we use a new env such as `AMORO_CONF_` ?

czy006:
I have checked the unit tests and have not found a suitable way to write them in (AmsEnvironment.java). If there is any way to intervene, please tell me.

baiyangtx (commented):
Please use 2 spaces for indentation.

ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
```diff
@@ -301,57 +302,72 @@ private class ConfigurationHelper {
     private JSONObject yamlConfig;
 
     public void init() throws IOException {
-      initServiceConfig();
+      Map<String, Object> envConfig = initEnvConfig();
+      initServiceConfig(envConfig);
       initContainerConfig();
```
baiyangtx, line 307: What's the order of ENV config and configs in file? The file config has higher priority ?

czy006: env variable is larger than the configuration file.When the env is configured, the file configuration properties does not take effect.If configured, it will be overwritten by environment variables. what do you think ?

ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
czy006: Does this mean that we need to extract all environment variables starting with AMORO_CONF_ ？

ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
baiyangtx: Have you tested the code? It looks like properties in config file will overwrite the envProperties.

ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
baiyangtx: Referring to the environment variable configuration methods of other open source software, such as 1, 2, and 3.

Considering the common practice of using uppercase letters and underscores in environment variables, I suggest:

    Using "AMORO_" as the prefix for environment variable configuration.
    Using underscores instead of dots.
    Keeping the configuration consistent with the config for ease of configuration.

For example, use `AMORO_http-server_bind-port` to represent the configuration item `ams.http-server.bind-port`.

@zhoujinsong  WDYT ?

baiyangtx: Here are some references:

- https://etcd.io/docs/v3.4/op-guide/configuration/
- https://github.com/apache/spark/blob/master/conf/spark-env.sh.template
- https://github.com/Gradiant/dockerized-hadoop/blob/master/README.md

zhoujinsong:
@czy006 Thanks a lot for your contribution!

As @baiyangtx said, Amoro project use 2 spaces as indent and tab size, you can configure it in IDEA and reformat the code.

I agree that environment variables should have a higher priority than configuration files, just like other projects such as Apache Hadoop.

Although we currently only want to pass the database password through environment variables, it is necessary to standardize the naming conventions of environment variables in advance so that more environment variables can be added in the future.
In my option, the names of environment variables should be consistent with the name in the configuration file for easy management, but we should use capital letters to store environment variables, which is in line with the common usage of environment variables. So `AMS_DATABASE_PASSWORD` may be a good option to store the `ams.database.password` configuration with the environment variables way.

@czy006 @baiyangtx HDYT?

czy006:
> @czy006 Thanks a lot for your contribution!
> 
> As @baiyangtx said, Amoro project use 2 spaces as indent and tab size, you can configure it in IDEA and reformat the code.
> 
> I agree that environment variables should have a higher priority than configuration files, just like other projects such as Apache Hadoop.
> 
> Although we currently only want to pass the database password through environment variables, it is necessary to standardize the naming conventions of environment variables in advance so that more environment variables can be added in the future. In my option, the names of environment variables should be consistent with the name in the configuration file for easy management, but we should use capital letters to store environment variables, which is in line with the common usage of environment variables. So `AMS_DATABASE_PASSWORD` may be a good option to store the `ams.database.password` configuration with the environment variables way.
> 
> @czy006 @baiyangtx HDYT?

I agree. If so, I will convert them to uppercase values ​​and use environment variables as the highest priority when detecting values ​​prefixed by ams.

Waiting for other Contributor opinions ～

I’ll deal with the formatting issues later.

czy006:
@zhoujinsong @baiyangtx PLAT

baiyangtx (commented):
The code style is fine. There is only one things. 

What the ENV prefix should be?  `AMORO_`  or `ams.*`. 

@zhoujinsong PTAL

baiyangtx:
图片


It has some problem when setting environment variables with '-', maybe we should replace with '__'

baiyangtx (approved):
LGTM, thanks for you contributions.

PR #1988, opened 2023-09-18
[AMORO-1935] Introduce a table property for task spliting size of self-optimizing

### LINKED ISSUE #1935: [Improvement]: Introduce a table property to set the task splitting size of self-optimizing

wangtaohz:

Close #1935.

## Brief change log

- add a new table property `self-optimizing.max-task-size-bytes`
- modify the docs

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingConfig.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingConfig.java
@@ -252,8 +252,8 @@ public class OptimizingConfig {
   public int hashCode() {
     return Objects.hashCode(enabled, targetQuota, optimizerGroup, maxExecuteRetryCount, maxCommitRetryCount, targetSize,
         maxTaskSize, maxFileCount, openFileCost, fragmentRatio, minorLeastFileCount, minorLeastInterval,
-        majorDuplicateRatio,
-        fullTriggerInterval, fullRewriteAllFiles, baseHashBucket, baseRefreshInterval, hiveRefreshInterval);
+        majorDuplicateRatio, fullTriggerInterval, fullRewriteAllFiles, baseHashBucket, baseRefreshInterval,
+        hiveRefreshInterval);
   }
 
   @Override
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingConfig.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingConfig.java
@@ -31,6 +31,9 @@ public class OptimizingConfig {
   //self-optimizing.target-size
   private long targetSize;
 
+  //self-optimizing.max-task-size-bytes
+  private long maxTaskSize;
+
   //self-optimizing.max-file-count
   private int maxFileCount;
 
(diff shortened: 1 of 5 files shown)


docs/user-guides/configurations.md
```diff
@@ -32,14 +32,16 @@ Self-optimizing configurations are applicable to both Iceberg Format and Mixed s
 | self-optimizing.enabled                             | true             | Enables Self-optimizing                                |
 | self-optimizing.group                               | default          | Optimizer group for Self-optimizing                                   |
 | self-optimizing.quota                               | 0.1              | Quota for Self-optimizing, indicating the CPU resource the table can take up                       |
-| self-optimizing.execute.num-retries                         | 5                | Number of retries after failure of Self-optimizing                       |
+| self-optimizing.execute.num-retries                 | 5                | Number of retries after failure of Self-optimizing                       |
 | self-optimizing.target-size                         | 134217728(128MB) | Target size for Self-optimizing                           |
 | self-optimizing.max-file-count                      | 10000            | Maximum number of files processed by a Self-optimizing process              |               |
```
zhoujinsong, line 37: It seems `self-optimizing.max-file-count ` is no longer used, shoulde we remove it from docs too?

wangtaohz: This parameter still works when performing a keyed table scan and is used to limit the number of cached change store files to avoid OOM of AMS.

I haven't found a better way to replace this parameter yet.

zhoujinsong: I found that `com.netease.arctic.server.optimizing.OptimizingConfig#getMaxFileCount` has never been called, so I mistakenly thought that this parameter was useless.

We can discuss in other issues/PRs about:

- whether to standardize the use of parameters in AMS
- whether to unify the configuration parameters of different formats.

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingConfig.java
```diff
@@ -239,7 +251,8 @@ public boolean equals(Object o) {
   @Override
   public int hashCode() {
     return Objects.hashCode(enabled, targetQuota, optimizerGroup, maxExecuteRetryCount, maxCommitRetryCount, targetSize,
-        maxFileCount, openFileCost, fragmentRatio, minorLeastFileCount, minorLeastInterval, majorDuplicateRatio,
+        maxTaskSize, maxFileCount, openFileCost, fragmentRatio, minorLeastFileCount, minorLeastInterval,
+        majorDuplicateRatio,
```
zhoujinsong, line 255: Leaving the single paramter in one line here makes the format quite strange.

wangtaohz: OK, I will modify it. Code refactoring brought about this strange state.

zhoujinsong (approved):
LGTM.

PR #1990, opened 2023-09-18
[AMORO-1857]: Check short name for principal before call constructUgi()

### LINKED ISSUE #1857: [Bug]:When using an HDFS cluster protected by Kerberos, the fs.hdfs.impl.disable.cache=false configuration is invalid, FileSystem is unable to hit the cache
### What happened?

Kerberos accounts usually consist of the principal and realm parts, in the form of principal@Realm. When using ams to create a catalog, since the uploaded krb5.conf file already contains the realm information, the account can be abbreviated to just fill in the principal. If the kerberos account information filled in during catalog creation is abbreviated, it will cause the FileSystem to be unable to hit the cache, constantly create new FileSystem instances without being garbage collected, eventually leading to an out of memory error.
It is recommended to modify the validation logic on line 239 of com.netease.arctic.table.TableMetaStore from !ugi.getUserName().equals(krbPrincipal) to !ugi.getUserName().startsWith(krbPrincipal) to adapt to the abbreviated kerberos account pattern.

### Affects Versions

0.5.0

_No response_

### How to reproduce

When creating a catalog with Kerberos authentication, only the principal part is filled in the Kerberos principal field, without the realm information.

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

861752346:
 

Close #1857.

## Brief change log
 
- Check short name for principal before call `constructUgi()`

## How was this patch tested?

## Documentation


baiyangtx:
There is a checkstyle problem. could you please fix it and run `mvn validate` before you push the code?

baiyangtx (approved):
LGTM, thanks for your contribution.

PR #2005, opened 2023-09-20
[AMORO-1977] Fix retry logic in ThriftClientPool

### LINKED ISSUE #1977: [Bug] Trino failed to query Amoro Mixed Format Table after restarting AMS
### What happened?

重启amoro ams后，发现trino无法查询amoro表了

### Affects Versions

0.5.0

_No response_

### How to reproduce

重启amoro ams，执行trino查询amoro表

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

HuangFru:

Close #1977.

## Brief change log

- Invalidate the object when the client is disconnected.
- Cancel the second time to get the client.

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/api/src/main/java/com/netease/arctic/ams/api/client/AmsClientPools.java
+++ b/ams/api/src/main/java/com/netease/arctic/ams/api/client/AmsClientPools.java
@@ -14,8 +14,9 @@ import org.apache.thrift.protocol.TProtocol;
  */
 public class AmsClientPools {
 
-  private static final int CLIENT_POOL_MIN = 0;
-  private static final int CLIENT_POOL_MAX = 5;
+  private static final int CLIENT_POOL_MIN_IDLE = 0;
+  private static final int CLIENT_POOL_MAX_IDLE = 5;
+  private static final int CLIENT_POOL_MAX_WAIT_MS = 5000;
 
   private static final LoadingCache<String, ThriftClientPool<ArcticTableMetastore.Client>> CLIENT_POOLS
       = Caffeine.newBuilder()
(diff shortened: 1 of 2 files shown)


shidayang (approved):

hameizi (approved):

zhoujinsong:
@HuangFru Can you add some description about the cause of the bug, it may help us understanding your PR.

HuangFru:
> @HuangFru Can you add some description about the cause of the bug？ It may help us understanding your PR.



> @HuangFru Can you add some description about the cause of the bug？ It may help us understanding your PR.

According to the original code logic, when a client object is borrowed from the object pool, if the AMS is not available at that time (possibly due to restart or service interruption), it will enter the retry logic. In the retry logic, only `pool.clear()` is used to clear the objects in the pool with the "Idle" status, and then another client object is borrowed from the pool. 

There is no logic to return the originally borrowed object(the first borrowed object), which causes that object to remain in the pool indefinitely. Furthermore, the additional object borrowed in the retry logic is unnecessary, and it is not returned to the pool either. If there are ongoing queries in the engine and the AMS undergoes a restart or similar operation, it could trigger multiple times of this "borrow without return" operation, ultimately resulting in the object pool being filled up with no available idle objects, causing the queries to be blocked indefinitely (although there is a parameter to set a maximum waiting time, which is not used in the code).

zhoujinsong:
@HuangFru Thanks for your detail reply. Based on it, I can understand the implementation of your PR more easily.

Do you think we should complete the setting of the maximum waiting time?

HuangFru:
> @HuangFru Thanks for your detail reply. Based on it, I can understand the implementation of your PR more easily.
> 
> Do you think we should complete the setting of the maximum waiting time?

I think it is necessary. After all, throwing a timeout exception is more friendly and easier to troubleshoot than blocking.
How about 5 seconds?

zhoujinsong:
Making the default timeout configuration to 5 seconds looks good to me.

HuangFru:
> Making the default timeout configuration to 5 seconds for the `TableMannager` service looks good to me. But the Optimizer client may need a longer timeout configuration for now as it may need to wait the planning process in AMS.

For https://github.com/NetEase/amoro/pull/1712 changed the max total size of optimizing client pool, the optimizer should always have a client object to use. I think the optimizing client pool does not need to set the max timeout.

wangtaohz (approved):
LGTM

PR #2016, opened 2023-09-21
[AMORO-2006] Fix value logic in startup shell

### LINKED ISSUE #2006: [Bug]: The startup script has a judgment error
### What happened?

When I first started AMS, it worked fine:
POPO20230920-170105
But when I accidentally started it again:
POPO20230920-170231
There was a wrong judgment here. It deleted the pid file and tried to restart AMS but failed.
So I can no longer stop the AMS service via shell because it will always get the wrong PID:
POPO20230920-170259
I have to use some other method to find the process ID of the AMS that is still running and kill it manually.

### Affects Versions

master

AMS

### How to reproduce

./bin/ams.sh start
./bin/ams.sh start
./bin/ams.sh stop

### Relevant log output

```shell
See "What happened?"
```

### Anything else

_No response_

### Code of Conduct

HuangFru:

Close #2006.

## Brief change log

- The return value of the previous command is only evaluated once

## How was this patch tested?

## Documentation

- If yes, how is the feature documented? (not applicabled)

### DIFF
--- a/dist/src/main/arctic-bin/bin/ams.sh
+++ b/dist/src/main/arctic-bin/bin/ams.sh
@@ -175,6 +175,7 @@ case "$1" in
     start)
         status;
         status_return=$?
+
         if [ $status_return -eq 2 ]; then
             echo 'starting app server.'
             start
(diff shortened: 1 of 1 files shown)


wangtaohz (approved):
LGTM.

I test locally, and it works well.

PR #2063, opened 2023-10-09
[Improvement]: Change socpe of `amoro:spark`  to runtime in `ams/server` dependency.

baiyangtx:

The future mixed-format related features should be plug-in based. In order to better organize the project structure, the AMS module should not directly depend on the code of MixedFormat. This PR will remove the hard-coded access to the spark module in the AMS module.

## Brief change log

- Change socpe of `amoro:spark` to runtime in `ams/server` dependencies part.
- Using static final string instead of `Class.getName` 

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

PR #2105, opened 2023-10-16
[Polish] Code polish

GOODBOY008:
Polish code


GOODBOY008:
@shidayang PTAL.

shidayang:
> @shidayang PTAL.

Thank you for your contribution. There are some formatting issues in the code that caused the CI to fail. Could you please fix them?

ams/server/src/main/java/com/netease/arctic/server/catalog/InternalCatalog.java
```diff
@@ -135,21 +135,17 @@ public boolean exist(String database, String tableName) {
   }
 
   private String getDatabaseDesc(String database) {
-    return new StringBuilder()
-        .append(name())
-        .append('.')
-        .append(database)
-        .toString();
+    return name() +
```
zhoujinsong, line 138: The project use 2 spaces as indent and tab size, you may reconfigure it an reformat the codes here.

zhoujinsong (commented):
Thanks a lot for your contribution!

I noticed that you used the wrong indent configuration to format the code. You may need to reconfigure and reformat it.

baiyangtx:
What is the standard for this code specification? Are there any relevant checking tools to detect whether the code meets the standard?

Is this standardization of the code a one-time event or can it be continuously enforced? It seems like a one-time optimization, and there is no guarantee that subsequent submissions will meet the relevant constraints. Can checks be added in Github CI?

baiyangtx:
The community is currently promoting another initiative to use the googleJavaFormatter + spotless plugin to format the code, please refer to this issue https://github.com/NetEase/amoro/issues/1402. Is the goal of this PR consistent with that initiative?

GOODBOY008:
@zhoujinsong @shidayang @baiyangtx PTAL , Thank you~

shidayang (approved):

shidayang:
LGTM

ams/optimizer/common/src/main/java/com/netease/arctic/optimizer/common/AbstractOptimizerOperator.java
```diff
@@ -16,6 +16,8 @@
 
 public class AbstractOptimizerOperator implements Serializable {
   private static final Logger LOG = LoggerFactory.getLogger(AbstractOptimizerOperator.class);
+
+  @SuppressWarnings("all")
```
zhoujinsong, line 20: Maybe add a final modifier rather than the `suppressWarnings` annotation?

zhoujinsong (commented):
Thanks a lot for your contribution!

core/src/test/java/com/netease/arctic/utils/map/TestStructLikeMap.java
```diff
@@ -11,6 +11,7 @@
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Random;
```
baiyangtx, line 16: Duplicated import

hive/src/main/java/com/netease/arctic/hive/HMSClient.java
```diff
@@ -44,59 +38,58 @@ public interface HMSClient {
...
   boolean dropPartition(String dbName, String tblName, List<String> partVals,
                         PartitionDropOptions options) throws TException;
 
   int addPartitions(List<Partition> partitions)
-      throws InvalidObjectException, AlreadyExistsException, MetaException, TException;
+      throws TException;
 
 
-  List<String> getAllTables(String dbName) throws MetaException, TException, UnknownDBException;
+  List<String> getAllTables(String dbName) throws TException;
```
baiyangtx, line 92: It makes sense to throw more semantically clear exceptions in an interface.

ams/optimizer/common/src/main/java/com/netease/arctic/optimizer/common/AbstractOptimizerOperator.java
GOODBOY008: > Maybe add a final modifier rather than the `suppressWarnings` annotation?

Because of unit test helper use reflection to change this value.

zhoujinsong: > How about expose overwrite / set method for this field and without final be applied to this field?

Yes, this is indeed a better way.

zhoujinsong (approved):
LGTM!
Thanks a lot for your great work!

PR #2082, opened 2023-10-12
[AMORO-2081] Add "/ams/v1/health/status" url to whitelist

ShawHee:
fix #2081 

## Why are the changes needed?
When using the "/ams/v1/health/status" url for health check, it was found that the whitelist was not added, making it difficult to use.

## Brief change log
ams/server/src/main/java/com/netease/arctic/server/dashboard/DashboardServer.java
```
private static final String[] urlWhiteList = {
      "/ams/v1/versionInfo",
      "/ams/v1/login",
      "/ams/v1/health/status",
      "/",
      "/overview",
      "/introduce",
      "/tables",
      "/optimizers",
      "/login",
      "/terminal",
      "/hive-tables/upgrade",
      "/hive-tables",
      "/index.html",
      "/favicon.ico",
      "/js/*",
      "/img/*",
      "/css/*",
      IcebergRestCatalogService.ICEBERG_REST_API_PREFIX + "/*"
  };
```
add /ams/v1/health/status to whitelist

## How was this patch tested?

- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (yes / no)  no
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #2134, opened 2023-10-17
[AMORO-1957] Support liveness port testing for standby instance

huyuanfeng2018:

Close ##1957

## Brief change log

- Start a thread to open the port for k8s when the container starts k8s use for livenessprobe

## How was this patch tested?

- [X] Add screenshots for manual tests if appropriate

In my environment it works fine.

image

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
@@ -49,6 +49,12 @@ public class ArcticManagementConf {
           .defaultValue("admin")
           .withDescription("The administrator password");
 
+  public static final ConfigOption<Integer> LIVENESS_PROBE_TCP_PORT =
+      ConfigOptions.key("liveness-probe-tcp-port")
+          .intType()
+          .defaultValue(-1)
+          .withDescription("livenessProbe tcp port of the server.");
+
   public static final ConfigOption<Long> REFRESH_EXTERNAL_CATALOGS_INTERVAL =
       ConfigOptions.key("refresh-external-catalogs.interval")
           .longType()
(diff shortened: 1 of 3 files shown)


baiyangtx (commented):
I think using a separate port as a liveness port is not a good solution. If the goal is to solve the Kubernetes pod liveness probe issue, using the "ps" command is also a viable option. This PR proposes reusing the HTTP port and obtaining the instance status through an HTTP API.

wangtaohz:
> I think using a separate port as a liveness port is not a good solution. If the goal is to solve the Kubernetes pod liveness probe issue, using the "ps" command is also a viable option. This PR proposes reusing the HTTP port and obtaining the instance status through an HTTP API.

Do you think using the HTTP API mentioned here #2082  is a resonable solution?

huyuanfeng2018:
> > I think using a separate port as a liveness port is not a good solution. If the goal is to solve the Kubernetes pod liveness probe issue, using the "ps" command is also a viable option. This PR proposes reusing the HTTP port and obtaining the instance status through an HTTP API.
> 
> Do you think using the HTTP API mentioned here #2082 is a resonable solution? @baiyangtx

Maybe take a look at this:
https://github.com/NetEase/amoro/blob/27bbc6991ce6d495a16527a083f45d89eb1a0268/ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/HealthCheckController.java#L27

baiyangtx:
> > I think using a separate port as a liveness port is not a good solution. If the goal is to solve the Kubernetes pod liveness probe issue, using the "ps" command is also a viable option. This PR proposes reusing the HTTP port and obtaining the instance status through an HTTP API.
> 
> Do you think using the HTTP API mentioned here #2082 is a resonable solution? @baiyangtx

Can this interface be accessed on standby nodes? @wangtaohz

wangtaohz:
> Can this interface be accessed on standby nodes? @wangtaohz

I don't think it can, just like any other HTTP API, only the leader AMS provides HTTP service.

baiyangtx:
Since this PR is not planed to proceed for now, can I close the PR temporarily?  @huyuanfeng2018

huyuanfeng2018:
> Since this PR is not planed to proceed for now, can I close the PR temporarily? @huyuanfeng2018

ok

PR #2138, opened 2023-10-18
[CI] Limit concurrency of workflow

GOODBOY008:
Limit concurrency of workflow to reduce task queuing in GitHub Actions.

### DIFF
--- a/.github/workflows/core-hadoop2-ci.yml
+++ b/.github/workflows/core-hadoop2-ci.yml
@@ -3,6 +3,10 @@
 
 name: Core/hadoop2 CI with Maven
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on:
   push:
     branches:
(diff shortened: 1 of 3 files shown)


GOODBOY008:
J`PS{WUJDOJYB`0UR0VJOBF
@shidayang PTAL.

baiyangtx:
Can you describe the context of this PR? What problem is it trying to solve? What approach is used to address the issue?

GOODBOY008:
Hi @baiyangtx, Limit concurrency of workflow to reduce task queuing in GitHub Actions. When our contributors append commits in a  pr and github action will excute many times until end  ,this will make ci task queuing  . So I want to cancel previous ci task and run latest commit's ci task. As the upper screenshot  shows.

baiyangtx:
@zhoujinsong PTAL

shidayang:
Very useful work. LGTM

shidayang (approved):

PR #2141, opened 2023-10-18
[AMORO-2037]Remove dependency on `org.openjdk.jol:jol-core`

### LINKED ISSUE #2037: [Subtask]: Remove dependency on `org.openjdk.jol:jol-core`
### Search before asking

### Description

As [org.openjdk.jol:jol-corea](https://mvnrepository.com/artifact/org.openjdk.jol/jol-core/0.17) is under `The GNU General Public License `which is conflict with `Apache Licence 2.0`, we need to remove our dependency on it to better comply with the specifications of the `Apache License 2.0`.

### Parent issue

#2012

### Code of Conduct

zhoujinsong:

Close #2037.

## Brief change log

- Remove dependency on `org.openjdk.jol:jol-core`
- Add `org.apache.lucene:lucene-core` dependency for module core
- Use `org.apache.lucene.util.RamUsageEstimator` to estimate the memory size of object for spillable map

## How was this patch tested?

## Documentation


zhoujinsong:
I have tested the results of calculating object memory size using `org.apache.lucene.util.RamUsageEstimator` and `org.openjdk.jol.info.GraphLayout` in my local environment and found:

- The results of calculating for primitive types and collection types are the same.
- The results of calculating for simple objects (objects that do not contain references to other objects) are the same.
- For complex objects, `org.apache.lucene.util.RamUsageEstimator` is inaccurate as it does not recursively calculate the size of referenced objects.
- `org.apache.lucene.util.RamUsageEstimator` has higher execution performance compared to `org.openjdk.jol.info.GraphLayout`, with the former being 20 times faster than the latter.

Because the current `SpillableMap` scenario only requires the calculation of structLike and simple objects, and the calculation process of structLike has been transformed into an array of simple objects, it is feasible to replace `org.openjdk.jol.info.GraphLayout` with `org.apache.lucene.util.RamUsageEstimator` in the current implementation and achieve better performance.

However, it should be noted that `org.apache.lucene.util.RamUsageEstimator` is not a common solution for calculating the size of Java objects.

zhoujinsong:
Additionally, when we downgrade the version of Lucene to 4.0.0, `org.apache.lucene.util.RamUsageEstimator` can accurately calculate the memory usage of objects, but it has compatibility issues with higher versions of JDK and has been abandoned by the Lucene project.

trino/pom.xml
```diff
@@ -353,6 +353,7 @@
         <dependency>
             <groupId>org.openjdk.jol</groupId>
             <artifactId>jol-core</artifactId>
+            <version>${jol.version}</version>
```
wangtaohz, line 356: Does trino moudle still need this dependency `org.openjdk.jol`?

If trino is the only module that must depend on `org.openjdk.jol`, maybe we should declare `jol.version` in the properties of `trino/pom.xml`.

zhoujinsong:
> I noticed that we are using Lucene with `<lucene.version>8.11.2</lucene.version>`. Does it imply that there won't be these compatibility issues with higher versions of JDK?

I think so, the compatibility issues of Lucene version 4.0.0 are because of the use of reflection and it had be removed in higher versions.

trino/pom.xml
zhoujinsong: Yes, I have moved it.

shidayang:
> I have tested the results of calculating object memory size using `org.apache.lucene.util.RamUsageEstimator` and `org.openjdk.jol.info.GraphLayout` in my local environment and found:
> 
> * The results of calculating for primitive types and collection types are the same.
> * The results of calculating for simple objects (objects that do not contain references to other objects) are the same.
> * For complex objects, `org.apache.lucene.util.RamUsageEstimator` is inaccurate as it does not recursively calculate the size of referenced objects.
> * `org.apache.lucene.util.RamUsageEstimator` has higher execution performance compared to `org.openjdk.jol.info.GraphLayout`, with the former being 20 times faster than the latter.
> 
> Because the current `SpillableMap` scenario only requires the calculation of structLike and simple objects, and the calculation process of structLike has been transformed into an array of simple objects, it is feasible to replace `org.openjdk.jol.info.GraphLayout` with `org.apache.lucene.util.RamUsageEstimator` in the current implementation and achieve better performance.
> 
> However, it should be noted that `org.apache.lucene.util.RamUsageEstimator` is not a common solution for calculating the size of Java objects.

It seem RamUsageEstimator only can calculator know type, other type is always 256

shidayang (approved):

wangtaohz (approved):
LGTM

PR #2142, opened 2023-10-18
[AMORO-2078] Replace set with bitmap when apply pos delete records

### LINKED ISSUE #2078: [Improvement]: Change position delete index from set to bitmap of optimizing
### Search before asking

Currently, the optimization task of Iceberg uses a Set data structure for indexing position-delete data, which consumes a lot of memory. It could be changed to a bitmap to reduce memory usage.

### How should we improve?

Modify com.netease.arctic.iceberg.CombinedDeleteFilter position delete index

### Subtasks

_No response_

### Code of Conduct

XBaith:

Close #2078.

## Brief change log

- Reduce memory for large pos-deletes

## How was this patch tested?

## Documentation

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -119,6 +119,12 @@
             <artifactId>paimon-bundle</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.roaringbitmap</groupId>
+            <artifactId>RoaringBitmap</artifactId>
+            <version>0.9.25</version>
+        </dependency>
+
         <!-- test dependencies -->
 
         <dependency>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -122,7 +122,6 @@
         <dependency>
             <groupId>org.roaringbitmap</groupId>
             <artifactId>RoaringBitmap</artifactId>
-            <version>0.9.25</version>
         </dependency>
 
         <!-- test dependencies -->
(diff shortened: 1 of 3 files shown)


core/pom.xml
```diff
@@ -119,6 +119,12 @@
             <artifactId>paimon-bundle</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.roaringbitmap</groupId>
+            <artifactId>RoaringBitmap</artifactId>
+            <version>0.9.25</version>
```
zhoujinsong, line 125: I recommend managing the dependency in the `DependencyManagement` blocker in parent pom and store the version in properties.

zhoujinsong: BTW, should we dependent on version 0.9.44 as Iceberg project do.

shidayang, line 125: Add dependency to dependencyManager

shidayang (changes requested):

core/src/main/java/com/netease/arctic/io/reader/CombinedDeleteFilter.java
```diff
@@ -79,7 +80,7 @@ public abstract class CombinedDeleteFilter<T extends StructLike> {
   private final List<DeleteFile> posDeletes;
   private final List<DeleteFile> eqDeletes;
 
-  private Map<String, Set<Long>> positionMap;
+  private Map<String, Roaring64Bitmap> positionMap;
```
zhoujinsong, line 83: Maybe we can resue the `org.apache.iceberg.deletesPositionDeleteIndex` here?

zhoujinsong (commented):
@XBaith I left some comments.

core/src/main/java/com/netease/arctic/io/reader/CombinedDeleteFilter.java
XBaith: `Deletes.positionDeleteIndex` reqire datafile path which is not exist in `CombinedDeleteFilter`

shidayang (approved):

shidayang:
LGTM

PR #2146, opened 2023-10-18
[hotfix] Catalog type always null in TerminalSession

link3280:
The configuration option is not used correctly and always returns null.

## Brief change log

Use the configuration option's key to get the value in a map.

## How was this patch tested?

## Documentation

- Does this pull request introduce a new feature? (yes / no) no
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented) not applicable


wangtaohz:
Thanks for your contribution! :+1: It is a good catch.

wangtaohz (approved):

PR #2147, opened 2023-10-18
[AMORO-2099] Dashboard Catalog page supports automatically filling of required properties

### LINKED ISSUE #2099: [Subtask]: Dashboard Catalog page supports automatically filling of required properties
### Search before asking

### Description

After select `Metastore` and `Table Format` in the Catalog page, some required properties can be automatically filling to the section `Properties` with empty value, if it doesn't exist, like
image

For different types of `Metastore`, the required properties are different as follows
| Metastore | Type                  | Required Properties |
| ---------- | ---------------- | -------------------- |
| ams           | Internal Catalog | warehouse     |
| hadoop     | External Catalog | warehouse    |
| hive           | External Catalog |    |
| glue           | External Catalog | warehouse,lock-impl,lock-table  |
| custom      | External Catalog | catalog-impl   |

And the Table Format is similar

| Table Format | Required Properties |
| ----------  | -------------------- |
| paimon      | warehouse     |

 I think this requirement can be implemented on the front-end.
### Parent issue

#2007

### Code of Conduct

minteliuwm:

Close #2099.

## Brief change log

- Dashboard Catalog page supports automatically filling of required properties

## How was this patch tested?

## Documentation


ams/dashboard/src/views/catalogs/Detail.vue
```diff
@@ -265,6 +265,13 @@ const authConfigMap = {
   'auth.kerberos.krb5': 'Kerberos Krb5'
 }
 
+const defaultPropertiesMap = {
+  ams: 'warehouse',
```
wangtaohz, line 269: It's better to use a list as value to store these properties for each key (ams,hadoop...), in order to support more required properties for each key in the future.

wangtaohz (commented):
Thanks for you contribution! I tested it locally and it works well.

However, retaining all previous properties after switching metastore or table format can be confusing, because the user is not sure about whether they are required properties. 

I have an idea to solve this problem, which is to delete any key that was required before the switch but is no longer required after the switch.

ams/dashboard/src/views/catalogs/Detail.vue
minteliuwm: @wangtaohz  Thanks for your comment, I've updated the code, please review again.

wangtaohz (approved):
LGTM

wangtaohz:
I tested it locally, it worked well.

image

image

ams/dashboard/src/views/catalogs/Detail.vue
```diff
@@ -313,6 +313,14 @@ const authConfigMap = {
   'auth.kerberos.krb5': 'Kerberos Krb5'
 }
 
+const defaultPropertiesMap = {
+  ams: ['warehouse'],
+  hadoop: ['warehouse'],
+  custom: ['catalog-impl'],
+  glue: ['warehouse', 'lock-impl', 'lock-table'],
```
XBaith, line 320: It should be `lock.table`, refer to https://iceberg.apache.org/docs/latest/configuration/#lock-catalog-properties

wangtaohz: Thank you for your feedback, I will fix it ASAP.

PR #2163, opened 2023-10-21
[Fix] Fix the format of thread name in DefaultTableService

GOODBOY008:
Fix the format of thread name in DefaultTableService


baiyangtx (approved):
That's a good catch!

ams/server/src/main/java/com/netease/arctic/server/table/DefaultTableService.java
```diff
@@ -58,7 +58,7 @@ public class DefaultTableService extends StatedPersistentBase implements TableSe
 
   private final ScheduledExecutorService tableExplorerScheduler = Executors.newSingleThreadScheduledExecutor(
       new ThreadFactoryBuilder()
-          .setNameFormat("table-explorer-scheduler")
+          .setNameFormat("table-explorer-scheduler-%d")
```
zhoujinsong, line 61: It seems to be fine as this is a single-threaded thread pool.

PR #2174, opened 2023-10-24
Rename update.sql file to update-0.5.1-to-0.6.0.sql

shidayang:


wangtaohz (commented):
It should be `upgrade-xxxx` instead of `update-xxx`

shidayang:
> It should be `upgrade-xxxx` instead of `update-xxx`

done

zhoujinsong:
@shidayang We'd better keep the `update.sql` and move the content into the new `uprade-0.5.1-to-0.6.0.sql`.

wangtaohz (approved):
LGTM

shidayang:
> @shidayang We'd better keep the `update.sql` and move the content into the new `uprade-0.5.1-to-0.6.0.sql`.

Your mean also contain a empty update.sql?

zhoujinsong (approved):
LGTM.

PR #2230, opened 2023-11-01
[AMORO-2225] Change the default value of `self-optimizing.major.trigger.duplicate-ratio` to 0.1

zhoujinsong:

Close #225.

## Brief change log

- Change the default value of `self-optimizing.major.trigger.duplicate-ratio` to 0.1
- Rename property `hive.consistent-write.enabled` to `base.hive.consistent-write.enabled` and move it to HiveTableProperties.
- Polish some codes.

## How was this patch tested?

## Documentation


baiyangtx (commented):
There is still code in `TestSparkWriter` that needs to be adjusted and renamed.

zhoujinsong:
> There is still code in `TestSparkWriter` that needs to be adjusted and renamed.

Thanks for the reminder, I have fixed the compile error, please take another look.

wangtaohz (approved):
LGTM

PR #2254, opened 2023-11-06
[AMORO-2248][AMS] Improve full optimizing trigger condition

### LINKED ISSUE #2248: [Improvement]: Improve full optimizing trigger condition
### Search before asking

There are multiple segment files in a table, which can be merged into one file, but the merge is not triggered.

iceberg table format
upsert : false

### How should we improve?

Improve trigger condition

### Subtasks

_No response_

### Code of Conduct

zhongqishang:

Close #2248.

## Brief change log

- Iceberg format add condition `segmentFileCount >= 2`
- mixed format add condition `segmentFileCount > getBaseSplitCount()`
- Filter do not need to execute task

## How was this patch tested?

## Documentation


baiyangtx (approved):
LGTM

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
```diff
@@ -316,6 +316,11 @@ public List<SplitTask> splitTasks(int targetTaskCount) {
                   rewritePosDataFiles.add(f.getFile());
                   deleteFiles.addAll(f.getDeleteFiles());
                 });
+        if (rewriteDataFiles.size() == 1
```
wangtaohz, line 319: Skipping files during task splitting will result in data loss for Mixed Hive Tables, because Mixed Hive tables requires all files within a partition must be involved in an optimizing process.

wangtaohz (changes requested):

wangtaohz:
After careful consideration, the reason why the previous full-optimizing did not trigger based on the number of segment files was to avoid repeatedly optimizing on partitions that no longer needed to be optimized, which would consume a significant amount of resources. 

For example, if all the files in a partition have already reached 128MB (they are all segment files) and no new data is being written, we do not want to repeatedly perform full-optimizing on that partition.

So the key is that we should treat _smaller segment files_ (maybe 16MB ~ 0.9 * 128MB) differently from _ideal segment files_ ( >0.9 * 128MB), and only _smaller segment files_ can trigger full-optimizing.

ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/AbstractPartitionPlan.java
zhongqishang: Maybe we can control whether to filter through a parameter?

wangtaohz: Yes, it may be a feasible approach, but I don't like quietly filtering out some files during splitting tasks, which goes against intuition. Perhaps we should exclude the unnecessary tasks after the split. 

You may introduce an "afterSplit" operation to perform this, similar to `com.netease.arctic.server.optimizing.plan.AbstractPartitionPlan#beforeSplit`.

zhongqishang:
> After careful consideration, the reason why the previous full-optimizing did not trigger based on the number of segment files was to avoid repeatedly optimizing on partitions that no longer needed to be optimized, which would consume a significant amount of resources.

If we filter out split tasks that do not require optimization(data file = 1, eq = 0), can we avoid the optimizer consuming a lot of resources?

> For example, if all the files in a partition have already reached 128MB (they are all segment files) and no new data is being written, we do not want to repeatedly perform full-optimizing on that partition.

In my opinion a waste of resources is probably a lower priority than a large number of files at 16m+

> So the key is that we should treat _smaller segment files_ (maybe 16MB ~ 0.9 * 128MB) differently from _ideal segment files_ ( >0.9 * 128MB), and only _smaller segment files_ can trigger full-optimizing.

A fixed value of 0.9 may not be a good choice, how about we set it to (128m - 16m)?
We can add this later to the discussion of major compation.

Issue #1914, opened 2023-09-04
[Bug]: `BasicUnkeyedTable` does not Override the `name` method


huyuanfeng2018:
### What happened?

`BasicUnkeyedTable` does not Override the `name` method

### Affects Versions

master

### What engines are you seeing the problem on?

Core

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

huyuanfeng2018:
TaskRuntime also does not implement the tostring method, resulting in printing a lot of invalid logs

PR #2256, opened 2023-11-06
[AMORO-2255] Fix BasicIcebergTable's name and serialization ID

### LINKED ISSUE #2255: [Bug]: Incorrect name of BasicIcebergTable
### What happened?

The `name()` of BasicIcebergTable is not correctly implemented and returns the default `toString()` which prints the memory address.

Note that the ArticTable uses Java Object serialization and hasn't set a serialization ID, which would block changes to the BasicIcebergTable class, thus I added a commit to pin the serialization ID before overriding the `name()` method.

### Affects Versions

0.6.0

AMS

### How to reproduce

_No response_

### Relevant log output

```shell
2023-10-31 09:48:10,994 INFO [ASYNC-TableRuntimeRefreshExecutor-3] [org.apache.iceberg.SnapshotScan] [] - Scanning table com.netease.arctic.catalog.IcebergCatalogWrapper$BasicIcebergTable@4550db2b snapshot 1659655291600117103 created at 2023-10-31T01:47:36.277+00:00 with filter true
```

### Anything else

_No response_

### Code of Conduct

link3280:
As titled.

Close #2255.

## Brief change log

-

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/catalog/IcebergCatalogWrapper.java
+++ b/core/src/main/java/com/netease/arctic/catalog/IcebergCatalogWrapper.java
@@ -333,6 +333,11 @@ public class IcebergCatalogWrapper implements ArcticCatalog {
     }
 
     @Override
+    public String name() {
+      return icebergTable.name();
+    }
+
+    @Override
     public TableFormat format() {
       return TableFormat.ICEBERG;
     }
--- a/core/src/main/java/com/netease/arctic/catalog/IcebergCatalogWrapper.java
+++ b/core/src/main/java/com/netease/arctic/catalog/IcebergCatalogWrapper.java
@@ -322,6 +322,8 @@ public class IcebergCatalogWrapper implements ArcticCatalog {
 
   public static class BasicIcebergTable extends BasicUnkeyedTable {
 
+    private static final long serialVersionUID = -5240430071467122543L;
+
     public BasicIcebergTable(
         TableIdentifier tableIdentifier,
         Table icebergTable,


core/src/main/java/com/netease/arctic/catalog/IcebergCatalogWrapper.java
```diff
@@ -330,6 +332,11 @@ public BasicIcebergTable(
       super(tableIdentifier, icebergTable, arcticFileIO, catalogProperties);
     }
 
+    @Override
+    public String name() {
+      return icebergTable.name();
```
zhoujinsong, line 337: The superclass has implemented the `name` method, ref: https://github.com/NetEase/amoro/blob/master/core/src/main/java/com/netease/arctic/table/BasicUnkeyedTable.java#L119

The logging error is caused by some classes in Iceberg that directly use `Table#toString` to obtain the name of the table when logging, so it is necessary to implement the `toString` method in `BasicUnkeyedTable`, ref:https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/SnapshotScan.java#L126

link3280: OK. It seems that it's already fixed by #1914. But how about the absent serialization ID?

zhoujinsong: It seems that there won't be multiple versions of the class in the runtime environment, so it's acceptable to not set a serialization ID.

link3280:
Fixed by https://github.com/NetEase/amoro/issues/1914 for future reference.

PR #2280, opened 2023-11-10
[AMORO-2279] Retry failed tasks when restarting AMS

XBaith:
## Why are the changes needed?

Close #2279

## Brief change log

- Avoding optimizing status blocked

## How was this patch tested?

- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (no)
- If yes, how is the feature documented? (not documented)

PR #2282, opened 2023-11-10
[AMORO-1951] Support parallelized planning in one optimizer group

### LINKED ISSUE #1951: [Improvement]: Support parallelized planning in one optimizer group
### Search before asking

Currently, it is widely reported that there may be idle resources in an optimizer group while multiple tables are in a pending state. The bottleneck is that only single-threaded planning is supported in an optimizer group. When a table requires a long time to planTasks, it can cause the entire resource pool to hang. It is hoped that a multi-threaded and asynchronous planning mechanism can be introduced to improve resource utilization. 

### How should we improve?

This mechanism considers introducing a parameter: 

`max_planning_threads_per-group (or similar name) `

to represent the maximum planning concurrency under a group. Even though there would be more features of task scheduling introduced in the future, this parameter will continue to be used. The current recommended approach is to construct a thread pool in the group based on max_planning_threads_per-group to perform planning. Every time the optimizer pollTask method is triggered, it will initiate asynchronous planning, and return directly when the concurrency reaches _max_planning_threads_per-group_.

### Subtasks

_No response_

### Code of Conduct

majin1102:

Close #1951 .

## Brief change log

- Support parallelized planning by a parameter named as _maxPlanningParallelism_
- Refactor OptimizingQueue and DefaultOptimizingService
- Refactor suspending optimizer checker by using DelayedQueue for better SLA
- add two other global parameters for optimizing:
```
  private final long pollingTimeout; // optimizing max waiting time when polling, default 3000ms
  private final long minPlanningInterval; // min table planning interval, default 60 seconds 
```
- add unit tests for OptimizingQueue and DefaultOptimizingService

## How was this patch tested?

## Documentation

- If yes, how is the feature documented? (under Configurations)


ams/server/src/main/java/com/netease/arctic/server/DefaultOptimizingService.java
```diff
@@ -544,37 +534,18 @@ public void run() {
...
-      if (isOptimzerExpired) {
-        LOG.info("Optimizer {} has been suspended, unregister it", token);
-        unregisterOptimizer(token);
-      } else {
-        OptimizerInstance optimizer = authOptimizers.get(token);
-        detectTimeout(optimizer);
-      }
+    private boolean checkOptimizerTimeout(OptimizerInstance optimizer) {
+      return Objects.equals(optimizer, authOptimizers.get(optimizer.getToken()))
+          && optimizer.getTouchTime() + optimizerTouchTimeout > System.currentTimeMillis();
```
Aireed, line 548: I think it should be less than here. If the last touch time plus the timeout interval is less than the current time, it means that there hasn't been another touch after exceeding the timeout interval.
`optimizer.getTouchTime() + optimizerTouchTimeout  <   System.currentTimeMillis();`

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -121,16 +119,7 @@ private void initTableRuntime(TableRuntimeMeta tableRuntimeMeta) {
...
-        process.getTaskMap().entrySet().stream()
-            .filter(
-                entry ->
-                    entry.getValue().getStatus() == TaskRuntime.Status.SCHEDULED
-                        || entry.getValue().getStatus() == TaskRuntime.Status.ACKED)
-            .forEach(entry -> executingTaskMap.put(entry.getKey(), entry.getValue()));
-        process.getTaskMap().values().stream()
-            .filter(task -> task.getStatus() == TaskRuntime.Status.PLANNED)
-            .forEach(taskQueue::offer);
+        tableQueue.offer(new TableOptimizingProcess(tableRuntimeMeta));
```
XBaith, line 122: When starting OptimizingQueue, we should consider the status of processing some tasks, such as failures. If there are still some tasks failing and the process has not finished yet, the table will block in the optimizing state. This is the reason why I put the failed tasks into the retry queue in PR #2280 . Please also note the status of other tasks so that they can be handled correctly.

majin1102: Yeah, that's correct for earlier codes.

For this PR, the situations of failure task status in running process have been eliminated. What you encountered in #2280 is task failure and task reset or process close operation  are not in one transaction which should be. this is fixed by calling doAsTransaction in TaskRuntime.complete...  check this out

ams/server/src/main/java/com/netease/arctic/server/DefaultOptimizingService.java
majin1102: The logic has been tested in TestDefaultOptimizingQueue, and as you mentioned, this is fixed.

Check it out plz

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -330,75 +216,105 @@ public void updateOptimizerGroup(ResourceGroup optimizerGroup) {
...
+                planningCompleted.signalAll();
+              } finally {
+                scheduleLock.unlock();
+              }
+            });
+  }
+
+  private CompletableFuture<TableOptimizingProcess> doPlanning(TableRuntime tableRuntime) {
+    CompletableFuture<TableOptimizingProcess> future = new CompletableFuture<>();
+    planExecutor.execute(
```
shidayang, line 287: The Future did not perceive the exceptional situation. You can using 
```
 future = CompletableFuture.supplyAsync(() -> xxx, planExecutor);
```
or Add
```
future.completeExceptionally(xxx)
```
in Catch statement

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -330,75 +216,105 @@ public void updateOptimizerGroup(ResourceGroup optimizerGroup) {
...
-        LOG.error(tableRuntime.getTableIdentifier() + " plan failed, continue", e);
-      }
+  }
+
+  private void triggerAsyncPlanning(
+      TableRuntime tableRuntime, Set<TableRuntime> skipTables, long startTime) {
+    LOG.info("Trigger planning table {}", tableRuntime.getTableIdentifier());
+    planningTables.add(tableRuntime);
+    doPlanning(tableRuntime)
+        .whenComplete(
```
shidayang, line 252: You can return the Future
```
return future.whenComplete()
```
and in waitTask using
```
future.get(waitDeadline - currentTime, TimeUnit.MILLISECONDS)
```
to replace  planningCompleted.await(waitDeadline - currentTime, TimeUnit.MILLISECONDS)

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -18,91 +18,89 @@
...
-  private final Map<OptimizingTaskId, TaskRuntime> executingTaskMap = new ConcurrentHashMap<>();
-  private final Map<String, OptimizerInstance> authOptimizers = new ConcurrentHashMap<>();
 
+  private final QuotaProvider quotaProvider;
+  private final Queue<TableOptimizingProcess> tableQueue = new LinkedTransferQueue<>();
+  private final Queue<TaskRuntime> retryTaskQueue = new LinkedTransferQueue<>();
+  private final SchedulingPolicy schedulingPolicy;
   private final TableManager tableManager;
+  private final Executor planExecutor;
+  private final Map<TableRuntime, Long> plannedKeepingTables = new HashMap<>();
```
shidayang, line 80: Need thread safe

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -18,91 +18,89 @@
...
+  private final Set<TableRuntime> planningTables = new HashSet<>();
```
shidayang, line 81: Need thread safe

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -172,154 +156,56 @@ public boolean containsTable(ServerTableIdentifier identifier) {
...
+    return deadline <= 0 ? Long.MAX_VALUE : deadline;
   }
 
-  private void retryTask(TaskRuntime taskRuntime, boolean incRetryCount) {
-    taskRuntime.reset(incRetryCount);
-    retryQueue.offer(taskRuntime);
+  private TaskRuntime fetchTask() {
+    return Optional.ofNullable(retryTaskQueue.poll())
+        .orElse(
+            Optional.ofNullable(tableQueue.peek()).map(TableOptimizingProcess::poll).orElse(null));
```
shidayang, line 182: As long as the TaskQueue is not empty, the plan will not be triggered. It will only be triggered when the TaskQueue is empty. Even when the TaskQueue is empty, it will still wait for a table plan to be completed, making it difficult to demonstrate the effects of multithreading. Perhaps our plan thread's triggering strategy can be based on the capacity of the TaskQueue, automatically triggering the plan in the background when the TaskQueue is relatively empty.

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
majin1102: They are kept safe in lock

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
majin1102: good idea

ams/server/src/main/java/com/netease/arctic/server/resource/QuotaProvider.java
```diff
@@ -0,0 +1,6 @@
+package com.netease.arctic.server.resource;
+
+public interface QuotaProvider {
```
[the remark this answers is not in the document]
majin1102: done

ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
```diff
@@ -236,6 +242,19 @@ public class ArcticManagementConf {
           .defaultValue(30000L)
           .withDescription("Timeout duration for task acknowledgment.");
 
+  /** config key prefix of terminal */
+  public static final ConfigOption<Integer> OPTIMIZER_MAX_PLANNING_PARALLELISM =
+      ConfigOptions.key("optimizer.task-ack-timeout")
```
HuangFru, line 247: wrong config, same with "optimizer.task-ack-timeout".

HuangFru (approved):
I've tested it with `max-planning-parallelism` 1 and 5. Works fine.

HuangFru:
There is a very important point here. In iceberg native plan logic, multi-threaded plan is used by default. And this multi-threaded thread pool is still public (including commit and other operations will also use this thread pool). So when a large table (which may have unhealthy metadata) takes up a lot of resources during planning. Even if multiple tables are allowed to plan at the same time, the plan speed of other tables will become very slow, and even the speed of commit will be greatly affected. This PR has not yet solved this problem.

ams/server/src/main/java/com/netease/arctic/server/persistence/mapper/OptimizingMapper.java
```diff
@@ -112,12 +112,12 @@ List<OptimizingProcessMeta> selectOptimizingProcesses(
         + "end_time, status, fail_reason, optimizer_token, thread_id, rewrite_output, metrics_summary, properties) "
         + "VALUES ",
     "<foreach collection='taskRuntimes' item='taskRuntime' index='index' separator=','>",
-    "(#{taskRuntime.taskId.processId}, #{taskRuntime.taskId.taskId}, #{taskRuntime.retry},"
+    "(#{taskRuntime.taskId.processId}, #{taskRuntime.taskId.taskId}, #{taskRuntime.runTimes},"
```
zhoujinsong, line 115: Do you think we should change the column name in the database table too?

ams/server/src/main/java/com/netease/arctic/server/resource/OptimizerInstance.java
```diff
@@ -53,16 +51,24 @@ public long getTouchTime() {
...
 
-  public OptimizingQueue.OptimizingThread getThread(int threadId) {
-    return new OptimizingQueue.OptimizingThread(token, threadId);
+  public OptimizerThread getThread(int threadId) {
+    return new OptimizerThread(threadId, this);
+  }
+
+  @Override
+  public String toString() {
+    return "OptimizerInstance{"
```
zhoujinsong, line 64: Generate `toString` method with Guava MoreObjects.

ams/server/src/main/java/com/netease/arctic/server/resource/OptimizerThread.java
```diff
@@ -0,0 +1,42 @@
+package com.netease.arctic.server.resource;
+
+import com.netease.arctic.server.optimizing.TaskRuntime;
+
+import java.util.Objects;
+
+public class OptimizerThread {
```
zhoujinsong, line 7: Java doc needed here.

zhoujinsong (commented):
I left some comments, please take another look.
The code style seems to be incorrect, you can follow this guide to correct it:https://github.com/NetEase/amoro/blob/master/CONTRIBUTING.md#code-formatting

ams/server/src/main/java/com/netease/arctic/server/persistence/mapper/OptimizingMapper.java
majin1102: I think we could do this next time.

The logic before is not completed and causes consistency problem. Right now even we don't change column name here, the only consequence is that the true retry num would be one less.

If we change column name here, the upgrade would be complicated which is not necessary for now

zhoujinsong: Looks fine to me.

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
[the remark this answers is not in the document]
majin1102: Use default Object's hashcode() and equals() is exactly what I want for TableRuntime.

There should be only one TableRuntime object for a table

ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
```diff
@@ -140,6 +140,12 @@ public class ArcticManagementConf {
           .defaultValue(60000L)
           .withDescription("Interval for refreshing table metadata.");
 
+  public static final ConfigOption<Long> GLOBAL_MIN_PLANNING_INTERVAL =
+      ConfigOptions.key("planning.min-interval")
```
zhoujinsong, line 144: Is `self-optimizing.min-plan-interval` a better configuration name as we already have a configuration named `self-optimizing.commit-thread-count`. We can group the configurations like below:
```
  self-optimizing:
    commit-thread-count: 10
    min-plan-interval: 60000 # 1min
```

zhoujinsong: BTW, please add the 3 new configurations into `ams/dist/src/main/arctic-bin/conf/config.yaml`.

zhoujinsong (commented):
@majin1102 I left some comments, please task a look.

ams/server/src/main/java/com/netease/arctic/server/resource/OptimizerThread.java
```diff
@@ -0,0 +1,39 @@
+package com.netease.arctic.server.resource;
```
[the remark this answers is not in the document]
majin1102: Thanks for the correction, take a look

ams/server/src/main/java/com/netease/arctic/server/DefaultOptimizingService.java
```diff
@@ -366,47 +439,125 @@ public void handleTableRemoved(TableRuntime tableRuntime) {
...
+    }
+
+    public OptimizerInstance getOptimizer() {
+      return optimizerInstance;
+    }
+  }
+
+  private class OptimizerKeeper implements Runnable {
+
+    private volatile boolean stopped = false;
```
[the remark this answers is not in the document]
majin1102: Running seems a status, not a command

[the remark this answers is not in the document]
zhongqishang: Yes, all are defined as `stopped` in Amoro. 
Ignore this.

ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
```diff
@@ -140,6 +140,12 @@ public class ArcticManagementConf {
           .defaultValue(60000L)
           .withDescription("Interval for refreshing table metadata.");
 
+  public static final ConfigOption<Long> GLOBAL_MIN_PLANNING_INTERVAL =
+      ConfigOptions.key("self-optimizing.min-plan-interval")
+          .longType()
+          .defaultValue(60000L)
+          .withDescription("Minimum interval for planning on a single table.");
```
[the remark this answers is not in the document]
majin1102: done

ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
```diff
@@ -254,6 +254,19 @@ public class ArcticManagementConf {
           .defaultValue(30000L)
           .withDescription("Timeout duration for task acknowledgment.");
 
+  /** config key prefix of terminal */
```
zhoujinsong, line 257: The comment here seems to be incorrect.

zhoujinsong (approved):
LGTM.

zhongqishang (approved):
LGTM.

PR #2298, opened 2023-11-13
[AMORO-2296] Add cache for loading tables in core and trino

### LINKED ISSUE #2296: [Improvement]: Add cache for loading table when get table meta in unkeyed table
### Search before asking

With a query in Trino, operations such as fetching metadata information require multiple loading tables. This can be affected by objective factors such as network latency, resulting in slower queries. 

In the Iceberg format, this caching is already implemented (native Iceberg implementation). We have this cache for our Mix-format's keyed tables, and we should add this cache for our unkeyed tables as well. This cache is also applicable to queries on the base store and change store.

### How should we improve?

Add a cache for loading tables.

### Subtasks

_No response_

### Code of Conduct

HuangFru:

Close #2296.

## Brief change log

- Add a cache for loading tables.
- Implement caching catalog in `BasicMixedIcebergCatalog`.

## How was this patch tested?

## Documentation

- Does this pull request introduce a new feature? (**no**)
- If yes, how is the feature documented? (**not applicable**)


trino/src/main/java/com/netease/arctic/trino/ArcticConnectorMetadata.java
```diff
@@ -81,6 +82,8 @@ public class ArcticConnectorMetadata implements ConnectorMetadata {
 
   private final ArcticCatalog arcticCatalog;
 
+  private final Map<SchemaTableName, ArcticTable> tableCache = new ConcurrentHashMap<>();
+
   public ArcticConnectorMetadata(
```
baiyangtx, line 87: Since Cache has already been implemented in the core, is it necessary to re-implement caching in Trino?

HuangFru: I think it's necessary because loading the mix table not only loads the iceberg table but also has some other operations and I'm not sure whether the impact is big or not.

baiyangtx: It just loads iceberg table.

baiyangtx: It just loads iceberg table.

HuangFru: There are still some 'if' judgment and wrapper operations in both 'trino' and 'core'. And the default cache expiration in the caching catalog is 30s. Trino query will easily exceed 30 seconds. For me, I don’t want to go to AMS to set the expiration time every time.

shidayang (approved):

shidayang:
LGTM

trino/src/main/java/com/netease/arctic/trino/ArcticConnectorMetadata.java
baiyangtx: Maybe you can overwrite AMS configured catalog properties in Trino engine configs.

HuangFru: The life cycles of these two caches should be different. If I set a long expiration time in the catalog properties on the trino side, then the engine-side cache that should have been destroyed when the trino "ConnectorMetadata" life cycle ends will now have to wait for the time to expire, which may not be reasonable. If the setting time is too small, cache utilization will decrease.

baiyangtx (approved):
LGTM

PR #2304, opened 2023-11-14
[AMORO-2302]: Donot trigger optimizing for mixed iceberg table store in Iceberg catalog.

### LINKED ISSUE #2302: [Bug]:  The "change store" was mistakenly triggered for optimization.
### What happened?

The "change store" was independently triggered for optimizing, resulting in duplicate data on the mixed-iceberg table.

### Affects Versions

master

AMS

### How to reproduce

Creating a mixed-iceberg catalog of type "hive" and then creating an iceberg catalog of type "hive." 

After creating a table under the mixed-iceberg catalog, two independent iceberg tables will be visible under the iceberg catalog. 

At this point, mistakenly triggering optimization on both iceberg tables will occur.

be4abcf3aadeb54b7d0ab140f7f77f3

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

baiyangtx:

Close #2302.

## Brief change log
 
- Ignore Iceberg if it is a table store of mixed-iceberg when sync table in external catalog

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/table/DefaultTableService.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/table/DefaultTableService.java
@@ -27,6 +27,7 @@ import com.netease.arctic.server.persistence.mapper.CatalogMetaMapper;
 import com.netease.arctic.server.persistence.mapper.TableMetaMapper;
 import com.netease.arctic.server.table.blocker.TableBlocker;
 import com.netease.arctic.server.utils.Configurations;
+import com.netease.arctic.utils.TablePropertyUtil;
 import org.apache.commons.lang.StringUtils;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.iceberg.relocated.com.google.common.base.MoreObjects;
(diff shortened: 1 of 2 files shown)


ams/server/src/main/java/com/netease/arctic/server/table/DefaultTableService.java
```diff
@@ -622,6 +623,11 @@ private boolean handleTableRuntimeAdded(
             tableIdentity.getDatabase(), tableIdentity.getTableName());
     AmoroTable<?> table =
         externalCatalog.loadTable(tableIdentity.getDatabase(), tableIdentity.getTableName());
+    if (TableFormat.ICEBERG == table.format()) {
```
wangtaohz, line 626: This would result in `table_identifier` having this table, but `table_runtime` and the `tableRuntimeMap` in memory do not have this table.

But I have checked the relevant codes, and it seems that it will not cause any issues. 👍

wangtaohz (approved):
LGTM.

core/src/main/java/com/netease/arctic/utils/TablePropertyUtil.java
```diff
@@ -133,4 +134,19 @@ public static long getTableWatermark(Map<String, String> properties) {
...
     }
   }
+
+  /**
+   * Check if an iceberg table is a table store of a mixed-iceberg table.
+   *
+   * @param properties table properties of iceberg
+   * @return true if this it a table store.
+   */
+  public static boolean isMixedTableStore(Map<String, String> properties) {
```
wangtaohz, line 144: The Mixed Hive Format also has the same problem, and the unkeyed Mixed Format table should also be considered.

baiyangtx: The table store of mixed-hive will not be loaded by iceberg catalog

PR #2310, opened 2023-11-15
[HotFix] Re-add table filter

HuangFru:

We've lost table-filter in #2232.

## Brief change log

- Add table filter in UnifiedCatalog

## How was this patch tested?

## Documentation


baiyangtx (commented):
BasicMixedIcebergCatalog   also need this.

wangtaohz:
I think table-filter should be a generic feature for the external catalog. Maybe you can help confirm if it is feasible. @HuangFru

HuangFru:
> I think table-filter should be a generic feature for the external catalog. Maybe you can help confirm if it is feasible. @HuangFru

What do you mean? Judging from the current code structure, it should be effective for all external catalogs.

shidayang:
Table filter is only for external catalog, Unified includes internal and external

HuangFru:
> Table filter is only for external catalog, Unified includes internal and external

I got you.

ams/server/src/main/java/com/netease/arctic/server/catalog/ExternalCatalog.java
```diff
@@ -15,19 +16,28 @@
...
   TableMetaStore tableMetaStore;
+  private Pattern tableFilterPattern;
 
   protected ExternalCatalog(CatalogMeta metadata) {
     super(metadata);
     this.tableMetaStore = CatalogUtil.buildMetaStore(metadata);
     this.unifiedCatalog =
         this.tableMetaStore.doAs(
             () -> new CommonUnifiedCatalog(this::getMetadata, Maps.newHashMap()));
+    if (metadata.getCatalogProperties().containsKey(CatalogMetaProperties.KEY_TABLE_FILTER)) {
```
shidayang, line 34: Can get CatalogMetaProperties.KEY_TABLE_FILTER directly, and then determine if tableFilter is null.

HuangFru: Fixed.

baiyangtx:
> Table filter is only for external catalog, Unified includes internal and external

Table filter works for AMS,  but InternalCatalog will not use UnifiedCatalog for ServerCatalog implementation.

baiyangtx:
> BasicMixedIcebergCatalog



> > What do you mean? Judging from the current code structure, it should be effective for all external catalogs.
> 
> Yes, but I find a `databaseFilterPattern` in `com.netease.arctic.mixed.BasicMixedIcebergCatalog`. Maybe @baiyangtx can provide more information.

`BasicMixedIcebergCatalog` is used in external catalog. 

The catalog type[Internal/External] is a concept of AMS Instance,  for engine side, all catalogs have no different.

HuangFru:
> > BasicMixedIcebergCatalog
> 
> > > What do you mean? Judging from the current code structure, it should be effective for all external catalogs.
> > 
> > 
> > Yes, but I find a `databaseFilterPattern` in `com.netease.arctic.mixed.BasicMixedIcebergCatalog`. Maybe @baiyangtx can provide more information.
> 
> `BasicMixedIcebergCatalog` is used in external catalog.
> 
> The catalog type[Internal/External] is a concept of AMS Instance, for engine side, all catalogs have no different.

So you mean the engine side also needs the table filter?

zhoujinsong:
AFAIK, the table filter feature helps AMS filter out tables that are not relevant for attention, and currently, only the external catalogs require it. Although the `BasicMixedIcebergCatalog` now has an implementation of a database filter, I believe it should be removed as the engine side does not have any scenarios that require filtering tables or databases. This feature should be specific to AMS.

HDYT? @HuangFru @wangtaohz @baiyangtx

HuangFru:
> AFAIK, the table filter feature helps AMS filter out tables that are not relevant for attention, and currently, only the external catalogs require it. Although the `BasicMixedIcebergCatalog` now has an implementation of a database filter, I believe it should be removed as the engine side does not have any scenarios that require filtering tables or databases. This feature should be specific to AMS.
> 
> HDYT? @HuangFru @wangtaohz @baiyangtx

There is indeed no scenario on the engine side that requires filtering tables or databases, I'll rollback the change.

ams/server/src/test/java/com/netease/arctic/server/catalog/TestServerCatalog.java
```diff
@@ -18,73 +18,134 @@
...
   }
 
   @Parameterized.Parameters(name = "{0}")
   public static Object[] parameters() {
     return new Object[] {
-      PaimonHadoopCatalogTestHelper.defaultHelper(),
-      PaimonHiveCatalogTestHelper.defaultHelper(),
-      IcebergHadoopCatalogTestHelper.defaultHelper(),
-      IcebergHiveCatalogTestHelper.defaultHelper()
+      new PaimonHadoopCatalogTestHelper(
```
wangtaohz, line 55: The changes will affect all test cases. I suggest not testing in this way. How about add a new test case like 
```java
  @Test
  public void listTablesWithTableFilter() {
    // 1.create some databases and tables
    // 2.check the tables without filter
    // 3.set table filter
    getExternalCatalog().updateMetadata(); 
    // 4.check the table list with filter
    // 5.finally unset the table filter, remove the databases and tables in this test case
  }

```

HuangFru: done.

wangtaohz (approved):
LGTM

HuangFru:
> Should `table-filter` needs to be applied to the `listDatabase` interface ?

I think it's hard to cut a db filter from the table filter to apply to the `listDatabase`, have some ideas?

zhongqishang:
> > Should `table-filter` needs to be applied to the `listDatabase` interface ?
> 
> I think it's hard to cut a db filter from the table filter to apply to the `listDatabase`, have some ideas?

Have you considered restoring the previous parameter `database.filter-regular-expression` to support listDatabase?

PR #2313, opened 2023-11-15
[BugFix][Spark]: Add default value of table-filters in spark scan builder to avoid NPE when comparing the table scan.

baiyangtx:
 
The NPE may be throwed when spark comparing the table scan during plan.

## Brief change log

- Add default value to SparkScanBuilder

## How was this patch tested?

## Documentation


wangtaohz:
Can we check the `com.netease.arctic.spark.reader.KeyedSparkBatchScan#filterExpressions` should not be null in the constructor? And then we can remove all the `filterExpressions != null` before using it.

baiyangtx:
> Can we check the `com.netease.arctic.spark.reader.KeyedSparkBatchScan#filterExpressions` should not be null in the constructor? And then we can remove all the `filterExpressions != null` before using it.

Fixed

wangtaohz (approved):
LGTM

PR #2322, opened 2023-11-18
[AMORO-2115] Support displaying Optimizing tasks

### LINKED ISSUE #2115: [Subtask]: Dashboard supports displaying detailed information for optimizing tasks on Web UI
### Search before asking

### Description

To support a Detail page for each Optimizing process, which displays the detailed information for each optimizing tasks, like
image

The Http interface is provided by #2117 

### Parent issue

- #2114

### Code of Conduct

wangtaohz:

Close #2115 .

## Brief change log

- add optimizing detail page with all optimizing taskes
- get optimizing taskes http interface return optimizer token

## How was this patch tested?

## Documentation


wangtaohz:
image

wangtaohz:
image

wangtaohz:
image

wangtaohz:
image

shidayang:
image
 In the task details, I found some information that is not very clear. Could you please prefix the newly added files with "new" and the old files with "reWrite"?

zhongqishang:
> In the task details, I found some information that is not very clear. Could you please prefix the newly added files with "new" and the old files with "reWrite"?

@wangtaohz I suggest use new define parameter to display in page, this is also the reason why I added new parameters.

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
```diff
@@ -94,6 +94,8 @@ private void finish(RewriteFilesOutput filesOutput) {
           summary.setNewDeleteFileCnt(OptimizingUtil.getFileCount(filesOutput.getDeleteFiles()));
           summary.setNewDeleteSize(OptimizingUtil.getFileSize(filesOutput.getDeleteFiles()));
           summary.setNewDeleteRecordCnt(OptimizingUtil.getRecordCnt(filesOutput.getDeleteFiles()));
+          summary.setNewFileSize(summary.getNewDataSize() + summary.getNewDeleteSize());
+          summary.setNewFileCnt(summary.getNewFileCnt() + summary.getNewFileCnt());
```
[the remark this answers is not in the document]
wangtaohz: fixed.

wangtaohz:
> In the task details, I found some information that is not very clear. Could you please prefix the newly added files with "new" and the old files with "reWrite"?

Thanks for your report. After merging this [PR](https://github.com/NetEase/amoro/pull/2262), and modify the order of metrics, the new metrics are more clearer
image

wangtaohz:
> @wangtaohz I suggest use new define parameter to display in page, this is also the reason why I added new parameters.

All the new define parameters are displayed now. You can refer to my last comment.

zhongqishang:
@wangtaohz 

Is it still necessary to display the deprecated parameters (`reRowDeletedDataFileCnt`, `positionalDeleteSize`, `newFileSize`, `newFileCnt`) on the page? 
It will cause some doubts to users, I recommend not display them.

wangtaohz:
There are two reasons for displaying the old parameters here. 
Firstly, since the new parameters of the old task are all 0, omitting the display of old parameters would result in the loss of some information for the old task. 
Secondly, I am not very familiar with how to remove these parameters on the front-end. 😅

tcodehuber:
> image
Looks great. Can we add an action to kill/cancel the running process in this optimizing process detail page?

wangtaohz:
> Looks great. Can we add an action to kill/cancel the running process in the optimizing process detail page?

Yes, that's what we are going to do after this [PR](https://github.com/NetEase/amoro/pull/2297) merged.

wangtaohz:
Show the metrics of optimizing process.
image

zhoujinsong:
@wangtaohz Thanks a lot for your work.

For the display of optimizing process details, I have some improvement suggestions:

1. Currently, the metric names for process/task are not very intuitive. I suggest redesigning them by referencing the names used in the Iceberg snapshot summary.
2. Hide metric information that is not necessary to focus on, rather than displaying all of it.
3. Can the display of process metrics be consistent with that of tasks, by adding a "+" button at the beginning of the table to expand?

wangtaohz:
After modification, the pages are like this:
image

image

zhongqishang:
Sorry, I may have some different opinions, from my opinion, this is a detailed page and should show all the information involved, like snapshot page. Instead of showing only part of it.
@zhoujinsong @wangtaohz

wangtaohz:
> Sorry, I may have some different opinions, from my opinion, this is a detailed page and should show all the information involved, like snapshot page. Instead of showing only part of it.

Do you mean `showing only part of the metrics` when you say `showing only part of it`? As far as I know, the metrics of the snapshot are also selectively displayed, not all information.

zhongqishang:
> > Sorry, I may have some different opinions, from my opinion, this is a detailed page and should show all the information involved, like snapshot page. Instead of showing only part of it.
> 
> Do you mean `showing only part of the metrics` when you say `showing only part of it`? As far as I know, the metrics of the snapshot are also selectively displayed, not all information.

I mean is that I can get from the information what exactly the current optimize task has done, is rewrite data file or rewrite data pos delele, rewriting 3 datafiles into 1, or rewriting datafile's pos delete file, or outout pos delete count ...

zhoujinsong:
> Sorry, I may have some different opinions, from my opinion, this is a detailed page and should show all the information involved, like snapshot page. Instead of showing only part of it. @zhoujinsong @wangtaohz

I agree that we should display the complete metric. However, this does not mean that all processes/tasks will have the same metrics. For instance, with respect to Iceberg v1 tables, there are no metrics related to delete files. Different processes/tasks should only display metrics relevant to themselves.

wangtaohz:
image
image

wangtaohz:
https://github.com/NetEase/amoro/assets/103108928/3954dec3-ba20-4baa-b4d7-b3879a5f426c

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
```diff
@@ -345,7 +347,10 @@ private Set<Status> getNext() {
     }
 
     public synchronized boolean tryAccepting(Status targetStatus) {
-      if (owner.isClosed() || !getNext().contains(targetStatus)) {
+      if (owner.isClosed() && targetStatus != Status.CANCELED) {
```
zhoujinsong, line 350: I'm not sure if treating `CANCELED` as a special status here is a good approach, or if we should include `CANCELED` in the next status list for all possible statuses to express this logic.

zhoujinsong (commented):
@wangtaohz I left a comment, please take a look.

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
wangtaohz: The owner here is the optimizing process. When I try to close the optimizing process and its tasks, the optimizing process is set to be `Closed` first, and then the tasks are rejected changing the status to `CANCELED`.

zhongqishang:
@wangtaohz After I canceled the optimize task, tasks without start time showed abnormal cost time values.

image

wangtaohz:
> @wangtaohz After I canceled the optimize task, tasks without start time showed abnormal cost time values.

Thanks for your report. I will fix it.

wangtaohz:
https://github.com/NetEase/amoro/assets/103108928/f40ea0ac-63a4-4d17-b764-ee13fac615f5

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
```diff
@@ -108,7 +110,9 @@ void reset() {
     invokeConsisitency(
         () -> {
           statusMachine.accept(Status.PLANNED);
-          doAs(OptimizingMapper.class, mapper -> mapper.updateTaskStatus(this, Status.PLANNED));
+          startTime = ArcticServiceConstants.INVALID_TIME;
+          endTime = ArcticServiceConstants.INVALID_TIME;
+          persistTaskRuntime(this);
```
zhongqishang, line 115: `reset()` should reset all run fields `costTime`, `token`, `threadId`, `failReason`.

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
```diff
@@ -128,8 +132,6 @@ public void ack(OptimizerThread thread) {
         () -> {
           validThread(thread);
           statusMachine.accept(Status.ACKED);
```
zhongqishang, line 134: Literally speaking, would it be more appropriate to set threadId and token in the ack stage? Of course schedule can also get `threadId`.

zhongqishang (commented):
I left some comments.

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
wangtaohz: Make sense to me.

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
wangtaohz: The `SCHEDULED` tasks also should have the correct token and threadId, when checking suspending tasks, this token will be used. Here is the related code `com.netease.arctic.server.DefaultOptimizingService.OptimizerKeeper#buildSuspendingPredication`
```java
    private Predicate<TaskRuntime> buildSuspendingPredication(
        String token, boolean isOptimizerExpired) {
      return task -> {
        if (isOptimizerExpired) {
          return token.equals(task.getToken()); // Here, tasks with the same token with expired optimizer should retry
        } else {
          return token.equals(task.getToken())
              && task.getStatus() == TaskRuntime.Status.SCHEDULED
              && task.getStartTime() + taskAckTimeout < System.currentTimeMillis();
        }
      };
    }
```

zhongqishang: I see, igonre.

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
wangtaohz: I changed it to canceling all tasks before closing the optimizing process.

zhongqishang (approved):
LGTM.

zhoujinsong (approved):
LGTM.

PR #2323, opened 2023-11-19
[AMORO-2290] Add `planning` optimizing status for tables

### LINKED ISSUE #2290: [Improvement]: Add `planning` optimizing status for tables
### Search before asking

Amoro tables have 6 optimizing status for now, which is idle、pending、committing、minor、major and full. However, this does not fully cover all the status of the table. For example, when the table is in the process of plan optimizing task, its status remains as `pending` which can be confusing for users.

We should add `planning` optimizing status for tables.``

### How should we improve?

Add `planning` optimizing status for tables. When the table starts the optimizing planning process, its status should change from `pending` to `planning` After generating tasks, its status should change to `minor`,  `major`, or `full`.

### Subtasks

_No response_

### Code of Conduct

huyuanfeng2018:

Close #2290.

## Brief change log

- Add planning optimizing status for tables

image

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM, thanks a lot for your contribution!

ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
```diff
@@ -208,10 +228,10 @@ public void cleanPendingInput() {
     invokeConsisitency(
         () -> {
           pendingInput = null;
-          if (optimizingStatus == OptimizingStatus.PENDING) {
+          if (optimizingStatus == OptimizingStatus.PLANNING) {
```
majin1102, line 231: Why pending status not handled?

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
```diff
@@ -370,6 +369,7 @@ private void planTasks() {
           continue;
         }
         plannedTables.add(table.id());
+        tableRuntime.beginPlanning();
```
majin1102, line 372: I suggest using a try finally block to begin planning and clean planning status

ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
majin1102: I see, this method is somehow coupled with beginPlanning

majin1102 (changes requested):

ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
huyuanfeng2018: In fact, the previous `pending` is split into two states: `pending` and `planning` , so the judgment conditions here need to be updated.

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
huyuanfeng2018: In the current try catch block, if an exception occurs, the `planning` state will be converted to `pending`. If it is normal, it will be converted to the corresponding state of `optmizing`, such as `major`. If there is no `plan` out of `processtask`, `cleanPendingInput` will be called

ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
majin1102: ye~

I mean what if current status is pending?
We know that there's no such situation here because this method is only called after beginPlanning. That's why I say there's a couple with beginPlanning, which could be decoupled. 

How about adding pending status logic here. Image some day we don't always call clearPendingInput after beingPlanning.

ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
majin1102: Make sense

ams/server/src/main/java/com/netease/arctic/server/table/TableRuntime.java
huyuanfeng2018: > mean what if current status is pending? We know that there's no such situation here because this method is only called after beginPlanning. That's why I say there's a couple with beginPlanning, which could be decoupled.
> 
> How about adding pending status logic here. Image some day we don't always call clearPendingInput after beingPlanning.

get.

HuangFru (approved):
I tested the entire optimizing process and the restart of AMS in the planning status locally, and there were no problems.
截屏2023-11-23 15 29 58

majin1102 (approved):
LGTM

PR #2362, opened 2023-11-26
[AMORO-1716] [Improvement]: sort the table list returned by server

### LINKED ISSUE #1716: [Improvement]: Needs to sort the external Iceberg table list
### Search before asking

Currently, an external Iceberg catalog's table list is returned by the Iceberg catalog's `listTables` method which can not guaranteed order.
For example, I have 500 hundred tables named 'mock_table_1' to 'mock_table_500', but the table list in AMS shows like this:

image

It's not very aesthetically pleasing and inconvenient.

### How should we improve?

Sorting it by the lexicographical order of the table names.

### Subtasks

_No response_

### Code of Conduct

chenyuzhi459:

Close #1716

## Brief change log

-

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/TableController.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/TableController.java
@@ -65,6 +65,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -476,7 +477,7 @@ public class TableController {
           .filter(e -> !arcticTables.contains(e))
           .forEach(e -> tables.add(new TableMeta(e, TableMeta.TableType.HIVE.toString())));
     }
-
+    tables.sort((Comparator.comparing(TableMeta::getName)));
     ctx.json(
         OkResponse.of(
             tables.stream()


ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/TableController.java
```diff
@@ -476,7 +477,7 @@ public void getTableList(Context ctx) {
           .filter(e -> !arcticTables.contains(e))
           .forEach(e -> tables.add(new TableMeta(e, TableMeta.TableType.HIVE.toString())));
     }
-
+    tables.sort((Comparator.comparing(TableMeta::getName)));
```
wangtaohz, line 480: Sorting here will mix the result of arctic tables and hive tables, how about sorting them separately?

chenyuzhi459: Well, the final result returned is the mix of arctic tables and hive tables, I have no idea about the meaning of sorting separately

wangtaohz: `sorting separately` means sorting arctic tables and hive tables separately, so all arctic tables will be placed before all hive tables.

By sorting separately rather than having them mixed together, users can quickly identify the existence of Arctic/Hive tables under a database.

zhoujinsong: @wangtaohz Are you suggesting that we should sort the returned results in the order of (format, table_name) instead of just (table_name)? Considering that the Catalog may indeed return tables with multiple formats, I'm not sure if we need to aggregate tables of different formats together. Additionally, do tables of different formats have any priority or precedence?

wangtaohz: Yes, I believe it is better **NOT** to mix tables with different formats together, and sorting the results in the order of (format, table_name) is a good implementation method. 

As for the priority of different formats,  maybe sorting them in alphabetical order is acceptable, which is `arctic`,`hive`,`iceberg`,`paimon`.

The code may be like
```java
    tables.sort(
        (table1, table2) -> {
          if (Objects.equals(table1.getType(), table2.getType())) {
            return table1.getName().compareTo(table2.getName());
          } else {
            return table1.getType().compareTo(table2.getType());
          }
        });
```

zhoujinsong: However, Hive format tables are quite different here in that they are not managed by AMS. AMS only displays them for the purpose of upgrades. Their priority seems to be lower.

Besides, the `Arctic` table type is out-of-date, we may use the format here directly, but it may require frontend changes, we can optimize it by other PR.

chenyuzhi459: Perhaps we should go back to the purpose of this pr, as described in the issue #1716 . The main purpose is to quickly locate a table among multiple tables, so it is more effective to do global sorting tables returned by a catalog. 


Considering that front-end displays now don't distinguish between formats, I think it is not very helpful for solving the problem of "quickly locate a table"  if the back-end separate sorting. And  separate sorting will bring more difficult to locate a table, and the code readability deteriorates, and you need to explain why this should be sorted separately.

HuangFru: Two aspects:
First, for this PR corresponding to the Issue, the purpose is to conveniently and quickly find the tables of interest. If I aim to look for tables in Iceberg format that start with the letter "b," but the sorting is based on (format, table_name), I might encounter a Paimon format table or a Mix format table first, which is not convenient.

Second, considering the current catalog architecture, we have different formats under one catalog. Mixing these formats together is neither aesthetically pleasing nor convenient.

Therefore, I believe we indeed need to sort according to (format, table_name), but the frontend also needs to change. The frontend needs to set up dropdowns (select boxes or others) for different formats under the catalog to display tables of different formats. This approach is both aesthetically pleasing and convenient, addressing various needs. For example, if I want to find an Iceberg table starting with the letter "b," I only need to click on "Iceberg format" and skip past tables starting with the letter "a" to find those starting with "b," without having to skip over tables of other formats that might exist above.
Something like this:
image

ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/TableController.java
```diff
@@ -528,9 +531,20 @@ public void getTableList(Context ctx) {
           tables.stream().map(TableMeta::getName).collect(Collectors.toSet());
       hiveTables.stream()
           .filter(e -> !arcticTables.contains(e))
-          .forEach(e -> tables.add(new TableMeta(e, TableMeta.TableType.HIVE.toString())));
+          .forEach(e -> lackHiveTables.add(new TableMeta(e, TableMeta.TableType.HIVE.toString())));
     }
-    tables.sort((Comparator.comparing(TableMeta::getName)));
+    // sort by format
+    tables.sort(
```
wangtaohz, line 537: We can sort within the stream API when the list of `tables` is built, then the codes can be significantly simplified, like 
```java
 List<TableMeta> tables =
        tableService.listTables(catalog, db).stream()
            .map(
                idWithFormat ->
                    new TableMeta(
                        idWithFormat.getIdentifier().getTableName(),
                        formatToType.apply(idWithFormat.getTableFormat())))
            .sorted(xxxxx) // changed here
            .collect(Collectors.toList());
```
```java
      hiveTables.stream()
          .filter(e -> !arcticTables.contains(e))
          .sorted() // changed here
          .forEach(e -> tables.add(new TableMeta(e, TableMeta.TableType.HIVE.toString())));

Comment (chenyuzhi459) [reply ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/TableController.java]: done

File: ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/TableController.java [hunk 3]
```diff
@@ -517,8 +518,18 @@ public void getTableList(Context ctx) {
...
+                (table1, table2) -> {
+                  if (Objects.equals(table1.getType(), table2.getType())) {
+                    return table1.getName().compareTo(table2.getName());
+                  } else {
+                    return table1.getType().compareTo(table2.getType());
+                  }
+                })
             .collect(Collectors.toList());
 
+    // hive tables has lower priority, append to the  end
```
zhoujinsong, line 532:
```suggestion
    // Hive tables have lower priority, append to the end
```

ams/server/src/main/java/com/netease/arctic/server/dashboard/controller/TableController.java
```diff
@@ -517,8 +518,18 @@ public void getTableList(Context ctx) {
                     new TableMeta(
                         idWithFormat.getIdentifier().getTableName(),
                         formatToType.apply(idWithFormat.getTableFormat())))
+            // sort by format
```
zhoujinsong, line 521:
```suggestion
            // Sort by table format and table name
```

zhoujinsong (commented):
I left some small suggestion about the comment. :)

wangtaohz:
@chenyuzhi459 You may need to sign NetEase's CLA before you start to contribute to Amoro, you can conveniently complete the signing via the link: https://cla-assistant.io/NetEase/amoro

chenyuzhi459:
> @chenyuzhi459 You may need to sign NetEase's CLA before you start to contribute to Amoro, you can conveniently complete the signing via the link: https://cla-assistant.io/NetEase/amoro

done

PR #2384, opened 2023-11-30
[AMORO-2383] Add serialVersionUID to RewriteFilesInput

### LINKED ISSUE #2383: [Improvement]: Add serialVersionUID to RewriteFilesInput
### Search before asking

In version compatibility scenarios, we avoid deserialization failures caused by the different automatically generated serialVersionUIDs by JDK.

d67433ec97330980c650799b4c1ee7b

### How should we improve?

Add serialVersionUID to RewriteFilesInput

### Subtasks

_No response_

### Code of Conduct

hameizi:

Close #2383 .

## Brief change log

- Add serialVersionUID to RewriteFilesInput

## How was this patch tested?

## Documentation


baiyangtx (approved):
LGTM;

core/src/main/java/com/netease/arctic/optimizing/BaseOptimizingInput.java
```diff
@@ -25,6 +25,7 @@
 
 public abstract class BaseOptimizingInput implements TableOptimizing.OptimizingInput {
 
+  private static final long serialVersionUID = -2176390709831230807L;
```
zhoujinsong, line 28: I don't think we should add the `serialVersionUID` for the abstract class.

core/src/main/java/com/netease/arctic/optimizing/RewriteFilesInput.java
```diff
@@ -32,6 +32,7 @@
 import java.util.stream.Collectors;
 
 public class RewriteFilesInput extends BaseOptimizingInput {
+  private static final long serialVersionUID = -696610318564333923L;
```
zhoujinsong, line 35: Maintaining a manual `serialVersionUID` for RewriteFilesInput is necessary as the serialized result of this class will be stored, allowing for potential cross-version usage. 

However, the change seems to bring about compatibility issues during the upgrade process. How about calculating the default `serialVersionUID` value for the class for now and declaring it to prevent upgrade issues?

hameizi: The serialization ID is generated based on the current code, and it is the same as the serialization ID generated by the JDK by default.

zhoujinsong: Nice catch, that is what I expected.

PR #2116, opened 2023-10-16
[ARMORO-1876]Support display paimon snapshots in 'Transactions' page

shidayang:
## Why are the changes needed?
Close #1876 .

## Brief change log
- Implement getTransactions and getTransactionDetail method of PaimonTableDescriptor

## How was this patch tested?

- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (yes / **no**)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / **not documented**)

PR #2393, opened 2023-11-30
[AMORO-2386][AMS] Configure `iceberg.worker.num-threads` in the config.yaml

### LINKED ISSUE #2386: [Improvement]: Configure `iceberg.worker.num-threads` in the configuration file
### Search before asking

When `Runtime.getRuntime().availableProcessors()` is not obtained correctly or the obtained value is less than or equal to 2, the default value of iceberg is 2, AMS is congested in thread `iceberg-work-pool-xx`.

### How should we improve?

Add parameters in `config.yaml` to overwrite the value of `iceberg.worker.num-threads`.

### Subtasks

_No response_

### Code of Conduct

zhongqishang:

Close #2386 .

## Brief change log

- Configure the `iceberg.worker.num-threads` parameter through `config.yaml`

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/dist/src/main/arctic-bin/conf/config.yaml
+++ b/ams/dist/src/main/arctic-bin/conf/config.yaml
@@ -17,6 +17,10 @@ ams:
   http-server:
     bind-port: 1630
 
+  iceberg:
+    worker:
+      num-threads: 20
+
   refresh-external-catalogs:
     interval: 180000 # 3min
     thread-count: 10
--- a/ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
@@ -55,6 +55,14 @@ public class ArcticManagementConf {
           .defaultValue(10)
           .withDescription("The number of threads used to read metadata.");
 
+  public static final ConfigOption<Integer> ICEBERG_WORKER_NUM_THREADS =
+      ConfigOptions.key("iceberg.worker.num-threads")
+          .intType()
+          .defaultValue(20)
+          .withDescription(
+              "Sets the size of the worker pool. The worker pool limits the number of tasks concurrently processing "
+                  + "manifests in the base table implementation across all concurrent planning or commit operations.");
+
   public static final ConfigOption<Long> REFRESH_EXTERNAL_CATALOGS_INTERVAL =
       ConfigOptions.key("refresh-external-catalogs.interval")
           .longType()
--- a/ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
@@ -48,6 +48,7 @@ import com.netease.arctic.server.utils.ThriftServiceProxy;
 import io.javalin.Javalin;
 import io.javalin.http.HttpCode;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.iceberg.SystemProperties;
 import org.apache.iceberg.relocated.com.google.common.annotations.VisibleForTesting;
 import org.apache.thrift.TMultiplexedProcessor;
 import org.apache.thrift.TProcessor;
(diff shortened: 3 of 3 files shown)


wangtaohz (approved):
LGTM!

ams/dist/src/main/arctic-bin/conf/config.yaml
```diff
@@ -17,6 +17,10 @@ ams:
   http-server:
     bind-port: 1630
 
+  iceberg:
+    worker:
+      num-threads: 20
+
```
baiyangtx, line 23: Should we directly introduce a configuration domain for system properties, instead of designing a separate configuration parameter for this? @zhoujinsong

zhoujinsong: I don't think the goal of this configuration is just to set an iceberg-specific system variable.
It should be a general configuration that applies to all table formats, more like `table-manifest-io.thread-count` introduced by #2116.

Maybe we should reuse the configuration and set the value to the system variable `iceberg.worker.num-threads` for now.

HDYT? @zhongqishang

zhongqishang: Yes, I agree. we can reuse configuration `table-manifest-io.thread-count`, and also reuse `ThreadPools.getWorkerPool()`.

zhoujinsong (approved):
LGTM.

ams/server/src/main/java/com/netease/arctic/server/dashboard/ServerTableDescriptor.java
```diff
@@ -52,13 +50,8 @@ public class ServerTableDescriptor extends PersistentBase {
...
     this.tableService = tableService;
 
-    ExecutorService executorService =
-        Executors.newFixedThreadPool(
-            serviceConfig.getInteger(ArcticManagementConf.TABLE_MANIFEST_IO_THREAD_COUNT),
-            new ThreadFactoryBuilder()
-                .setDaemon(true)
-                .setNameFormat("table-manifest-io-%d")
-                .build());
+    // All table formats will jointly reuse the work thread pool named iceberg-worker-pool-%d
```
shidayang, line 53: The dashboard requires a separate thread pool to prevent user queries from being blocked by background plan tasks.

ams/server/src/main/java/com/netease/arctic/server/ArcticServiceContainer.java
```diff
@@ -473,6 +475,16 @@ private void validateThreadCount(
...
     }
 
+    /** Override the value of {@link SystemProperties}. */
+    private void setIcebergSystemProperties() {
+      int workerThreadPoolSize =
+          Math.max(
+              Runtime.getRuntime().availableProcessors(),
+              serviceConfig.getInteger(ArcticManagementConf.TABLE_MANIFEST_IO_THREAD_COUNT));
+      System.setProperty(
+          SystemProperties.WORKER_THREAD_POOL_SIZE_PROP, String.valueOf(workerThreadPoolSize));
```
shidayang, line 485: I think we should avoid directly setting system variables. There are many different plan scenarios now, and we should differentiate their thread pools. How do you think?@Zhoujinsong

ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
```diff
@@ -52,8 +52,10 @@ public class ArcticManagementConf {
   public static final ConfigOption<Integer> TABLE_MANIFEST_IO_THREAD_COUNT =
       ConfigOptions.key("table-manifest-io.thread-count")
           .intType()
-          .defaultValue(10)
-          .withDescription("The number of threads used to read metadata.");
+          .defaultValue(20)
+          .withDescription(
+              "Sets the size of the worker pool. The worker pool limits the number of tasks concurrently processing "
+                  + "manifests in the base table implementation across all concurrent planning or commit operations.");
```
shidayang, line 58: This is only for dashboard, you should add another params

PR #2411, opened 2023-12-06
 [AMORO-2257] Automatically create tags on snapshots hour-level for iceberg Table Formats

### LINKED ISSUE #2257: [Subtask]: Automatically create tags on snapshots hour-level for iceberg Table Formats
### Search before asking

### Description

Automatically create tags on snapshots hour-level for iceberg Table Formats

### Parent issue

https://github.com/NetEase/amoro/issues/1354

### Code of Conduct

huyuanfeng2018:
Automatically create tags on snapshots hour-level for iceberg Table Formats

Close #2257.

## Brief change log

- Add hour level (HOURLY) in `TagConfiguration.Period`

## How was this patch tested?

## Documentation


wangtaohz:
It seems that we forget to implement this method in `TagsAutoCreatingExecutor`.
```java
  @Override
  public void handleConfigChanged(TableRuntime tableRuntime, TableConfiguration originalConfig) {
    scheduleIfNecessary(tableRuntime, getStartDelay());
  }
```

Can you add this method to this PR? @huyuanfeng2018

huyuanfeng2018:
> It seems that we forget to implement this method in `TagsAutoCreatingExecutor`.
> 
> ```java
>   @Override
>   public void handleConfigChanged(TableRuntime tableRuntime, TableConfiguration originalConfig) {
>     scheduleIfNecessary(tableRuntime, getStartDelay());
>   }
> ```
> 
> Can you add this method to this PR? @huyuanfeng2018

ok

ams/server/src/main/java/com/netease/arctic/server/table/TagConfiguration.java
```diff
@@ -73,6 +95,32 @@ public String propertyName() {
...
+
+    /**
+     * calculate the end of the period adjusted with a specified offset.
+     *
+     * <p>The calculation takes into account the provided period size and offset. It first
+     * calculates an offset timestamp, which is essentially `checkTime` minus offset minutes. Then
+     * it calculates the delta of the division of that timestamp by the period size, then applies
+     * the delta to the original timestamp to get the period start timestamp.
+     */
+    protected Long getPeriodEndWithOffset(
```
wangtaohz, line 109: This implementation is a little complex. I attempted to simplify it using the existing API of LocalDateTime, as shown below.

https://github.com/huyuanfeng2018/amoro/pull/1/files

huyuanfeng2018: > This implementation is a little complex. I attempted to simplify it using the existing API of LocalDateTime, as shown below.
> 
> https://github.com/huyuanfeng2018/amoro/pull/1/files
 
Well, this does make that logic simpler, but it is because our current `Period` only exists DAILY and HOURLY. Just imagine, what if the `Period` here is 2 hours?

wangtaohz: I suggest keeping the current code simple before introducing configurations like `two-hourly`.

Furthermore, I don't think introducing `Period` like `two-hourly`, `four-hourly`, `half-day` is a generic solution. We can discuss  how to support such requirements in the future useless you currently require it.

huyuanfeng2018: > I suggest keeping the current code simple before introducing configurations like `two-hourly`.
> 
> Furthermore, I don't think introducing `Period` like `two-hourly`, `four-hourly`, `half-day` is a generic solution. We can discuss how to support such requirements in the future useless you currently require it.

ok. I adjusted some code to make it more readable

ams/server/src/main/java/com/netease/arctic/server/table/TagConfiguration.java
```diff
@@ -65,14 +87,28 @@ public String propertyName() {
       return propertyName;
     }
 
+    protected abstract Duration onePeriodSize();
```
wangtaohz, line 90: How about changing the method name to `periodDuration()`

ams/server/src/main/java/com/netease/arctic/server/table/TagConfiguration.java
```diff
@@ -65,14 +87,28 @@ public String propertyName() {
...
+     * <p>For example, when creating a daily tag, the offset is set to be 30 min, if the check time
+     * is 2022-08-08 02:00:00, the ideal trigger time is 2022-08-08 00:30:00; if the check time is
+     * 2022-08-09 00:20:00 (before 00:30 of the next day), the ideal trigger time is still
+     * 2022-08-08 00:30:00.
      */
-    public abstract long getTagTriggerTime(LocalDateTime checkTime, int triggerOffsetMinutes);
+    public abstract LocalDateTime getTagTriggerTime(
+        LocalDateTime checkTime, int triggerOffsetMinutes);
+
+    public LocalDateTime normalizeToTagTime(long triggerTime, int triggerOffsetMinutes) {
```
wangtaohz, line 107: I suggest changing the type of triggerTime to `LocalDateTime`, like 
```
public LocalDateTime normalizeToTagTime(LocalDateTime triggerTime, int triggerOffsetMinutes) 
```.

File: core/src/main/java/com/netease/arctic/table/TableProperties.java [hunk 4]
```diff
@@ -213,7 +213,8 @@ private TableProperties() {}
   public static final int AUTO_CREATE_TAG_MAX_DELAY_MINUTES_DEFAULT = 60;
 
   public static final String AUTO_CREATE_TAG_DAILY_FORMAT = "tag.auto-create.daily.tag-format";
```
wangtaohz, line 215: How about add a new hourly format property?
```java
public static final String AUTO_CREATE_TAG_HOURLY_FORMAT = "tag.auto-create.hourly.tag-format";
public static final String AUTO_CREATE_TAG_HOURLY_FORMAT_DEFAULT = "'tag-'yyyyMMddHH";
```

wangtaohz: An alternative solution is to remove "daily" from the property name, then it will be 
```java
public static final String AUTO_CREATE_TAG_FORMAT = "tag.auto-create.tag-format";
public static final String AUTO_CREATE_TAG_FORMAT_DAILY_DEFAULT = "'tag-'yyyyMMdd";
public static final String AUTO_CREATE_TAG_FORMAT_HOURLY_DEFAULT = "'tag-'yyyyMMddHH";
```
I think both are acceptable, which one do you prefer?

wangtaohz (commented):
@huyuanfeng2018 I left some comments, please take a look.

core/src/main/java/com/netease/arctic/table/TableProperties.java
huyuanfeng2018: > An alternative solution is to remove "daily" from the property name, then it will be
> 
> ```java
> public static final String AUTO_CREATE_TAG_FORMAT = "tag.auto-create.tag-format";
> public static final String AUTO_CREATE_TAG_FORMAT_DAILY_DEFAULT = "'tag-'yyyyMMdd";
> public static final String AUTO_CREATE_TAG_FORMAT_HOURLY_DEFAULT = "'tag-'yyyyMMddHH";
> ```
> 
> I think both are acceptable, which one do you prefer?

There may be a problem here. For example, `hourly` is configured but the user passes 'tag-'yyyyMMdd.
 Should the parameters tag.auto-create.tag-format-prefix and tag.auto-create.tag-format-suffix be used directly here?  WDYT?

wangtaohz: Users may also have costom needs in terms of time format(like `yyyy-MM-dd`).
Another case, when using `hourly` tag with a  offset like 30min, it is also reasonable to set the format accurate to the minute, like `tag-20230102-12:30`.
So only `prefix` and `suffix` can't meet the above requirements, and 3 parts are needed for this format configuration (`prefix`,`time-format`,`suffix`), which is too complicated.

wangtaohz: > There may be a problem here. For example, hourly is configured but the user passes 'tag-'yyyyMMdd.

This is a situation of configuration error, which may lead to abnormal situations, such as the inability to create tags.  We cannot prevent user's misconfigurations.

huyuanfeng2018: > > There may be a problem here. For example, hourly is configured but the user passes 'tag-'yyyyMMdd.
> 
> This is a situation of configuration error, which may lead to abnormal situations, such as the inability to create tags. We cannot prevent user's misconfigurations, and I believe this falls outside the scope of the issues we should address.

ok, Regarding the two configuration methods, I prefer to use the second one. The configuration will be a less compared to the first one.

huyuanfeng2018:
> @huyuanfeng2018 I left some comments, please take a look.

done.

ams/server/src/main/java/com/netease/arctic/server/table/TagConfiguration.java
```diff
@@ -65,14 +85,26 @@ public String propertyName() {
...
+     * <p>For example, when creating a daily tag, the offset is set to be 30 min, if the check time
+     * is 2022-08-08 02:00:00, the ideal trigger time is 2022-08-08 00:30:00; if the check time is
+     * 2022-08-09 00:20:00 (before 00:30 of the next day), the ideal trigger time is still
+     * 2022-08-08 00:30:00.
      */
-    public abstract long getTagTriggerTime(LocalDateTime checkTime, int triggerOffsetMinutes);
+    public abstract LocalDateTime getTagTriggerTime(
+        LocalDateTime checkTime, int triggerOffsetMinutes);
+
+    public LocalDateTime normalizeToTagTime(LocalDateTime triggerTime, int triggerOffsetMinutes) {
```
wangtaohz, line 105: We had better add comments to this public method so that other developers can better understand it.

wangtaohz (commented):
Thanks for you contribution! @huyuanfeng2018 

The code looks fine to me, only with some details needs to be modified.

wangtaohz (approved):
LGTM!

ams/server/src/main/java/com/netease/arctic/server/optimizing/maintainer/AutoCreateIcebergTagAction.java
```diff
@@ -109,22 +104,27 @@ private boolean exceedMaxDelay(Snapshot snapshot) {
...
 
   private String generateTagName() {
-    if (tagConfig.getTriggerPeriod() == TagConfiguration.Period.DAILY) {
-      String tagFormat = tagConfig.getTagFormat();
-      return now.minusDays(1).format(DateTimeFormatter.ofPattern(tagFormat));
-    } else {
-      throw new IllegalArgumentException(
-          "unsupported trigger period " + tagConfig.getTriggerPeriod());
-    }
+    LocalDateTime tagTime =
```
zhoujinsong, line 112: Hi, when I reviewed the code here, these three time-related variables made it difficult for me to understand. After carefully reviewing the code again, I have summarized the names and meanings of the three parameters:

- Check time: the timestamp when the action is started.
- Trigger time: the timestamp of the last tag should be triggered.
- Tag time: the timestamp of the last tag represents.

And we calculate the 3 variables with the methods:

- checkTime = now()
- triggerTime = (checkTime - triggerOffset) % period + triggerOffest
- TagTime = triggerTime - triggerOffest - period

then,
-  TagTime = (checkTime - triggerOffset) % period - period
- TriggerTime = TagTime + period + triggerOffset

And I think we can make the code much easier to understand by:

- Put the 3 variables together and try to calculate them at the beginning, put some comments to tell the meaning and the calculating rules.
- Calculate the tag time first and then calculate the trigger time based on tag time.

Of course, the second point is just my personal suggestion, as this calculation process is easier for me to understand.

zhoujinsong (commented):
@huyuanfeng2018 @wangtaohz  
I left some comments, please take another look.

docs/user-guides/configurations.md
```diff
@@ -70,13 +70,13 @@ Data-cleaning configurations are applicable to both Iceberg Format and Mixed str
...
+| Key                                        | Default                                          | Description                                                 |
+|--------------------------------------------|--------------------------------------------------|-------------------------------------------------------------|
+| tag.auto-create.enabled                    | false                                            | Enables automatically creating tags                         |
+| tag.auto-create.trigger.period             | daily                                            | Period of creating tags, support `daily`,`hourly` now       |
+| tag.auto-create.trigger.offset.minutes     | 0                                                | The minutes by which the tag is created after midnight (00:00) |
+| tag.auto-create.trigger.max-delay.minutes  | 60                                               | The maximum delay time for creating a tag                   |
+| tag.auto-create.tag-format            | 'tag-'yyyyMMdd for daily and 'tag-'yyyyMMddHH for hourly periods | The format of the <br/>name for daily tag  <br>             |
```
zhoujinsong, line 79:
```suggestion
| tag.auto-create.tag-format            | 'tag-'yyyyMMdd for daily and 'tag-'yyyyMMddHH for hourly periods | The format of the name for tag              |
```

zhoujinsong (approved):
LGTM.
Thanks a lot for your great work!

PR #2421, opened 2023-12-09
[AMORO-1812] Support spark-based external optimizer

### LINKED ISSUE #1812: [Improvement]: Have you considered implementing a spark-based optimizer?
### Search before asking

Have you considered implementing a spark-based optimizer? Now there are only optimized implementations based on local and flink. Have you considered the implementation of spark? especially external.
I think spark is still the main computing engine in some companies or teams.

### How should we improve?

_No response_

### Subtasks

_No response_

### Code of Conduct

tcodehuber:

Close #1812 .

## Brief change log

-

## How was this patch tested?

## Documentation


tcodehuber:
@wangtaohz @baiyangtx Can you give me some suggestions about this PR, thx?

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkExecutor.java
```diff
@@ -0,0 +1,128 @@
...
+          try {
+            String threadName = Thread.currentThread().getName();
+            long startTime = System.currentTimeMillis();
+            LOG.info("Now [{}] execute task {}", threadName, task);
+            ImmutableList<OptimizingTask> of = ImmutableList.of(task);
+            String jobDesc =
+                String.format("%s | TableName(%s)", task.getTaskId(), parseFullTableName(task));
+            jsc.setJobDescription(jobDesc);
+            List<OptimizingTaskResult> result = jsc.parallelize(of, 1).map(sparkFunction).collect();
+            result.forEach(SparkExecutor.this::completeTask);
```
XBaith, line 83: Why perform optimization tasks via thread pool? Seems like a simple matter of wrapping the execution into Spark rather than utilizing Spark's distributed processing capabilities?

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizer.java
```diff
@@ -0,0 +1,161 @@
...
+        ThreadPoolExecutor executorService = sparkExecutor.getExecutorService();
+        LOG.info("Starting to poll optimizing tasks.");
+        while (true) {
+          int activeCount = executorService.getActiveCount();
+          if (activeCount >= executorService.getMaximumPoolSize()) {
+            LOG.info("Polled {} tasks, the queue is full", activeCount);
+            waitAShortTime();
+          } else {
+            OptimizingTask task = pollTask();
+            if (task != null && ackTask(task)) {
```
XBaith, line 107: If I understand correctly, the goal of the Spark optimizer is to leverage batch processing. So we may need to fetch a batch of tasks at each optimization?
WDYT @shidayang , @zhoujinsong

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizer.java
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.thrift.TException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+
+public class SparkOptimizer {
```
baiyangtx, line 39: All class must have class docs.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkExecutor.java
```diff
@@ -0,0 +1,128 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class SparkExecutor extends AbstractOptimizerOperator {
```
baiyangtx, line 40: `SparkExecutor` is a misleading name, easily leading one to think of the executor process in the Spark architecture. 

Based on the code logic, it is suggested to change it to `SparkOptimizingTaskSubmitter`.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizer.java
baiyangtx, line 107: I think we may not need the "taskPuller" thread. Instead, we can create a thread pool equal to the size of "executionParallel" right from the start, with each thread being a `SparkOptimizerTaskSubmitter`  

Within each `SparkOptimizerTaskSubmitter`  we can complete the entire logic of "pollTask" -> "ackTask" -> "jsc.parallelize.map.collect" -> "completeTask." 

This way, when calling "ackTask," we can return the correct "threadId" to the AMS.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkFunction.java
```diff
@@ -0,0 +1,86 @@
...
+import com.netease.arctic.utils.ExceptionUtil;
+import com.netease.arctic.utils.SerializationUtil;
+import org.apache.iceberg.common.DynConstructors;
+import org.apache.spark.api.java.function.Function;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.ByteBuffer;
+
+public class SparkFunction implements Function<OptimizingTask, OptimizingTaskResult> {
```
baiyangtx, line 36: Class docs is required, and I suggest naming the class "SparkOptimizingTaskExecuteFunction," as it is clearer.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizerConfig.java
```diff
@@ -0,0 +1,64 @@
...
+    super(args);
+  }
+
+  public long getPollingTimeout() {
+    return pollingTimeout;
+  }
+
+  public void setPollingTimeout(long pollingTimeout) {
+    this.pollingTimeout = pollingTimeout;
+  }
```
baiyangtx, line 48: Why do we need to set a pollingTimeout?

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizer.java
```diff
@@ -0,0 +1,161 @@
...
+          LOG.info("Optimizer executor[{}] polled task[{}] from ams", threadId, task.getTaskId());
+          break;
+        } else {
+          waitAShortTime();
+        }
+        if (System.currentTimeMillis() - startTime > config.getPollingTimeout()
+            && sparkExecutor.getExecutorService().getActiveCount() == 0) {
+          LOG.warn("Waited for {} s, the optimizer will exit", config.getPollingTimeout());
+          System.exit(0);
+        }
```
baiyangtx, line 156: Why do we want the driver to exit after a pollingTimeout in pollTask?

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkExecutor.java
tcodehuber: Yes, we just reuse the impl of OptimizingExecutor#execute in core module. But by setting spark.dynamicAllocation.enabled and parallel of the optimizer, the compute resource can be fully used and is more flexible.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizerConfig.java
tcodehuber: Since we want schedule the spark optimizer at a fixed time interval, such as one hour(data changes are not very frequent). If during the pollingTime, we don't get OptimizingTask and there is no OptimizingTask in running state , it's a suitable time to release the optimizer.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizer.java
tcodehuber: Sounds good, I will follow the suggestion, thx.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkExecutor.java
XBaith: Okay, so SparkExecutor just submit the tasks to spark application. That's make sense

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizer.java
baiyangtx: I don't agree. I believe that compared to the Flink Optimizer, the Spark Optimizer is more flexible in actual resource utilization. It can adjust the number of executors based on the actual load of the Optimizing-Group. 

When the spark executer is idle, it can automatically reduce the number of executor nodes. In comparison, under the Flink Optimizer, concurrency strictly equals resource usage, while under the Spark Optimizer, concurrency represents the upper limit of resource usage.

If we adopt the approach of acquiring a batch of tasks before execution, we cannot guarantee the timeliness of optimization on the table. 

In the current design, plans are triggered based on statistics of small files on the table, which means that the tasks generated by the plan must be executed as soon as possible.

tcodehuber:
@baiyangtx Can you review it again?

baiyangtx (commented):
Recently, I discussed this PR with @zhoujinsong  and overall agree with its implementation approach. However, I hope we can reuse the code in the `Optimizer/Common` package as much as possible. The refactoring should follow the logic below:

1. Have SparkOptimizer extend the `common.Optimizer` class to reuse its `OptimizerToucher` and `OptimizerExecutor` management capabilities.

2. Modify the part of the Optimizer constructor that initializes the executor array to use a `protected newOptimizerExecutor() ` method instead of directly creating a new `OptimizerExecutor` object.

3. Implement the SparkOptimizerExecutor class under the Spark module, which extends the `common.OptimizerExecutor `class type and overrides the `private OptimizingTaskResult executeTask(OptimizingTask task) method`.

4. In the overridden `OptimizingTaskResult executeTask(OptimizingTask task) ` method of `SparkOptimizerExecutor`, implement the logic currently found in `SparkOptimizingTaskSubmitter.submitOptimizingTask.`

In this way, we can reuse the code in the `Optimizer/Common` module as much as possible, and in the future, we can also implement an optimizer based on Flink batch in the Flink session using this framework.

tcodehuber:
> Recently, I discussed this PR with @zhoujinsong and overall agree with its implementation approach. However, I hope we can reuse the code in the `Optimizer/Common` package as much as possible. The refactoring should follow the logic below:
> 
> 1. Have SparkOptimizer extend the `common.Optimizer` class to reuse its `OptimizerToucher` and `OptimizerExecutor` management capabilities.
> 2. Modify the part of the Optimizer constructor that initializes the executor array to use a `protected newOptimizerExecutor() ` method instead of directly creating a new `OptimizerExecutor` object.
> 3. Implement the SparkOptimizerExecutor class under the Spark module, which extends the `common.OptimizerExecutor `class type and overrides the `private OptimizingTaskResult executeTask(OptimizingTask task) method`.
> 4. In the overridden `OptimizingTaskResult executeTask(OptimizingTask task) ` method of `SparkOptimizerExecutor`, implement the logic currently found in `SparkOptimizingTaskSubmitter.submitOptimizingTask.`
> 
> In this way, we can reuse the code in the `Optimizer/Common` module as much as possible, and in the future, we can also implement an optimizer based on Flink batch in the Flink session using this framework.

OK, I will try to work on it later.

ams/dashboard/src/main/resources/static/index.html
```diff
@@ -23,8 +23,8 @@
     <link rel="icon" type="image/svg+xml" href="./favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Amoro</title>
-    <script type="module" crossorigin src="./assets/index-Cv3Z2rcY.js"></script>
-    <link rel="stylesheet" crossorigin href="./assets/index-4N-LbzM3.css">
+    <script type="module" crossorigin src="./assets/index-_J0gz4e7.js"></script>
+    <link rel="stylesheet" crossorigin href="./assets/index-CsgCj9G5.css">
```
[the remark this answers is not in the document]
tcodehuber: @baiyangtx I have refactored some code according to your suggestions. Can you review it again, thx?

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizer.java
```diff
@@ -0,0 +1,92 @@
...
+              Arrays.stream(optimizer.getExecutors()).filter(e -> e.isStarted()).count();
+          if (runningCnt == 0) {
+            LOG.info(
+                "Waited for {} s, the spark optimizer will exit", config.getTaskPollingTimeout());
+            System.exit(0);
+          }
+        },
+        0,
+        1,
+        TimeUnit.MINUTES);
```
baiyangtx, line 87: I think there is no need to make the Spark driver exit actively.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizerConfig.java
```diff
@@ -0,0 +1,69 @@
...
+      "task-polling-timeout"; // the time interval since the last time we polled an optimizing task
+
+  @Option(
+      name = "-tpt",
+      aliases = "--" + AMS_TASK_POLLING_TIMEOUT,
+      usage = "The task polling timeout, default 10min",
+      required = true)
+  private long taskPollingTimeout = 600000; // 10min
+
+  public SparkOptimizerConfig() {}
```
baiyangtx, line 41: I don't think it's necessary to make Spark Driver actively exit, is this parameter necessary?

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizerExecutor.java
```diff
@@ -0,0 +1,122 @@
...
+      if (System.currentTimeMillis() - startTime > sparkOptimizerConfig.getTaskPollingTimeout()) {
+        LOG.warn(
+            "Waited for {} s, the SparkOptimizerExecutor-{} will stop",
+            sparkOptimizerConfig.getTaskPollingTimeout(),
+            threadId);
+        stop();
+      }
+    }
+    return task;
+  }
```
baiyangtx, line 109: "There should be no need to override the methods of the parent class. There is no need to actively exit the Spark driver.

baiyangtx (commented):
There is no need to develop separate logic for the active exit of the Spark Optimizer. Therefore, the related code in this area should be unnecessary. I have already commented it out, and the other parts should be fine.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizer.java
tcodehuber: Do you mean that we should let the driver continue to run without exiting?

zhoujinsong: I agreed with keeping the Optimizer even when there are no tasks. In this scenario, the Spark Optimizer will only consume resources on the driver side.

ams/optimizer/spark-optimizer/src/main/java/com/netease/arctic/optimizer/spark/SparkOptimizerExecutor.java
```diff
@@ -0,0 +1,92 @@
...
+
+  private String jobDescription(OptimizingTask task) {
+    String description;
+    TableOptimizing.OptimizingInput input =
+        SerializationUtil.simpleDeserialize(task.getTaskInput());
+    if (input instanceof RewriteFilesInput) {
+      description =
+          String.format(
+              "Amoro rewrite files task, table name:%s, task id:%s",
+              task.getTaskId(), ((RewriteFilesInput) input).getTable().name());
```
zhoujinsong, line 86: We should exchange the task ID and table name, sorry for my mistake.

zhoujinsong (commented):
@tcodehuber All codes look fine to me expect a small issue should be fixed.

zhoujinsong (approved):
LGTM.
Validated in my local environment.

PR #2446, opened 2023-12-18
[AMORO-2438] Mixed Format KeyedTable support bin-packing splitting optimizing tasks

### LINKED ISSUE #2438: [Improvement]: Mixed Format KeyedTable Self-Optimizing supports bin-packing tasks after splitting tasks by nodes
### Search before asking

The Mixed Format KeyedTable splits tasks based on nodes (hash). If the number of split tasks is insufficient, it will lead to inadequate concurrent utilization by the optimizer, resulting in a longer process execution time.
image

### How should we improve?

We should use bin-packing to split the files in one node into several tasks when it's necessary.

### Subtasks

_No response_

### Code of Conduct

wangtaohz:

Close #2438 .

## Brief change log

- fix cost calculation when optimizing planning
- split tasks by bin-packing in a tree node

## How was this patch tested?

## Documentation


ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/OptimizingPlanner.java
```diff
@@ -146,14 +147,14 @@ public List<TaskDescriptor> planTasks() {
     double avgThreadCost = actualInputSize / availableCore;
     List<TaskDescriptor> tasks = Lists.newArrayList();
     for (AbstractPartitionPlan partitionPlan : actualPartitionPlans) {
-      tasks.addAll(partitionPlan.splitTasks((int) (actualInputSize / avgThreadCost)));
+      tasks.addAll(partitionPlan.splitTasks((int) (partitionPlan.getCost() / avgThreadCost)));
```
wangtaohz, line 150: Fix calculating the target task count for each partition.

PR #2486, opened 2023-12-29
[HOTFIX] Update metric-reporters.yaml

tcodehuber:

Commenting out this part of the configuration will cause an error when starting AMS：
`throw new IllegalArgumentException("map is null.");`

## Brief change log

-

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/dist/src/main/arctic-bin/conf/plugins/metric-reporters.yaml
+++ b/ams/dist/src/main/arctic-bin/conf/plugins/metric-reporters.yaml
@@ -18,6 +18,6 @@
 # This file is config file for plugin metric reporter
 # configurations of metric reporters
 
-#metric-reporters:
-#  - name: prometheus-exporter
-#    enabled: false
+metric-reporters:
+  - name: prometheus-exporter
+    enabled: false


baiyangtx (changes requested):
I suggest modifying the code to be compatible with situations where there is no plugin config.yaml or where the file is empty because the plugin is not necessarily needed.

tcodehuber:
> I suggest modifying the code to be compatible with situations where there is no plugin config.yaml or where the file is empty because the plugin is not necessarily needed.

OK, I will try it

zhoujinsong:
> OK, I will try it

@tcodehuber Thank you very much for raising and trying to fix this issue. I also agree that making the code adaptable to the absence of this configuration is a better solution. Could you please update me on the current progress of the solution? Also, feel free to reach out to me if there are any implementation issues.

tcodehuber:
@zhoujinsong 
I will commit the code this weekend.

tcodehuber:
> > OK, I will try it
> 
> @tcodehuber Thank you very much for raising and trying to fix this issue. I also agree that making the code adaptable to the absence of this configuration is a better solution. Could you please update me on the current progress of the solution? Also, feel free to reach out to me if there are any implementation issues.

https://github.com/NetEase/amoro/pull/2512

PR #2492, opened 2023-12-29
[Feature]: Make local terminal cores configurable

link3280:

As titled.

## Brief change log

As titled.

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
@@ -313,6 +313,12 @@ public class ArcticManagementConf {
           .defaultValue(30)
           .withDescription("session timeout in minute");
 
+  public static final ConfigOption<Integer> TERMINAL_LOCAL_CORES =
+    ConfigOptions.key("terminal.local.cores")
+      .intType()
+      .defaultValue(1)
+      .withDescription("cores for local spark terminal");
+
   /** configs of data expiration */
   public static final ConfigOption<Boolean> DATA_EXPIRATION_ENABLED =
       ConfigOptions.key("data-expiration.enabled")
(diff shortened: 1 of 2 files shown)


ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
```diff
@@ -313,6 +313,12 @@ public class ArcticManagementConf {
           .defaultValue(30)
           .withDescription("session timeout in minute");
 
+  public static final ConfigOption<Integer> TERMINAL_LOCAL_CORES =
```
zhoujinsong, line 316: It seems to be a configuration only for the local terminal.
If yes, we may want to move it into `LocalSessionFactory` and name it `cores` because the `TerminalManager` will reorganize configurations based on different backend types.

You can find similar configurations only for the Kyuubi terminal in `KyuubiTerminalSessionFactory`.

zhoujinsong (commented):
@link3280 I left some comments, please take a look.

Besides, I found there are no docs about the configurations of terminals, it will make users hard to get these configurations.

Maybe we should add some description in docs and examples in default configuration files, how do you think? @link3280

ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
link3280: I agree that `LocalSessionFactory` is a better place for the config option. But do you think `cores` should be prefixed with the terminal type `local`, as it's only effective for the local terminal. That makes the full key `terminal.local.cores`.

zhoujinsong: Sorry, my previous description was not clear. The complete configuration name in the configuration file should indeed be `terminal.local.cores`. Just as you have implemented, in the code, we only need to name the configuration `cores`, and `TerminalManager` will handle the decomposition of different backend configurations.

zhoujinsong (approved):
LGTM.

zhoujinsong:
I created #2496 to follow the docs issues.

PR #2504, opened 2024-01-12
[AMORO-2149]: Add filter select by optimizeGroup for Optimize Table

### LINKED ISSUE #2149: [Subtask]: Support filter for `Optimizer Group` column of Optimizing Tables page
### Search before asking

### Description

Support filter for this `Optimizer Group` column of Optimizing Tables page would bring convenience to users who use multiple optimizer groups.

image

### Parent issue

#885

### Code of Conduct

zhangmo8:

Add filter select by optimizeGroup for Optimize Table

image

Close #2149.

## Brief change log

-

## How was this patch tested?

## Documentation


songjianet (approved):

zhoujinsong:
@zhangmo8 Thanks a lot for your work!

However, considering that we will be adding more filters on additional fields in the table in the future, would it be more appropriate to add a filter icon in the table header for filtering operations?

Of course, this requires your assistance in assessing its feasibility for implementation.
I understand that adding sorting, filtering, and searching on specific table fields may be a potential requirement in the future.

zhangmo8:
> @zhangmo8 Thanks a lot for your work!
> 
> However, considering that we will be adding more filters on additional fields in the table in the future, would it be more appropriate to add a filter icon in the table header for filtering operations?
> 
> Of course, this requires your assistance in assessing its feasibility for implementation. I understand that adding sorting, filtering, and searching on specific table fields may be a potential requirement in the future.

It is true that adding filter items to the table header is actually easier for building the front end, but I personally think that using a selection box outside is more in line with the existing operation interaction. @zhoujinsong

zhoujinsong:
Yes, I have also noticed that some pages now have additional selectors to filter page content, such as selecting partitions on the `Files` page in the table details and selecting operations on the `Snapshots` page.

Perhaps we should establish a unified interaction method and strive for consistency in both new and old implementations. In my opinion, I still advocate for adding filters in the table headers, as this would allow us to add filters to more columns. Having separate filters can appear cluttered when there is a need to filter a lot of content.

zhoujinsong:
After discussing offline with @zhangmo8 @minteliuwm , we reached an agreement:

- Place the filtering and searching component above the table rather than in the header, making it easier for users to see the current filtering status.
- Place the sorting component directly in the table header.

All similar interaction requirements in the future should follow this guideline.

zhoujinsong (approved):
I tested it in my local environment and it works fine.

PR #2514, opened 2024-01-23
[AMORO-2509] Rollback the trigger condition of minor compaction

### LINKED ISSUE #2509: [Improvement]: Improve minor compaction trigger about eq delete file.
### Search before asking

In PR #2332 , I added a new condition (rewritePosSegmentFileCount > 0) for EQ DELETE rewrite to POS DELETE. Although it more accurately determines minor-level optimization, it causes an eq-delete of any size to trigger the rewrite pos operation of all data files , taking up a lot of resources. 
I guess this condition was not added at the beginning to reduce excessive usage of optimizer resources.

### How should we improve?

1. Remove the condition (rewritePosSegmentFileCount > 0)
2. Add new logic for EQ-DELETE same with POS-DELETE, Add new config.

### Subtasks

_No response_

### Code of Conduct

zhongqishang:

close #2509 improve 1

## Brief change log

- Remove the condition (rewritePosSegmentFileCount > 0)

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/plan/CommonPartitionEvaluator.java
@@ -317,8 +317,7 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
   public boolean isMinorNecessary() {
     int smallFileCount = fragmentFileCount + equalityDeleteFileCount;
     return smallFileCount >= config.getMinorLeastFileCount()
-        || (smallFileCount > 1 && reachMinorInterval())
-        || rewritePosSegmentFileCount > 0;
+        || (smallFileCount > 1 && reachMinorInterval());
   }
 
   protected boolean reachMinorInterval() {


zhoujinsong:
@zhongqishang 
I noticed that you added an implementation plan: `Add new logic for EQ-DELETE same with POS-DELETE, Add new config.` in #2509.

I would like to have a discussion about that.
If the insert file is associated with any equality delete file, it will trigger minor optimizing as equality delete files are now uniformly considered as small files.
 If the insert file is only associated with one position delete file, major optimizing will be triggered based on the size of the position delete file.
 In the case where the insert file is associated with multiple position delete files, although they are determined to require rewriting the position delete file, the triggering conditions for this scenario are not covered.

I assumed that what you are describing is this scenario. If yes, I agree that new triggering conditions should be added for this scenario, and it should be considered as minor optimizing. However, using only `rewritePosSegmentFileCount > 0` as a condition at this stage is indeed inappropriate, as it also covers cases with any equality delete files.

zhoujinsong (approved):
We can merge this PR and add new condition for minor optimizing in another one.

PR #2526, opened 2024-02-04
[AMORO-2412] Automatically delete expired iceberg Table tag

### LINKED ISSUE #2412: [Subtask]: Automatically delete expired iceberg Table tag and branch
### Search before asking

### Description

Automatically expire the tags of the iceberg table

### Parent issue

https://github.com/NetEase/amoro/issues/1354

### Code of Conduct

huyuanfeng2018:

Close #2412 .

## Brief change log

- New parameters are added to set the automatically created Tag expiration configuration.

## How was this patch tested?

## Documentation


huyuanfeng2018:
###  Some questions about tag configuration

####   when will detection be performed to delete these expired snapshots?
1.  if tag expiration is enabled, it is similar to other modules in AMS, such as clean orphan files implementation. Periodically check and delete tag.
2.  configure the tag parameter `max-ref-age- ms` when creating a tag, which will expire through the expire snapshot action of iceberg.
- If we use the first, we need to add a periodic detection module, the second can directly reuse the function of expire snapshot, but the second method can not support the use max keep num parameter, that is, the number of automatically created snapshots cannot be set. But in my opinion, setting the retention time is good enough, so I think it's good for us to stay close to the community and only support the second way.


####  What actions do we need to take if the tag configuration changes?
1. If the tag format changes, do I need to update these old tags and rename them to the new tag format?
I think that this kind of configuration change is not very frequent. If there is such a change, the user can manually delete it through amoro's visual method. I think it can be acceptable. 

2. If the configured expiration time changes, do we need to modify the expiration time configuration of these tags?
For the same reason, I don’t think it’s necessary

@zhoujinsong @wangtaohz  cc

zhoujinsong:
> When will detection be performed to delete these expired snapshots?

I also agree with directly using the tag expiration feature provided by Iceberg. This approach maintains consistency with community practices and, in my opinion, already meets the needs of users.

> What actions do we need to take if the tag configuration changes?

I think it's acceptable that configuration changes only take effect for subsequent tags. We may need to clarify this feature in the documentation.

zhoujinsong (commented):
@huyuanfeng2018 Thanks a lot for your contribution!
I have a suggestion about the new configuration name, please take another look!

ams/server/src/main/java/com/netease/arctic/server/table/TagConfiguration.java
```diff
@@ -211,6 +232,7 @@ public String toString() {
         .add("triggerPeriod", triggerPeriod)
         .add("triggerOffsetMinutes", triggerOffsetMinutes)
         .add("maxDelayMinutes", maxDelayMinutes)
+        .add("expireKeepMinutes", tagMaxAgeMs)
```
zhoujinsong, line 235:
```suggestion
        .add("tagMaxAgeMs", tagMaxAgeMs)
```

ams/server/src/test/java/com/netease/arctic/server/optimizing/maintainer/TestAutoCreateIcebergTagAction.java
```diff
@@ -356,4 +388,17 @@ private void checkTag(Table table, String tagName, Snapshot snapshot) {
   private void checkSnapshots(Table table, int count) {
     Assert.assertEquals(Iterables.size(table.snapshots()), count);
   }
+
+  private long waitUntilAfter(long timestampMillis) {
+    long current = System.currentTimeMillis();
+    while (current <= timestampMillis) {
+      current = System.currentTimeMillis();
```
zhoujinsong, line 395: Sleeping for a period of time rather than looping indefinitely seems better.

huyuanfeng2018: My code is copy from `com.netease.arctic.server.optimizing.maintainer.TestSnapshotExpire`, and the while loop is also used directly in iceberg tester . But I don’t understand why you can’t just use sleep directly.

ams/server/src/main/java/com/netease/arctic/server/table/TagConfiguration.java
huyuanfeng2018: fixed

ams/server/src/test/java/com/netease/arctic/server/optimizing/maintainer/TestAutoCreateIcebergTagAction.java
huyuanfeng2018: I guess it may be because iceberg directly uses system.currentTimeMillis as the value to compare with snapshot when it expires. Will `Sleeping` as a native method cause some problems?

zhoujinsong (approved):
LGTM.

ams/server/src/test/java/com/netease/arctic/server/optimizing/maintainer/TestAutoCreateIcebergTagAction.java
zhoujinsong: got it.

PR #2543, opened 2024-02-06
[AMORO-2542] Fix Using `transform` partition and the partition value is `null`, the problem of `optimizing` cannot be triggered

### LINKED ISSUE #2542: [Bug]: partition values is `null` ​​will not trigger `Optimizing`
### What happened?

image
For example, if I set the partition mode to `transform` and my data value is null, `Optimizing` will not be triggered at this time.

### Affects Versions

master

AMS

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

huyuanfeng2018:

Close #2542 .

## Brief change log

- Add detection to avoid throwing exceptions in `org.apache.iceberg.expressions`

## How was this patch tested?

## Documentation

### DIFF
--- a/core/src/main/java/com/netease/arctic/utils/ExpressionUtil.java
+++ b/core/src/main/java/com/netease/arctic/utils/ExpressionUtil.java
@@ -71,7 +71,11 @@ public class ExpressionUtil {
       Class<?> resultType =
           partitionField.transform().getResultType(sourceField.type()).typeId().javaClass();
       Object partitionValue = partition.get(i, resultType);
-      filter = Expressions.and(filter, Expressions.equal(transform, partitionValue));
+      if (partitionValue != null) {
+        filter = Expressions.and(filter, Expressions.equal(transform, partitionValue));
+      } else {
+        filter = Expressions.and(filter, Expressions.isNull(transform));
+      }
     }
     return filter;
   }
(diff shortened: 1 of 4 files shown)


HuangFru (approved):
LGTM.

huyuanfeng2018:
> @huyuanfeng2018 The codes look good to me, do you think we should add some unit test case about it?

Of course, but there is a problem here is that there is no `optional` column defined in the `BasicTableTestHelper.TABLE_SCHEMA` definition, so if I want to modify this scenario, I need to define an optional column, which involves many other test cases. Do you think it’s appropriate to add these to this PR?

XBaith:
> Of course, but there is a problem here is that there is no `optional` column defined in the `BasicTableTestHelper.TABLE_SCHEMA`

Maybe you can define a new table schema in `TestUnkeyedExpressionUtil` and `TestKeyedExpressionUtil`?

huyuanfeng2018:
> @huyuanfeng2018 The codes look good to me, do you think we should add some unit test case about it?

done.

huyuanfeng2018:
> > Of course, but there is a problem here is that there is no `optional` column defined in the `BasicTableTestHelper.TABLE_SCHEMA`
> 
> Maybe you can define a new table schema in `TestUnkeyedExpressionUtil` and `TestKeyedExpressionUtil`?

done.

XBaith (approved):
LGTM

PR #2566, opened 2024-02-24
[AMORO-2565] Optimization task loss leads to the optimization process remaining in a running state indefinitely

### LINKED ISSUE #2565: [Bug]: Optimization task loss leads to the optimization process remaining in a running state indefinitely
### What happened?

Task-14 does not exist in the `taskQueue` of AMS, so this task remains in the planned state.
uTools_1708763128753

The root cause of this bug is that even if `retryTaskQueue.poll()` returns a task,`fetchScheduledTask()` will still be called as usual. Once `fetchScheduledTask()` also fetches a task, this task will be lost. 
uTools_1708764475759

### Affects Versions

master

AMS

### Relevant log output

uTools_1708765703735

### Anything else

_No response_

### Code of Conduct

rfyu:

Close #2565.

## Brief change log

-

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/OptimizingQueue.java
@@ -201,9 +201,9 @@ public class OptimizingQueue extends PersistentBase {
 
   private TaskRuntime fetchScheduledTask() {
     return tableQueue.stream()
+        .findFirst()
         .map(TableOptimizingProcess::poll)
         .filter(Objects::nonNull)
-        .findFirst()
         .orElse(null);
   }
 


huyuanfeng2018:
Can add some test cases?

rfyu:
> Can add some test cases?

This change is relatively simple, but I don't know how to write concurrent tests. This scenario only occurs when polling task from multiple tables in the same `OptimizingQueue` simultaneously.

huyuanfeng2018:
> > Can add some test cases?
> 
> This change is relatively simple, but I don't know how to write concurrent tests. This scenario only occurs when polling task from multiple tables in the same `OptimizingQueue` simultaneously.

I couldn't reproduce this bug  locally. The main reason is that after this modification, If the `clearProcess` method is not called, it seems that we cannot make the tasks of other tables enter the optmizer.

rfyu:
> > > Can add some test cases?
> > 
> > 
> > This change is relatively simple, but I don't know how to write concurrent tests. This scenario only occurs when polling task from multiple tables in the same `OptimizingQueue` simultaneously.
> 
> I couldn't reproduce this bug locally. The main reason is that after this modification, If the `clearProcess` method is not called, it seems that we cannot make the tasks of other tables enter the optmizer.

You are right, my modification is wrong.

PR #2582, opened 2024-02-28
[Hotfix] Dependent on log4j2 related packages in compile scope in the `ams-server` module

zhoujinsong:

#2578 add test scope dependencies for log4j2 related packages in the root module, but the `ams-server` module relies on them in compile scope, we should fix it.

## Brief change log

- Declare the dependent scope to compile for log4j2 related packages in the `ams-server` module

## How was this patch tested?

## Documentation


XBaith (approved):
LGTM

PR #2658, opened 2024-03-20
[Hotfix] Optimize codes of task status transition map

link3280:
Optimize codes and reduce memory footprints.

## Brief change log

Replace mutable collections with the immutable ones.

## How was this patch tested?

## Documentation


zhoujinsong (approved):
Nice catch and thanks for the contrtibution!

PR #2664, opened 2024-03-22
[WIP][AMORO-2623] Add CAS-based state mechanism to TaskRuntime

### LINKED ISSUE #2623: [Bug]: Optimizing process stuck after a few failures
### What happened?

If an optimizing task fails constantly (e.g. because of corrupted files), the optimizing task would be stuck at the `RUNNING` state after a few retries, while the task has already failed at the optimizer side.

截屏2024-03-11 16 55 27

### Affects Versions

0.5.1

AMS, Optimizer

### How to reproduce

1. create a table with some data that meets the optimizing conditions.
2. truncate a data file that would take part in the optimizing compaction.
3. AMS tries to optimize the table but fails and retries.
4. after a few retries, AMS would lose track of the task, and leave the task stuck at RUNNING state.

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

link3280:

Close #2623.

## Brief change log

- Add CAS check based on optimizer thread to TaskRuntime.

## How was this patch tested?

## Documentation

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
@@ -110,7 +110,7 @@ public class TaskRuntime extends CasStatedPersistentBase<Integer> {
 
   void reset() {
     resetCasRef();
-    invokeConsisitency(
+    invokeConsistency(
         () -> {
           statusMachine.accept(Status.PLANNED);
           startTime = ArcticServiceConstants.INVALID_TIME;
@@ -126,7 +126,7 @@ public class TaskRuntime extends CasStatedPersistentBase<Integer> {
   }
 
   public void schedule(OptimizerThread thread) {
-    invokeConsisitency(
+    invokeConsistency(
         () -> {
           statusMachine.accept(Status.SCHEDULED);
           token = thread.getToken();
@@ -137,7 +137,7 @@ public class TaskRuntime extends CasStatedPersistentBase<Integer> {
   }
 
   public void ack(OptimizerThread thread) {
-    invokeConsisitency(
+    invokeConsistency(
         () -> {
           validThread(thread);
           statusMachine.accept(Status.ACKED);
@@ -146,7 +146,7 @@ public class TaskRuntime extends CasStatedPersistentBase<Integer> {
   }
 
   void tryCanceling() {
-    invokeConsisitency(
+    invokeConsistency(
         () -> {
           if (statusMachine.tryAccepting(Status.CANCELED)) {
             endTime = System.currentTimeMillis();
--- a/ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
@@ -52,7 +52,7 @@ public class TaskRuntime extends CasStatedPersistentBase<Integer> {
   @StateField private long endTime = ArcticServiceConstants.INVALID_TIME;
   @StateField private long costTime = ArcticServiceConstants.INVALID_TIME;
   @StateField private String token;
-  @StateField private int threadId = -1;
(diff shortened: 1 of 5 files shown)


ams/server/src/main/java/com/netease/arctic/server/persistence/CasStatedPersistentBase.java
```diff
@@ -0,0 +1,67 @@
...
+    this((T) null);
+  }
+
+  protected CasStatedPersistentBase(T initialValue) {
+    super();
+    this.initialValue = initialValue;
+    this.casRef = new AtomicReference<>(initialValue);
+  }
+
+  protected final void invokeConsistencyWithCas(T expected, T target, Runnable runnable) {
```
majin1102, line 40: I think this method is unnecessary since you have created an extended implementation.

The simplest way I think is to extend the lock() and unlock(), what do you think

ams/server/src/main/java/com/netease/arctic/server/persistence/CasStatedPersistentBase.java
```diff
@@ -0,0 +1,67 @@
...
+
+package com.netease.arctic.server.persistence;
+
+import java.lang.reflect.Field;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class CasStatedPersistentBase<T> extends StatedPersistentBase {
+
+  private final AtomicReference<T> casRef;
```
majin1102, line 27: Let's discuss if AtomicReference<Thread> is proper here and make it decouple from extended classes?

ams/server/src/main/java/com/netease/arctic/server/persistence/CasStatedPersistentBase.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  protected final void invokeConsistencyWithCas(T expected, T target, Runnable runnable) {
+    T original = casRef.get();
+    Map<Field, Object> states = retainStates();
+    try {
+      doAsTransaction(runnable, () -> setCasRefOrError(expected, target));
+    } catch (Throwable throwable) {
+      restoreStates(states);
+      casRef.set(original);
```
majin1102, line 47: lock() is to set reference. unlock() is to clear.
I think set original is unnecessary

ams/server/src/main/java/com/netease/arctic/server/optimizing/TaskRuntime.java
```diff
@@ -52,18 +52,23 @@ public class TaskRuntime extends StatedPersistentBase {
...
   private RewriteFilesInput input;
   @StateField private RewriteFilesOutput output;
   @StateField private MetricsSummary summary;
   private Map<String, String> properties;
 
-  private TaskRuntime() {}
+  private static final int DUMMY_THREAD_ID = -1;
+
+  private TaskRuntime() {
+    super(DUMMY_THREAD_ID);
```
majin1102, line 66: Hey, I wonder if we could make cas decoupled with specific extended classes?

ams/server/src/main/java/com/netease/arctic/server/persistence/CasStatedPersistentBase.java
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+
+package com.netease.arctic.server.persistence;
+
+import java.lang.reflect.Field;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+
+public abstract class CasStatedPersistentBase<T> extends StatedPersistentBase {
```
majin1102, line 25: Cas is a good sense of implementation logic.

I would prefer OptimisticStatedPersistency, which could tell why we use this

majin1102 (changes requested):

majin1102:
Thanks for your contribution, I left some comments, PTAL

link3280:
Close in favor of #2684

Issue #2661, opened 2024-03-22
[Bug]: Button doesn't work in optimizing process detail page


XBaith:
### What happened?

When I want to cancel the running process, the OK and Cancel button was not working.
I run a front page locally, cancel button worked normally. But the OK button is most likely a back-end bug since it shows pending until the request timeout.

### Affects Versions

master

### Anything else

![image](https://github.com/NetEase/amoro/assets/54210725/1253f1ec-48bd-4066-b581-1a47d213b3b3)

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #2660, opened 2024-03-21
[Bug]: Expand icon doesn't work in optimizing tab


XBaith:
### What happened?

Do not display any information after clicking expand icon

### Affects Versions

master

### What engines are you seeing the problem on?

AMS

### Anything else

![image](https://github.com/NetEase/amoro/assets/54210725/35393034-2b18-4193-9e0f-94810903693b)

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

zhangmo8:
#2668 maybe this is helpful

PR #2670, opened 2024-03-25
[Hotfix] Add frontend plugin to build frontend automatically

zhoujinsong:

When I tried to find the cause of #2660 and #2661, I found that the reason of the issues was that the static files of `ams-dashboard` were not updated.

At present, every time when we make changes to the front-end code, we have to manually build and push the front-end static files, this process is complex and esily introduces unnecessary errors. So I've removed the pre-built front-end files from the repository and added a plugin for automatic building.

## Brief change log

- Remove static files from ams-dashboard.
- Add frontend plugin to build frontend automatically.

## How was this patch tested?

## Documentation


zhangmo8 (approved):
LGTM

zhongqishang:
@zhoujinsong Thanks for your contribution.
Do we not need the `frontend` profile? Would it be better to keep this and change the default value to true?

tcodehuber:
@zhoujinsong Maybe we should update the Markdown in `ams/dashboard/README.md`.
image

zhoujinsong:
> @zhoujinsong Thanks for your contribution. Do we not need the `frontend` profile? Would it be better to keep this and change the default value to true?

It seems necessary to build the frontend now, otherwise the unit tests for AMS will fail, and the packaged AMS won't run properly. Therefore, I don't think there's any need to retain this profile anymore.

zhoujinsong:
> @zhoujinsong Thx for your great job. Maybe we should update the Markdown in `ams/dashboard/README.md`. 

Thanks a lot for the review and suggestion. I have removed the out of date messages.

zhoujinsong:
@zhongqishang @tcodehuber 
I have made some changes according to the review comments, please take another look.

tcodehuber (approved):
LGTM, thanks for your great job.

zhongqishang (approved):
LGTM.

PR #2676, opened 2024-03-26
[AMORO-2652] Support Flink Optimizer run in session cluster

### LINKED ISSUE #2652: [Feature]: Support Flink optmizer run in session cluster
### Description

Currently Amoro already supports running tasks in application-mode and yarn-per-job. Both methods can run optmizer very well, but I would like to point out that maybe running tasks in session mode is also a good choice.

### Use case/motivation

1. Faster startup: session cluster is a pre-started Flink Cluster, which can be k8s, yarn or Standalone. At this time, we save the time required to deploy the cluster and start it directly in the session cluster.

2. Lower resource consumption: amoro's flink optmizer is essentially not a flink task in the traditional sense. It is a distributed service packaged by flink. In fact, it does not consume much resources on the jobmanager. If it is each flink optmizer Both require a cluster, so each job needs to start a jm, but in fact one jobmanager running multiple jobmasters is completely sufficient for flink optmizer, so it can save jm resources.

3. Faster demo: For example, like other frameworks, if the user wants to simply run amoro's flink optmizer, he can directly and quickly deploy a Flink Standalone cluster without configuring complex environments such as yarn k8s, and checkpoint storage media such as hdfs are also no required. No need (because essentially flink optmizer does not have checkpoint)

4. Automatic expansion and contraction? I think session mode may be used in amoro's optmizer to expand the capacity better. When expansion is needed, call the API to add a Taskmanager and submit the task. The same goes for shrinking. (Here you can use flink’s adapter scheduling mode？Perhaps further discussion is needed)

so, I think it is a nice feature to support running flink optmizer in session mode

### Describe the solution

The user configures an existing flink session cluster. Amoro can submit tasks through flink rest client, just like flink k8s operator.

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

huyuanfeng2018:

Close #2652.

## Brief change log

- Support Flink Optimizer run in session cluster by Flink Restclient.

## How was this patch tested?

#### Run multiple optimizer jobs in the same flink session
It has been verified that both Release and Scale-Out work normally
image

## Documentation


pom.xml
```diff
@@ -126,6 +126,7 @@
         <lucene.version>8.11.2</lucene.version>
         <bitmap.version>1.0.1</bitmap.version>
         <prometheus.version>0.16.0</prometheus.version>
+        <flink.version>1.18.1</flink.version>
     </properties>
```
czy006, line 130: If we introduce flink dependencies here, will it affect the flink version in the **mixed-format** and other dependencies? **flink-optimizer.flink-version** can specify flink dependencies

docs/admin-guides/managing-optimizers.md
```diff
@@ -89,6 +89,19 @@ containers:
...
+
+```yaml
+containers:
+  - name: flinkContainer
+    container-impl: com.netease.arctic.server.manager.FlinkOptimizerContainer
+    properties:
+      target: session                                                                # Flink run in session cluster
+      job-uri: "local:///opt/flink/usrlib/optimizer-job.jar"                         # Optimizer job main jar
+      ams-optimizing-uri: thrift://ams.amoro.service.local:1261                      # AMS optimizing uri 
+      export.FLINK_CONF_DIR: /opt/flink/conf/                                        # Flink config dir, flink-conf.yaml should e in this dir, contains the rest connection parameters of the session cluster
```
czy006, line 102: session cluster (jobmanager url) must be set in flink-conf?

ams/server/src/main/java/com/netease/arctic/server/manager/FlinkOptimizerContainer.java
```diff
@@ -371,6 +438,93 @@ private String buildReleaseKubernetesCommand(Resource resource) {
...
+      if (numSlotsAvailable < resource.getThreadCount()) {
+        throw new RuntimeException("No enough slots available in the cluster.");
+      }
+      File jarFile = new File(jobUri);
+      Preconditions.checkArgument(
+          jarFile.exists(), String.format("The jar file %s not exists", jarFile.getAbsolutePath()));
+
+      URL url = new URL(client.getWebInterfaceURL());
+      return runJar(
+          uploadJar(jarFile, url.getHost(), url.getPort(), flinkRestConf), flinkRestConf, resource);
```
czy006, line 458: need to upload the JAR every time, or will you take the previous JAR?

czy006 (commented):
I left some comments and question,PLAT. I also have a question, it seems that they all have the same assignment names, should we add some distinguishing elements @huyuanfeng2018

pom.xml
huyuanfeng2018: No, the dependency version of Flink is specified separately in other modules.

docs/admin-guides/managing-optimizers.md
huyuanfeng2018: Not necessarily. If it is a high-availability cluster, such as zk, you can directly specify the zk configuration. There is no need to configure the jobmanager address, such as:
```
high-availability: zookeeper
high-availability.zookeeper.quorum: xxx:2181
high-availability.zookeeper.path.root: /flink
high-availability.cluster-id: amoro-optimizer-cluster
high-availability.storageDir: hdfs://xxx/xxx/xxx
```

ams/server/src/main/java/com/netease/arctic/server/manager/FlinkOptimizerContainer.java
huyuanfeng2018: No, before uploading, the  flink Rest api will be called to check whether the optmizer jar package exists in the current cluster. The current stage is to determine whether it already exists through the file name. This may cause problems (the name is the same but the content is different), but I I think this is enough in most cases. If the jar package is indeed updated but the jar package name is not updated, just go to flink web ui and delete it manually.

huyuanfeng2018:
> it seems that they all have the same assignment names, should we add some distinguishing elements

+1, Maybe we need to open a new issue to do this

huyuanfeng2018:
> > it seems that they all have the same assignment names, should we add some distinguishing elements
> 
> +1, Maybe we need to open a new issue to do this

#2683

zhoujinsong (commented):
@huyuanfeng2018  I left some comments, PTAL.

ams/server/src/main/java/com/netease/arctic/server/manager/FlinkOptimizerContainer.java
```diff
@@ -18,28 +18,55 @@
 
 package com.netease.arctic.server.manager;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.netease.arctic.api.OptimizerProperties;
 import com.netease.arctic.api.resource.Resource;
+import com.netease.arctic.server.utils.FlinkClientUtil;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.curator.shaded.com.google.common.annotations.VisibleForTesting;
```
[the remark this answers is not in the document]
zhongqishang: We can add the non iceberg shade package to the checkstyle.

docs/admin-guides/managing-optimizers.md
baiyangtx: I suggest adding the parameters that must be set in session mode to the example. For example, jobmanager.url or zk-related settings

baiyangtx: Should optimizer-job.jar be added to the Flink Path when a session is started in session mode?

baiyangtx (commented):
Since this PR introduces flink dependencies, is it possible to implement several other modes of optimizer management based on FlinkClient?

ams/server/src/main/java/com/netease/arctic/server/manager/FlinkOptimizerContainer.java
```diff
@@ -307,27 +360,41 @@ private <T> T fetchCommandOutput(Process exec, Function<String, T> commandReader
...
-    }
+    if (target.runByFlinkRestClient()) {
+      Preconditions.checkArgument(
+          resource.getProperties().containsKey(SESSION_CLUSTER_JOB_ID),
+          "Cannot find {} from optimizer start up stats",
+          SESSION_CLUSTER_JOB_ID);
+      String jobId = resource.getProperties().get(SESSION_CLUSTER_JOB_ID);
+      try (RestClusterClient<String> restClusterClient =
+          FlinkClientUtil.getRestClusterClient(loadFlinkConfig())) {
+        Acknowledge ignore = restClusterClient.cancel(JobID.fromHexString(jobId)).get();
```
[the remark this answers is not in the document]
huyuanfeng2018: It is not needed here because `Acknowledge` is just a generic acknowledgement message. return result in flink client. If an exception occurs, Exception will be thrown directly here. However, may need to add a timeout detection here.

ams/server/src/main/java/com/netease/arctic/server/manager/FlinkOptimizerContainer.java
```diff
@@ -128,30 +170,41 @@ public void init(String name, Map<String, String> containerProperties) {
...
-          if (applicationId != null) {
-            startUpStatesMap.put(YARN_APPLICATION_ID_PROPERTY, applicationId);
-          }
-          break;
-        case KUBERNETES_APPLICATION:
-          startUpStatesMap.put(KUBERNETES_CLUSTER_ID_PROPERTY, kubernetesClusterId(resource));
-          break;
+    if (target.runByFlinkRestClient()) {
+      try {
+        JobID jobID = runFlinkOptimizerJob(resource, loadFlinkConfig());
```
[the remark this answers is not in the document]
huyuanfeng2018: thanks for your reminder! This logic is really missing here.

ams/server/src/main/java/com/netease/arctic/server/manager/FlinkOptimizerContainer.java
```diff
@@ -18,28 +18,55 @@
 
 package com.netease.arctic.server.manager;
 
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
```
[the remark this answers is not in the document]
huyuanfeng2018: done.

ams/server/src/main/java/com/netease/arctic/server/manager/FlinkOptimizerContainer.java
```diff
@@ -87,16 +117,28 @@ public class FlinkOptimizerContainer extends AbstractResourceContainer {
         return null;
       };
 
+  private final Supplier<ExecutorService> clientExecutorServiceSupplier =
+      Suppliers.memoize(
+          () ->
+              Executors.newFixedThreadPool(
+                  5,
+                  new ThreadFactoryBuilder()
+                      .setNameFormat("Flink-RestClusterClient-IO-%d")
```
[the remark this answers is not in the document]
huyuanfeng2018: fixed.

huyuanfeng2018:
> @huyuanfeng2018 I left some comments, PTAL.



> Since this PR introduces flink dependencies, is it possible to implement several other modes of optimizer management based on FlinkClient?

Theoretically, you can stop the job, but if you start the job, application mode should not be able to use FlinkClient to execute it.

docs/admin-guides/managing-optimizers.md
huyuanfeng2018: > I suggest adding the parameters that must be set in session mode to the example. For example, jobmanager.url or zk-related settings

+1, done.

huyuanfeng2018: > Should optimizer-job.jar be added to the Flink Path when a session is started in session mode?

sry. I don’t quite understand the meaning of adding `optimizer-job.jar` to the flink path. Can you explain it in a little more detail?

ams/server/src/main/java/com/netease/arctic/server/manager/FlinkOptimizerContainer.java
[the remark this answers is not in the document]
huyuanfeng2018: done.

baiyangtx (approved):
LGTM, This feature is very useful. Thanks for your contributions.

PR #2687, opened 2024-03-27
[AMORO-2667] Implement basic authentication for REST APIs

### LINKED ISSUE #2667: [Feature]: Add basic authentication to REST API
### Description

Currently, the REST APIs need an API token which is hard to use and integrate with external systems. Thus we've developed a basic auth support and want to contribute back to the community.

### Use case/motivation

_No response_

### Describe the solution

_No response_

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

link3280:

Close #2667.

## Brief change log

-

## How was this patch tested?

- [x]  Run test locally before making a pull request

## Documentation


baiyangtx (commented):
I don't think it's a good choice to introduce HTTP Basic Authentication in the REST API.

1. The HTTP Basic Authentication protocol uses plaintext to transmit passwords, which is not a very secure protocol.

2. For REST APIs, authentication only needs to verify that a caller is a legal caller. If you need to know which user triggered the call to the REST API, there are other mechanisms that can be used to solve this problem.

But on the other hand, I also agree that the current REST API authentication method is not user-friendly, and we should use a more standard REST API authentication protocol.

baiyangtx:
@zhoujinsong  WDYT?

link3280:
@baiyangtx Thanks for your input.

> 1. The HTTP Basic Authentication protocol uses plaintext to transmit passwords, which is not a very secure protocol.

Mostly, we should use Basic Authentication with HTTPS instead of HTTP, which is a common use case.  For example, [Starrocks](https://docs.starrocks.io/docs/reference/HTTP_API/SQL/#authentication-method) and [Kylin](https://cwiki.apache.org/confluence/display/KYLIN/Access+and+Authentication+API) support Basic Authentication.

WRT your concern, please see https://security.stackexchange.com/questions/988/is-basic-auth-secure-if-done-over-https

> 2. For REST APIs, authentication only needs to verify that a caller is a legal caller. If you need to know which user triggered the call to the REST API, there are other mechanisms that can be used to solve this problem.

Yes, Basic Auth only deals with Authn and is compatible with any other Authz (of course). But currently Amoro has no Authz functionalities (and I don't see the need for it), so the Authz issue could be discussed later if needed.

ams/server/src/main/java/com/netease/arctic/server/ArcticManagementConf.java
```diff
@@ -220,6 +220,18 @@ public class ArcticManagementConf {
...
+          .stringType()
+          .noDefaultValue()
+          .withDescription("The user name of the basic authentication used by REST APIs.");
+
+  public static final ConfigOption<String> HTTP_SERVER_BASIC_AUTH_PASSWORD =
+      ConfigOptions.key("http-server.basic-auth-password")
+          .stringType()
+          .noDefaultValue()
+          .withDescription("The password of the basic authentication used by REST APIs.");
+
```
baiyangtx, line 234: As we already have a `admin-username` and `admin-password`, maybe we could reuse them for REST API authentication.

And you can add a new config `rest-api.auth-type`,  and the value maybe `http-basic-auth` or `token`. the default could be `token`.

baiyangtx (commented):
@link3280 Thanks for your input, I agree with using Basic Authentication as a type of REST API auth implement.

But as for this PR, there are still some issues exists, and I left some comments.

link3280:
@baiyangtx updated as suggested. PTAL.

ams/server/src/main/java/org/apache/amoro/server/dashboard/DashboardServer.java
```diff
@@ -387,12 +398,29 @@ public EndpointGroup endpoints() {
...
+              basicAuthUser.equals(cred.component1())
+                  && basicAuthPassword.equals(cred.component2()),
+              "Invalid basic auth username or password.");
+        } catch (Exception e) {
+          LOG.error(
+              String.format(
+                  "Failed to authenticate via basic authentication. " + "Request url: %s %s.",
+                  ctx.req.getMethod(), uriPath),
+              e);
+          throw new SignatureCheckException();
```
baiyangtx, line 414:
```suggestion
      if ("basic".equalsIgnoreCase(authType)) {
        BasicAuthCredentials cred = ctx.basicAuthCredentials();
        if( !(basicAuthUser.equals(cred.component1())
                  && basicAuthPassword.equals(cred.component2()) {
             LOG.info(
              String.format(
                  "Failed to authenticate via basic authentication.  Request url: %s %s.",
                  ctx.req.getMethod(), uriPath));
             throw new SignatureCheckException();
         } 
```

baiyangtx: Info level is enough if authenticate is failed.

link3280:
@baiyangtx Fixed. PTAL.

baiyangtx (commented):
Cloud you please fix the conflict with master? 

And as this PR introduce new config paramters, please add them to the docs.

majin1102 (changes requested):
Thank you for introducing this feature.

I think this feature should be documented. Could you add some introduction in docs?

link3280:
> Thank you for introducing this feature.
> 
> I think this feature should be documented. Could you add some introduction in docs?

Sure.

link3280:
@baiyangtx @majin1102 Docs added as suggested. PTAL.

baiyangtx (approved):
LGTM, thanks for your contributions

baiyangtx:
@majin1102 PTAL

majin1102 (approved):
LGTM

link3280:
@baiyangtx @majin1102 May you kindly get this PR merged? Thanks!

link3280:
@zhoujinsong Thank you!

PR #2689, opened 2024-03-28
[AMORO-2688] Fix flink runtime info NullPointException

### LINKED ISSUE #2688: [Bug]: When optimizer flink runtime info is null flink optimizer initialization exception
### What happened?

image
In some deployment modes, flink `taskmanager.resource-id` cannot be successfully obtained. This time, the value is empty and an exception is caused.

https://github.com/apache/flink/blob/0e70d89ad9f807a5816290e9808720e71bdad655/flink-core/src/main/java/org/apache/flink/configuration/TaskManagerOptions.java#L343

### Affects Versions

master

Optimizer

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

huyuanfeng2018:

Close #2688.

## Brief change log

-

## How was this patch tested?

## Documentation


XBaith:
Friendly question: If the application_id and task_id is empty, how to identifier the flink executor

huyuanfeng2018:
> Friendly question: If the application_id and task_id is empty, how to identifier the flink executor

Good question. First of all, these two values ​​are used to locate the taskmanager executed by the optimizer. Theoretically speaking, `application_id` will not be empty. It has a default value. However, `taskmanager_id` may be empty in many cases. In this case, we can Use `subtaskid` to find the taskmanager corresponding to this `subtaskid` in flink webui. However, the `subtaskid` here is not stored as runtimeinfo. I will add this

Aireed:
@huyuanfeng2018  hi, additionally, may I ask for a detailed description of the circumstances under which these two operational information fields cannot be obtained?   I have tested scenarios for both k8s and yarn, but I'm not sure if I missed some cases?

huyuanfeng2018:
> @huyuanfeng2018 hi, additionally, may I ask for a detailed description of the circumstances under which these two operational information fields cannot be obtained? I have tested scenarios for both k8s and yarn, but I'm not sure if I missed some cases?

I'm running it locally using standlone mode，flink1.16 , This configuration item is indeed not generated. I am not sure if it is a flink bug.



image

huyuanfeng2018:
@Aireed hi,
I found flink's source for this:

https://issues.apache.org/jira/browse/FLINK-17579

It that only yarn and k8s clusters will override this parameter, and then give the corresponding `org.apache.flink.runtime.clusterframework.types.ResourceID` is used for initialization assignment. If there is no configuration, use `"RpcAddress:RpcPort" and a 6-character random string`. In standlone mode, if the user does not set `taskmanager.resource by himself -id` then this value will be empty.

zhongqishang:
Thanks for you contribution.

When i run the optimizer with IDEA, I meet this error. IMO, there are no other problems with protecting NPE.

+1 for this.

huyuanfeng2018:
Added subtask_index information to help determine taskmanager location when `taskmanager.resource-id` is empty.

cc @XBaith @Aireed @zhongqishang PTAL.

ams/optimizer/flink-optimizer/src/main/java/com/netease/arctic/optimizer/flink/FlinkExecutor.java
```diff
@@ -51,8 +51,13 @@ public void open() throws Exception {
...
     // set optimizer flink runtime info, including application_id, tm_id, host
-    executor.addRuntimeContext("application_id", applicationId);
-    executor.addRuntimeContext("tm_id", taskManagerId);
+    if (applicationId != null) {
+      executor.addRuntimeContext("application_id", applicationId);
+    }
+    if (taskManagerId != null) {
+      executor.addRuntimeContext("tm_id", taskManagerId);
+    }
+    executor.addRuntimeContext("subtask_index", String.valueOf(subTaskIndex));
```
zhongqishang, line 60: `subtask_index` doesn't seem to help determine which taskmanager it is, I can't see it on the flink web or in the logs.

huyuanfeng2018: > `subtask_index` doesn't seem to help determine which taskmanager it is, I can't see it on the flink web or in the logs.

thank you for your reply.
image
You can judge from here

zhongqishang: Get.

zhongqishang (approved):
LGTM.

PR #2695, opened 2024-04-01
[AMORO-1958] Implement KubernetesOptimizerContainer to support running Optimizer natively on K8S

### LINKED ISSUE #1958: [Feature]: Implement KubernetesOptimizerContainer to support running Optimizer natively on K8S.
### Description

Based on the discussion(https://github.com/NetEase/amoro/discussions/1929),  we should support native K8s optimizer scheduling.

### Use case/motivation

_No response_

### Describe the solution

_No response_

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

czy006:

Close #1958.

## Brief change log

- KubernetesOptimizerContainer Support Running On K8S

## How was this patch tested?

## Documentation

### DIFF
--- a/charts/amoro/templates/_optimizer.tpl
+++ b/charts/amoro/templates/_optimizer.tpl
@@ -84,3 +84,12 @@ properties:
   {{- end -}}
 {{- end -}}
 
+{{- define "amoro.optimizer.container.kubernetes" -}}
+container-impl: org.apache.amoro.server.manager.KubernetesOptimizerContainer
+properties:
+  namespace: {{ .Release.Namespace | quote }}
+  {{- with .Values.optimizer.kubernetes.properties -}}
+    {{- toYaml . | nindent 2 }}
+  {{- end -}}
+{{- end -}}
+
--- a/charts/amoro/templates/_optimizer.tpl
+++ b/charts/amoro/templates/_optimizer.tpl
@@ -87,7 +87,6 @@ properties:
 {{- define "amoro.optimizer.container.kubernetes" -}}
 container-impl: org.apache.amoro.server.manager.KubernetesOptimizerContainer
 properties:
-  namespace: {{ .Release.Namespace | quote }}
   {{- with .Values.optimizer.kubernetes.properties -}}
     {{- toYaml . | nindent 2 }}
   {{- end -}}
--- a/charts/amoro/values.yaml
+++ b/charts/amoro/values.yaml
@@ -174,10 +174,10 @@ optimizer:
   kubernetes:
     # enable the kubernetes optimizer container
     enabled: true
-    properties: {
-      kube-config-path: "～/.kube/config",
+    properties:
+      namespace: "default"
+      kube-config-path: "～/.kube/config"
       image: "arctic163/amoro:0.6"
-    }
   extra: []
 
 
--- a/charts/amoro/values.yaml
+++ b/charts/amoro/values.yaml
@@ -171,6 +171,13 @@ optimizer:
     name: ~
     # properties for optimizer container.
     properties: ~
+  kubernetes:
+    # enable the kubernetes optimizer container
+    enabled: true
+    properties: {
+      kube-config-path: "～/.kube/config",
+      image: "arctic163/amoro:0.6"
+    }
   extra: []
 
 
--- a/docs/admin-guides/managing-optimizers.md
+++ b/docs/admin-guides/managing-optimizers.md
@@ -30,6 +30,34 @@ containers:
       export.JAVA_HOME: "/opt/java"   # JDK environment
 ```
 
+### Kubernetes container
+
+Kubernetes Optimizer Container is a way to start Optimizer On K8s with standalone Optimizer.
+To use flink container, you need to add a new container configuration. 
+with container-impl as `org.apache.amoro.server.manager.KubernetesOptimizerContainer`
+
+KubernetesOptimizerContainer support the following properties:
+
+
+| Property Name             | Required | Default Value | Description                                                                                                   |
+|---------------------------|----------|---------------|---------------------------------------------------------------------------------------------------------------|
+| kube-config-path          | true     | N/A           | Kubernetes config location.                                                                                   |
+| image                     | true     | N/A           | Optimizer Image name                                                                                          |
+| namespace                 | false    | "default"     | The namespace of optimizer to deploy                                                                          |
+| ams-optimizing-uri        | false    | N/A           | uri of AMS thrift self-optimizing endpoint. This could be used if the ams.server-expose-host is not available |
+| cpu.factor                | false    | "1.0"         | Cpu factor when request kubernetes resource. Default 1 Cpu pre thread                                         |
+| memory                    | true     | N/A           | Memory usage for pre thread                                                                                   |
+
+
+```yaml
+containers:
+  - name: KubernetesContainer
+    container-impl: org.apache.amoro.server.manager.KubernetesOptimizerContainer
+    properties:
+      kube-config-path: ～/.kube/config
+      image: apache/amoro:0.6
+```
+
 ### Flink container
 Flink container is a way to start Optimizer through Flink jobs. With Flink, you can easily deploy Optimizer
 on yarn clusters or kubernetes clusters to support large-scale data scenarios. To use flink container, 
--- a/pom.xml
+++ b/pom.xml
@@ -127,6 +127,7 @@
         <bitmap.version>1.0.1</bitmap.version>
         <prometheus.version>0.16.0</prometheus.version>
         <flink.version>1.18.1</flink.version>
+        <fabric8-kubernetes-client.version.version>6.12.1</fabric8-kubernetes-client.version.version>
     </properties>
 
     <dependencies>
@@ -783,6 +784,13 @@
                 <artifactId>flink-runtime-web</artifactId>
                 <version>${flink.version}</version>
             </dependency>
+
+            <dependency>
+                <groupId>io.fabric8</groupId>
+                <artifactId>kubernetes-client</artifactId>
+                <version>${fabric8-kubernetes-client.version.version}</version>
+            </dependency>
+
 
             <dependency>
                 <groupId>junit</groupId>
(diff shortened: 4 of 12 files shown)


czy006:
@zhoujinsong @baiyangtx Currently optimizer.sh does not start in K8S, I try to find the reason and find that it is because of the lack of start-foreground logic in entrypoint.sh

If optimizer.sh is also added similar to ams.sh start
What do you think about the start-foreground stop logic?

ams/dist/src/main/arctic-bin/bin/optimizer.sh
```diff
@@ -37,14 +37,35 @@ else
...
     touch $STDERR_LOG
 fi
 
-JAVA_OPTS="-Xmx$1m -Dlog.home=${OPTIMIZER_LOG_DIR}"
+JAVA_OPTS="-Xmx$2m -Dlog.home=${OPTIMIZER_LOG_DIR}"
 RUN_SERVER="com.netease.arctic.optimizer.standalone.StandaloneOptimizer"
 CMDS="$JAVA_RUN $JAVA_OPTS $RUN_SERVER $ARGS"
-nohup ${CMDS} >/dev/null 2>${STDERR_LOG} &
+
+function start {
```
wolfboys, line 51: In sh, the defining function syntax is `function_name() { ... }`, while in bash, the commonly used syntax is function function_name { ... }, bash also supports the `function_name() { ... }` syntax, so it is recommended to use `function_name() { ... }`

ams/server/src/main/java/com/netease/arctic/server/manager/KubernetesOptimizerContainer.java
```diff
@@ -0,0 +1,155 @@
...
+    // point at amoro home in docker image
+    String startUpArgs =
+        String.format(
+            "/entrypoint.sh optimizer %s %s",
+            memory, super.buildOptimizerStartupArgsString(resource));
+    LOG.info("Starting k8s optimizer using k8s client with start command : {}", startUpArgs);
+
+    String namespace = resource.getProperties().getOrDefault(NAMESPACE, "default");
+    String image = checkAndGetProperty(resource.getProperties(), IMAGE);
+    String cpu = checkAndGetProperty(resource.getProperties(), CPU_PROPERTY);
```
baiyangtx, line 96: Is CPU requests/limits a must? 

I think we could ask for a bit more lenient and make it optional. This will provide more flexibility in some test scenarios

ams/server/src/main/java/com/netease/arctic/server/manager/KubernetesOptimizerContainer.java
```diff
@@ -0,0 +1,155 @@
...
+
+    String resourceId = resource.getResourceId();
+    String groupName = resource.getGroupName();
+
+    Deployment deployment =
+        new DeploymentBuilder()
+            .withNewMetadata()
+            .withName(NAME + "-" + resourceId)
+            .endMetadata()
```
baiyangtx, line 105: I think that we should support to assign a `K8S Namespace`.  

In a production scenario, the namespace of optimizer is usually different from AMS.

ams/server/src/main/java/com/netease/arctic/server/manager/KubernetesOptimizerContainer.java
```diff
@@ -0,0 +1,155 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(KubernetesOptimizerContainer.class);
+
+  public static final String MEMORY_PROPERTY = "memory";
+  public static final String CPU_PROPERTY = "cpu";
+  public static final String NAMESPACE = "namespace";
+  public static final String IMAGE = "image";
+  public static final String KUBE_CONFIG_PATH = "kube-config-path";
+
+  public static final String NAME = "amoro-kubernetes-optimizer";
```
baiyangtx, line 52: I think “amoro-optimizer"  is a better prefix.  

There are some display issues if resource name is too long when using kubctl.

ams/server/src/main/java/com/netease/arctic/server/manager/KubernetesOptimizerContainer.java
```diff
@@ -0,0 +1,155 @@
...
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public void requestResource(Resource resource) {
+    // generate pod start args
+    long memoryPerThread =
+        Long.parseLong(checkAndGetProperty(resource.getProperties(), MEMORY_PROPERTY));
+    long memory = memoryPerThread * resource.getThreadCount();
```
baiyangtx, line 86: You shoud implement all of those logic in the `protected Map<String, String> doScaleOut(Resource resource)`

ams/server/src/main/java/com/netease/arctic/server/manager/KubernetesOptimizerContainer.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  @Override
+  public void releaseOptimizer(Resource resource) {
+    String resourceId = resource.getResourceId();
+    LOG.info("release Kubernetes Optimizer Container {}", resourceId);
+    client
+        .apps()
+        .deployments()
+        .inNamespace(resource.getGroupName())
```
baiyangtx, line 151: resource group name is not `k8s namespace`

ams/server/src/main/java/com/netease/arctic/server/manager/KubernetesOptimizerContainer.java
```diff
@@ -0,0 +1,155 @@
...
+        .withName(NAME + "-" + resourceId)
```
baiyangtx, line 152: I suggest that you store the namespace and deployment's name information in the `doScaleOut`'s return and read it from resource properties during the relesae optimzier, so that there is better compatibility in future code refactoring.

charts/amoro/templates/_optimizer.tpl
```diff
@@ -84,3 +84,13 @@ properties:
   {{- end -}}
 {{- end -}}
 
+{{- define "amoro.optimizer.container.k8s" -}}
+container-impl: com.netease.arctic.server.manager.KubernetesOptimizerContainer
+properties:
+  kube-config-path: "～/.kube/config"
+  image: arctic163/amoro:0.6
```
baiyangtx, line 91: Template the image information from `values.yml` 

Or you can implement Helm charts in the other PR.

docs/admin-guides/managing-optimizers.md
```diff
@@ -30,6 +30,19 @@ containers:
...
+Kubernetes Optimizer Container is a way to start Optimizer On K8s with standalone Optimizer.
+
+```yaml
+containers:
+  - name: KubernetesContainer
+    container-impl: com.netease.arctic.server.manager.KubernetesOptimizerContainer
+    properties:
+      kube-config-path: ～/.kube/config
+      image: arctic163/amoro:0.6
+```
```
baiyangtx, line 44: We should support assign namespace. And user could assign namespace in `$AMS_HOME/conf/config.yaml` or `ResourceGroup`'s properties. 

The resource group has high priority.

baiyangtx (commented):
Thanks for your contributions, I left some comments. PTAL

ams/server/src/main/java/com/netease/arctic/server/manager/KubernetesOptimizerContainer.java
czy006: I think the first version will not introduce too complicated implementation, we may be able to unlock some advanced features when we implement PodTemplate later, the current default CPU Limit and Request is a simple implementation

baiyangtx:
>  This implementation is only a preliminary integration support, more advanced features such as PodTemplate, parameter merging, etc

As a basic implementation, I would suggest that you can leave memory and cpu limit out of support at first.  The rules for these two parameters can be more complex. For memory, you also need to reserve off-heap memory, and for CPU, you need to consider the number of threads, so I recommend that you don't limit it yet.

And we should support assign optimizer's namespace as well, I think we can do this in this pr.  @czy006

czy006:
Thanks to @baiyangtx for review and fix my code, now I am working well after local testing.

pom.xml
```diff
@@ -127,6 +127,7 @@
         <bitmap.version>1.0.1</bitmap.version>
         <prometheus.version>0.16.0</prometheus.version>
         <flink.version>1.18.1</flink.version>
+        <fabric8.version>6.11.0</fabric8.version>
```
zhoujinsong, line 130: Is `kubernetes-client.version` a proper name here as the fabric8 does not only mean Kubernetes?

czy006: Fabric8 kubernetes is the client an implementation, if use **kubernetes-client** name with fabric8. we may be confused with [kubernetes-client](https://github.com/kubernetes-client/java), use fabric8 client which has more powerful features

zhoujinsong: Then how about `fabric8-kubernetes-client.version`?

czy006: Now resolve to fabric8-kubernetes-client.version

charts/amoro/values.yaml
```diff
@@ -171,6 +171,13 @@ optimizer:
     name: ~
     # properties for optimizer container.
     properties: ~
+  kubernetes:
+    # enable the kubernetes optimizer container
+    enabled: true
+    properties: {
+      kube-config-path: "～/.kube/config",
+      image: "arctic163/amoro:0.6"
```
baiyangtx, line 179: Don't use hard code version in the default values.yaml

czy006:
image
if user set properties,it will be override this. use  helm `template --debug . -f ${user_yaml_file_path}/user_values_config.yaml ` can test it @baiyangtx

charts/amoro/values.yaml
czy006: I think this is just a means of filling the field by default, as long as the user writes the value will be overwritten as the true value, if not, follow the default

baiyangtx (approved):
LGTM, thanks for your contributions

PR #2715, opened 2024-04-03
[AMORO-2714]Delete resource when optimizer instance expired

### LINKED ISSUE #2714: [Improvement]: clean the resource when optimizer instance expired
### Search before asking

when release the optimizer, the amoro will remove the resource in the <db table :resource> and then remove the instance

when the optimizer expired, the amoro will remove the instance and will not remove resouce in the db, which will cause this dirty data,

### How should we improve?

we should flow the step when release the optimzier, when we detect the optimizer instance expired, not only should we remove the optimzier, we also should remove the resource record in the db

### Subtasks

_No response_

### Code of Conduct

Yao-MR:

Close #2714

## Brief change log
Delete resource when optimizer instance expired which will cause dirty data in db. and this data will never be used again.

## How was this patch tested?

## Documentation

- Does this pull request introduce a new feature? (yes )
- If yes, how is the feature documented? (not applicable )

### DIFF
--- a/ams/server/src/main/java/com/netease/arctic/server/DefaultOptimizingService.java
+++ b/ams/server/src/main/java/com/netease/arctic/server/DefaultOptimizingService.java
@@ -548,6 +548,7 @@ public class DefaultOptimizingService extends StatedPersistentBase
                           .forEach(task -> retryTask(task, queue)));
           if (isExpired) {
             LOG.info("Optimizer {} has been expired, unregister it", keepingTask.getOptimizer());
+            deleteResource(keepingTask.getOptimizer().getResourceId());
             unregisterOptimizer(token);
           } else {
             LOG.debug("Optimizer {} is being touched, keep it", keepingTask.getOptimizer());


ams/server/src/main/java/com/netease/arctic/server/DefaultOptimizingService.java
```diff
@@ -548,6 +548,7 @@ public void run() {
                           .forEach(task -> retryTask(task, queue)));
           if (isExpired) {
             LOG.info("Optimizer {} has been expired, unregister it", keepingTask.getOptimizer());
+            deleteResource(keepingTask.getOptimizer().getResourceId());
```
majin1102, line 551: The createResource() maybe called by user.
Is it proper that system auto deleting resources managed by user himself?

majin1102 (changes requested):

ams/server/src/main/java/com/netease/arctic/server/DefaultOptimizingService.java
Yao-MR: the createResource is actually called by user,
and the deleteResource is also will be called by user, if user want the release the optimzier,

but if the optimizer instance failed by some unexpected casuse, the system will delete the instance auto, but will
not clean the resource record in the record, and this resource will never be used again and will cause dead dirty data in the db forever. 

so when the instance is expired and be deleted by system, the resource associate with this instance should also be cleand

zhoujinsong:
@Yao-MR Thanks a lot for driving this improvement.

I have some questions about this PR:

- The resource id of the optimizer may be null, you may want to consider the situation.
- The optimizer may reconnect to the AMS even if the token is expired, and it will request a new token with the same resource id. So it may be dangerous to delete the resources as soon as the token expires.

ams/server/src/main/java/com/netease/arctic/server/DefaultOptimizingService.java
majin1102: I don't quite agree with that.

The record left in db has its meaning, cause when the optimizer came back, AMS could identify this resource. 
What you said I think is a dashboard design issue, the optimizer page don't reveal inactive optimizers. I think we could improve in that way.
A good and simple princeple is who creates who deletes.

Yao-MR: thanks your comment jin, agree with it, and will check the lifetime of resource again ,and get one graceful idea the resolve it.

Yao-MR:
consider the pr do not resolve this issue in an graceful way, will review this issue again, and redesign the code

PR #2716, opened 2024-04-05
[Hotfix] Change file seperator in suppressions file to adapt to the Windows system

zhoujinsong:

The suppressions file for the check style plugin does not work in the Windows system.
The reason is the file separator for Windows and Linux is not the same, we should change the file separator in the `tool/maven/suppressions.xml` to adapt to the Windows OS.

## Brief change log

- Change the file separator `/` with `[/\\]` to be compatible with both Linux and Windows systems.

## How was this patch tested?

## Documentation


baiyangtx (approved):
LGTM, thanks for your contribution.

PR #2722, opened 2024-04-08
chore: remove copyright.txt

caicancai:

remove copyright.txt

## Brief change log

-

## How was this patch tested?

## Documentation


caicancai:
@zhoujinsong I think the copyright can be deleted, what do you think?

zhongqishang:
@caicancai  Thanks for your contribution.

`copyright.txt` is used with the spotless plugin to generate and validate the copyright of the java/scala file. 
What is the reason for remove it?

zhongqishang:
> > @caicancai Thanks for your contribution.
> > `copyright.txt` is used with the spotless plugin to generate and validate the copyright of the java/scala file. What is the reason for remove it?
> 
> I'm not sure about this. I have the impression that many apache projects have deleted this file. According to you, we should keep it. Thank you for your reply.

This way can only include java/scala files and does not take effect for other files. I think other projects have been deleted because there is a better way to verify, like [flink license_check.sh](https://github.com/apache/flink/blob/master/tools/ci/license_check.sh), or use apache rat plugin, like kyuubi.

I think we should include all files as well.

caicancai:
close first

PR #2790, opened 2024-04-27
[AMORO-2623] Avoid deadlock among TaskRuntime objects

### LINKED ISSUE #2623: [Bug]: Optimizing process stuck after a few failures

majin1102:

Close #2623.

## Brief change log

- avoid deadlock among TaskRuntimes

## How was this patch tested?

## Documentation


majin1102:
@link3280

amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/optimizing/TaskRuntime.java
```diff
@@ -105,6 +105,7 @@ public void complete(OptimizerThread thread, OptimizingTaskResult result) {
           token = null;
           threadId = -1;
         });
+    owner.releaseResourcesIfNecessary();
```
link3280, line 108: How does it avoid the deadlock? The deadlock would happen when 2 thread blocks on `owner.acceptResult(this)` before this line.

link3280 (commented):
Hi @majin1102 , IIUC the main change is to check the process status and release tasks early right after every completed task (in TaskRuntime), but I don't get how it avoids the deadlock. Please take a look at the comments inline.

link3280:
ping @majin1102

amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/optimizing/TaskRuntime.java
link3280: @rfyu It sounds like what #2684 did.

XBaith (commented):
What's the difference between cancelTasks when 2 thread are failed at same time and releaseResourcesIfNecessary

amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/optimizing/TaskRuntime.java
majin1102: > I think two threads are blocked inside `cancelTasks` within `owner.acceptResult(this)`, so wrapping `cancelTask` within `releaseResourcesIfNecessary` and placing it outside `owner.acceptResult(this)` should indeed solve the deadlock issue

Sorry for late replying.

Just like @rfyu said, place canceling All Tasks out of task and process lock could prevent all lock conflict cases involving canceling all tasks.

When two threads steping into owner.acceptResult(), one will wait for lock anyway, I don't quite get your question

majin1102:
> What's the difference between cancelTasks when 2 thread are failed at same time and releaseResourcesIfNecessary

If you mean if  the operation of  canceling all tasks itself would lead to dead lock.

I think the answer is no, concurrent threads would step into task locks in the same order, which would not lead to any dead locks

amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/optimizing/TaskRuntime.java
link3280: I get the idea now. The cancellation is now lock-free. No task lock is needed, thus no deadlock.

zhoujinsong (approved):
LGTM.

PR #2793, opened 2024-04-28
[Hotfix] Apply the volatile modifier to stopped variable.

xleoken:

hotfix

## Brief change log

FlinkOptimizerExecutor runs in an independent thread, so the `stopped` variable is maintained by multiple threads.

image

## How was this patch tested?

## Documentation


XBaith (approved):
LGTM

PR #2801, opened 2024-04-29
Change Arctic to Amoro for thrift name in module core

zhongqishang:

Change Arctic to Amoro for thrift name in module core

## Brief change log

- modify thrift file name
- modify thrift class name
- regenerate thrift code 

## How was this patch tested?

## Documentation


xleoken (approved):

xleoken:
lgtm

zhoujinsong:
Hi

It seems this PR will make the older version client fail to connect to the server.
It may cause some problems when users try to upgrade to the new version.
Even if I think it is accepted as we have entered the ASF incubator and the next version will be a big new one and the name finally should be changed and this is a good time. But I still think we should discuss it in the dev mail list to make sure other developers and users know about it.

zhoujinsong (approved):
According the inputs from https://lists.apache.org/thread/swow5z1npcpq8j3yx4f5lhs0cr77fjlw, it is fine to merge this PR.

PR #2817, opened 2024-05-08
[Hotfix] Fix wrong usage of transient in HighAvailabilityContainer

link3280:

As titled.

## Brief change log

Replace transient keyword with volatile, cause followerLath could be updated by different threads.

## How was this patch tested?

## Documentation


majin1102 (approved):
LGTM

PR #2838, opened 2024-05-16
[AMORO-1744] Rename name with arctic identifier in amoro-mixed-format-trino module

zhoujinsong:

As titiled

## Brief change log

- Change the name of classes with the arctic identifier.
- Change the name of variables and methods with the arctic identifier.
- Change the comments with the arctic identifier.
- Change related documentation.

## How was this patch tested?

## Documentation


amoro-mixed-format/amoro-mixed-format-trino/src/main/java/org/apache/amoro/trino/MixedFormatConfig.java
```diff
@@ -21,8 +21,8 @@
 import io.airlift.configuration.Config;
 import io.airlift.configuration.ConfigDescription;
 
-/** Arctic config */
-public class ArcticConfig {
+/** Mixed-format configurations */
+public class MixedFormatConfig {
```
HuangFru, line 25: Maybe `AmoroConfig`? The same applies to the following configuration. I feel amoro.xxx would be better, because it actually represents the connector name. Users set these configurations according to connector.xxx, such as iceberg.xxx. Or, does this mean our connector name is mix-forma, if so, please ignore.

amoro-mixed-format/amoro-mixed-format-trino/src/main/java/org/apache/amoro/trino/MixedFormatConnectorFactory.java
```diff
@@ -56,22 +56,17 @@
...
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 
 /** Factory to generate {@link Connector} */
-public class ArcticConnectorFactory implements ConnectorFactory {
-
-  private static final Logger LOG = LoggerFactory.getLogger(ArcticConnectorFactory.class);
-
+public class MixedFormatConnectorFactory implements ConnectorFactory {
```
HuangFru, line 66: Same as above comment.

amoro-mixed-format/amoro-mixed-format-trino/src/main/java/org/apache/amoro/trino/MixedFormatConfig.java
zhoujinsong: Yes，the new connector name is mixed-format, and it is used for mixed-format only now.

HuangFru (approved):
LGTM

zhoujinsong:
Thanks for the review, @HuangFru

PR #2924, opened 2024-06-13
[AMORO-2925] Fix the issue of clean-orphan-file mistakenly deleting data files.

lintingbin:
When creating an Iceberg table on OSS and executing clean-orphan-file, there are normal data_files (just created but not yet committed to Iceberg) being cleaned up. The clean-orphan-file.min-existing-time-minutes parameter is not taking effect.  

Through debugging, it was found that because OSS does not record the file access time, getAccessTime returns 0, causing the clean-orphan-file.min-existing-time-minutes parameter to become ineffective.  

At the same time, the [listPrefix](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/hadoop/HadoopFileIO.java#L133-L152) function of Iceberg also uses getModificationTime. So using getModificationTime should be a better choice.

## Brief change log
- In the listDirectory function, replace getAccessTime with getModificationTime.

## How was this patch tested?

image

## Documentation

### DIFF
--- a/amoro-core/src/main/java/org/apache/amoro/io/AuthenticatedHadoopFileIO.java
+++ b/amoro-core/src/main/java/org/apache/amoro/io/AuthenticatedHadoopFileIO.java
@@ -100,7 +100,7 @@ public class AuthenticatedHadoopFileIO extends HadoopFileIO
                             new PathInfo(
                                 status.getPath().toString(),
                                 status.getLen(),
-                                status.getAccessTime(),
+                                status.getModificationTime(),
                                 status.isDirectory()))
                     .iterator();
             return () -> it;


zhoujinsong:
@lintingbin Thanks for reporting this bug and trying to fix it.

I suggest you create an issue and link this PR to it.
It will help a lot when others try to search for the same issues.

zhoujinsong (approved):
LGTM.

It will be fine to keep the same logic with the Iceberg action implemention.

lintingbin:
@zhoujinsong The issue has been added and linked to this PR.

zhoujinsong:
> @zhoujinsong The issue has been added and linked to this PR.

Thanks for that.
We can add a prefix `[AMORO-${issue number}]` to the PR's title(I have added it for you this time).

PR #2966, opened 2024-06-26
[AMORO-2965] Add ak/sk properties to TableMetaStore

xleoken:

As title.

## Brief change log

Add ak/sk properties to TableMetaStore.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-core/src/main/java/org/apache/amoro/table/TableMetaStore.java
+++ b/amoro-core/src/main/java/org/apache/amoro/table/TableMetaStore.java
@@ -541,6 +541,8 @@ public class TableMetaStore implements Serializable {
     private byte[] krbKeyTab;
     private byte[] krbConf;
     private String krbPrincipal;
+    private String accessKey;
+    private String secretKey;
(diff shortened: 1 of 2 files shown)


amoro-core/src/main/java/org/apache/amoro/table/TableMetaStore.java
```diff
@@ -541,6 +541,8 @@ public static class Builder {
     private byte[] krbKeyTab;
     private byte[] krbConf;
     private String krbPrincipal;
+    private String accessKey;
+    private String secretKey;
```
baiyangtx, line 545: It doesn't look like these two values ​​are read anywhere

xleoken: hi @baiyangtx, we need to read ak/sk in PaimonCatalogFactory when using s3 based catalog.

xleoken:
Thanks for reviewing @baiyangtx. Close this issue due to https://github.com/apache/amoro/pull/2972

PR #2973, opened 2024-06-27
[Improvement]: Eliminate AMS Full GC impact deriving from local terminal clean spark context #2969

### LINKED ISSUE #2969: [Improvement]: Eliminate AMS Full GC impact deriving from local terminal clean spark context
### Search before asking

We experienced and suffered from AMS full GC for a long period. we dig into this case and finally found out  local spark session terminal triggers java.lang.System.gc() every 30 minutes  to clean contexts. As arthas stack traces as shown below:
image

### How should we improve?

For external kyuubi session terminal, there is no need to revises anything. The problem is only related to local session which is embedding into AMS itself. Simply add  config 'spark.cleaner.referenceTracking' = 'false' to solve the problem.

### Subtasks

_No response_

### Code of Conduct

nicochen:
This pr is related to issue #2969 and has been manually tested locally.

Close #2969.

## Brief change log

-

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/terminal/local/LocalSessionFactory.java
+++ b/amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/terminal/local/LocalSessionFactory.java
@@ -132,6 +132,7 @@ public class LocalSessionFactory implements TerminalSessionFactory {
       sparkconf.set(
           "spark.sql.extensions",
           SparkContextUtil.MIXED_FORMAT_EXTENSION + "," + SparkContextUtil.ICEBERG_EXTENSION);
+      sparkconf.set("spark.cleaner.referenceTracking", "false");
 
       for (String key : this.conf.keySet()) {
         if (key.startsWith(SPARK_CONF_PREFIX)) {


daragu (approved):

baiyangtx:
I think it might be more reasonable to adjust the value of `spark.cleaner.periodicGC.interval`

From codes of spark, the parameter `spark.cleaner.referenceTracking` controls whether to create the context cleaner.

image

And inside the ContextCleaner, there are two theads managed.

- A cleanThread, will clean RDD nolonger be used.
- A gcThread, call `System.gc` every `spark.cleaner.periodicGC.interval` period.

image
image
image

I think the role of the cleanThread is quite important, especially when we start a resident Spark Context, which can release disk resources in time.
On the other hand, GC triggered by AMS is not a big problem. The parameter `spark.cleaner.periodicGC.interval` can be adjusted to a very large value to avoid actively triggering GC.

nicochen:
@baiyangtx I also considered the 'spark.cleaner.periodicGC.interval ' config key before, but as you said we can only set it an extremely large number rather than disabling it. In our production case, the full GC really matters. The 1.8 JDK uses parallel gc as full gc strategy for G1,.As a result, it takes more than 30 secs and triggers zk timeout and AMS failover which is unacceptable. Also, I believe the local terminal is designed to take lightweight and less frequent sql tasks and it would not produce too much RDD \ shuffle trash. The heavy tasks should go to kyubi or spark . Thus, I choose overall stability instead of enlarge number of gc interval.

zhoujinsong (approved):
LGTM.

zhoujinsong:
@baiyangtx  What do you think about the PR now?

baiyangtx (approved):
LGTM,  Diable RDD clean does not matter for local spark session.

PR #3007, opened 2024-07-05
[Hotfix] Meet NPE when loading tables which use s3 storage

tcodehuber:

Using ams internal catalog and s3 as storage, but NPE occurs. The root cause is that 
`org.apache.amoro.server.utils.InternalTableUtil#newIcebergFileIo L86 -> Configuration conf = store.getConfiguration();`

core-site and hdfs-site is null when using s3 storage.

>  

## Brief change log

-

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #3070, opened 2024-07-30
fix: Set up services according to the relative path of the project de…

zhangmo8:
…ployment structure

When the project is deployed under the non -service root, there will be a problem that the request cannot be requested to the back -end service

Close #xxx.

## Brief change log

-

## How was this patch tested?

## Documentation


chouchouji (approved):

czy006 (commented):
Can you explain why the pnpm-lock file changed? Generally this file is used to control some component versions of the front end cc @zhangmo8

zhangmo8:
> Can you explain why the pnpm-lock file changed? Generally this file is used to control some component versions of the front end cc @zhangmo8

Hi @czy006 

Because the version of pnpm in the pipeline was upgraded in this [commit](https://github.com/apache/amoro/commit/8b9822d0e74018e55bfd5cd68314a9ec6267acf9). In order to keep the versions of the development environment and the production environment consistent, I switched to pnpm 9.6 and regenerated the pnpm-lock file

zhoujinsong (approved):
LGTM.

PR #3071, opened 2024-07-30
[AMORO-3084] Upgrade to iceberg 1.6.1

czy006:

#3084.

## Brief change log

- Upgrade iceberg to 1.6.X
- Mixed Format iceberg version keep to 1.4.3

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/IcebergRewriteExecutor.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/IcebergRewriteExecutor.java
@@ -24,7 +24,6 @@ import org.apache.amoro.io.writer.IcebergFanoutPosDeleteWriter;
 import org.apache.amoro.table.MixedTable;
 import org.apache.amoro.utils.map.StructLikeCollections;
 import org.apache.iceberg.PartitionSpec;
-import org.apache.iceberg.StructLike;
 import org.apache.iceberg.TableProperties;
 import org.apache.iceberg.data.GenericAppenderFactory;
 import org.apache.iceberg.data.IdentityPartitionConverters;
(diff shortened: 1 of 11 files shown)


czy006:
- https://github.com/apache/iceberg/blob/main/flink/v1.18/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java#L161
- https://github.com/apache/iceberg/blob/apache-iceberg-1.4.3/flink/v1.15/flink/src/main/java/org/apache/iceberg/flink/source/ScanContext.java#L149

Diff,Run TestFlinkSource.testUnkeyedSnapshotRead Have Exception
image

czy006:
ops,some time ci was blocker have long time.Maybe we should check

baiyangtx:
It looks like there will be many engine versions that will no longer be supported.

@zhoujinsong PTAL

zhoujinsong:
Maybe we should follow Iceberg 1.6, support:

- Flink 1.17/1.18/1.19
- Spark 3.3/3.4/3.5

What do you think?

czy006:
> Details

This will be supported in the future PR, Flink 1.20 as the LTS long-term support version, I recommend as the default version option

czy006:
> > Details
> 
> This will be supported in the future PR, Flink 1.20 as the LTS long-term support version, I recommend as the default version option



> Maybe we should follow Iceberg 1.6, support:
> 
> * Flink 1.17/1.18/1.19
> * Spark 3.3/3.4/3.5
> 
> What do you think?

This will be support in the future PR, Flink 1.20 as the LTS long-term support version, I recommend as the default version option

czy006:
image

Optimizing Task Run Normal

zhongqishang:
I think it is not a good idea to remove so many versions support. 
What is the reason for deleting the support? Is it just that the official does not support it?

Maybe we can keep the combination of old flink and old iceberg versions. @zhoujinsong @baiyangtx @czy006

czy006:
> I think it is not a good idea to remove so many versions support. What is the reason for deleting the support? Is it just that the official does not support it?
> 
> Maybe we can keep the combination of old flink and old iceberg versions. @zhoujinsong @baiyangtx @czy006

mixed format relies on the spark and flink versions, and currently the iceberg community only supports the latest 3 versions. For example, they have to give up [spark 3.2](https://mvnrepository.com/artifact/org.apache.iceberg/iceberg-spark-runtime-3.2) their support.the last support in 1.4.3, iceberg if we upgrade version, we must give up its support it.flink version seems(I will add flink1.19 1.20 in the next pr),Spark now just support 3.3

czy006:
I have initiated relevant discussions on the email list:
https://lists.apache.org/thread/1crbstg7prjxz8xokd7dtj1o5mgv1034

czy006:
cc @zhoujinsong

zhoujinsong (approved):
LGTM.

PR #3113, opened 2024-08-19
[AMORO-3110] Avoid optimization status blocking in running

### LINKED ISSUE #3110: [Bug]: Optimization state blocking in running
### What happened?

When the insertion into table `optimizing_task_quota` fails, subsequent logic cannot proceed because all database operations are wrapped in the same session. Even if the exception is caught, `com.netease.arctic.server.persistence.NestedSqlSession#isRollingback` will still be true.Therefore, an exception will be thrown when checking the session status in `org.apache.amoro.server.persistence.NestedSqlSession#checkState`. So the table optimization status is stuck in running.

### Affects Versions

master

Iceberg

AMS

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

rfyu:

Close #3110.

## Brief change log

-

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
@@ -468,6 +468,7 @@ public class OptimizingQueue extends PersistentBase {
         }
       } catch (Exception e) {
         LOG.error("accept result error:", e);
+        throw e;
       } finally {
         lock.unlock();
       }
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
@@ -466,9 +466,6 @@ public class OptimizingQueue extends PersistentBase {
             persistProcessCompleted(false);
           }
         }
-      } catch (Exception e) {
-        LOG.error("accept result error:", e);
-        throw e;
       } finally {
         lock.unlock();
       }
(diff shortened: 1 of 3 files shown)


amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -468,6 +468,7 @@ public void acceptResult(TaskRuntime taskRuntime) {
         }
       } catch (Exception e) {
         LOG.error("accept result error:", e);
+        throw e;
```
zhoujinsong, line 471: I agree that we should throw the exception rather than just printing logs.
But we may just remove the catch blocks because we will print error logs in the service proxy.

amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/table/TableRuntime.java
```diff
@@ -409,7 +409,7 @@ private boolean updateConfigInternal(Map<String, String> properties) {
   }
 
   public void addTaskQuota(TaskRuntime.TaskQuota taskQuota) {
-    doAs(OptimizingMapper.class, mapper -> mapper.insertTaskQuota(taskQuota));
+    doAsAlone(OptimizingMapper.class, mapper -> mapper.insertTaskQuota(taskQuota));
```
zhoujinsong, line 412: It seems we started another transaction to update the quota.
Another method to solve this issue is using the same transaction and ignoring the persistent error. We can add a method like `doAsIgnoreError` in PersistentBase to get that.

What do you think is the better method? @rfyu

zhoujinsong (commented):
@rfyu Thanks for trying to solve this issue.

I left some comments, PTAL.

amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/table/TableRuntime.java
rfyu: thanks for your suggestion. It is acceptable to handle it using the same transaction. I will make the necessary modifications accordingly.

zhoujinsong:
LGTM.

However, there are some conflicts that should be fixed. @rfyu

zhoujinsong (approved):
LGTM.

PR #3115, opened 2024-08-19
[AMORO-3111]Avoid optimization state blocking in planning

### LINKED ISSUE #3111: [Bug]: Optimization state blocking in planning
### What happened?

The exception thrown by the `org.apache.amoro.server.optimizing.OptimizingQueue#planInternal` was not handled, and the catch content will also cause an exception because the jdbc connection is not obtained. So the table is always in the running state.
image

### Affects Versions

master

Iceberg

AMS

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

rfyu:

Close #3111.

## Brief change log

-

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
@@ -305,7 +305,7 @@ public class AmoroManagementConf {
   public static final ConfigOption<Long> DB_CONNECT_MAX_WAIT_MILLIS =
       ConfigOptions.key("database.connection-pool-max-wait-millis")
           .longType()
-          .defaultValue(1000L)
+          .defaultValue(-1L)
           .withDescription("Max wait time before getting a connection timeout.");
 
   public static final ConfigOption<Long> OPTIMIZER_HB_TIMEOUT =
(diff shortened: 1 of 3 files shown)


amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/table/TableRuntime.java
```diff
@@ -188,13 +188,21 @@ public void beginPlanning() {
...
+          () -> {
+            OptimizingStatus originalStatus = optimizingStatus;
+            updateOptimizingStatus(OptimizingStatus.PENDING);
+            persistUpdatingRuntime();
+            tableHandler.handleTableChanged(this, originalStatus);
+          });
+    } catch (Exception e) {
+      OptimizingStatus originalStatus = optimizingStatus;
+      updateOptimizingStatus(OptimizingStatus.PENDING);
+      LOG.error(
```
zhoujinsong, line 202: Warn logging level may be proper here.

amoro-ams/amoro-ams-server/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
```diff
@@ -299,7 +299,7 @@ public class AmoroManagementConf {
   public static final ConfigOption<Long> DB_CONNECT_MAX_WAIT_MILLIS =
       ConfigOptions.key("database.connection-pool-max-wait-millis")
           .longType()
-          .defaultValue(1000L)
+          .defaultValue(-1L)
```
zhoujinsong, line 302: We may need to roll back the change here.

zhoujinsong (commented):
@rfyu Thanks for the contribution, I left some comments, PTAL.

zhoujinsong (approved):
LGTM.

PR #3126, opened 2024-08-22
[AMORO-3152] Make git-commit-plugin fail on no git dir configurable

### LINKED ISSUE #3125: [Improvement]: Move failOnNoGitDirectory of git-commit-plugin to profile
### Search before asking

In  #3042 we add configuration `failOnNoGitDirectory` for git-commit-plugin so that it would not block the release process, but in the normal process, it would not fail even if there is not git dir, we want it fail if there is not git dir in the normal case.

### How should we improve?

move the configuration to profile, and enable it when release

### Subtasks

_No response_

### Code of Conduct

klion26:

We add conf to fail silent when no git dir before, so that it would not block the
release process, but it may not fail in normal mode also, in this commit
we move the configuration to profile and can be enabled in the release process

Close #3125 .

## Brief change log

-

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/amoro-ams-server/pom.xml
+++ b/amoro-ams/amoro-ams-server/pom.xml
@@ -34,6 +34,10 @@
     <name>Amoro Project AMS Server</name>
     <url>https://amoro.apache.org</url>
 
+    <properties>
+        <git-commit-id-plugin.fail-on-no-git-dir>false</git-commit-id-plugin.fail-on-no-git-dir>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>org.apache.amoro</groupId>
@@ -415,8 +419,8 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <failOnNoGitDirectory>false</failOnNoGitDirectory>
-                    <failOnUnableToExtractRepoInfo>false</failOnUnableToExtractRepoInfo>
+                    <failOnNoGitDirectory>${git-commit-id-plugin.fail-on-no-git-dir}</failOnNoGitDirectory>
+                    <failOnUnableToExtractRepoInfo>${git-commit-id-plugin.fail-on-no-git-dir}</failOnUnableToExtractRepoInfo>
                     <dotGitDirectory>${project.basedir}/.git</dotGitDirectory>
                     <generateGitPropertiesFile>true</generateGitPropertiesFile>
                     <generateGitPropertiesFilename>
@@ -490,6 +494,12 @@
                     <artifactId>amoro-hudi-format</artifactId>
                 </dependency>
             </dependencies>
+        </profile>
+        <profile>
+            <id>fail-on-no-git-dir</id>
+            <properties>
+                <git-commit-id-plugin.fail-on-no-git-dir>true</git-commit-id-plugin.fail-on-no-git-dir>
+            </properties>
         </profile>
     </profiles>
 </project>


klion26:
@zhoujinsong @zhongqishang Could you please have a look at this when you're free, thanks.

Could you please share where the release script is located in `$AMORO_DIR/tools/releasing/create_binary_release.sh`, if not where is it, I can change it to respect this after this commit.

zhoujinsong:
> Could you please share where the release script is located in `$AMORO_DIR/tools/releasing/create_binary_release.sh`, if not where is it, I can change it to respect this after this commit.

You can add this new profile here: https://github.com/apache/amoro/blob/master/tools/releasing/create_binary_release.sh#L68

pom.xml
```diff
@@ -1324,5 +1324,21 @@
...
         </profile>
+        <profile>
+            <id>fail-silent-on-no-git-dir</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>pl.project13.maven</groupId>
+                        <artifactId>git-commit-id-plugin</artifactId>
+                        <version>4.0.0</version>
+                        <configuration>
```
zhoujinsong, line 1335: I am not sure if the profile plugin configuration will overwrite the default configuration or merge into the default configuration.

As you just added the new configurations, it may seem to be they will be merged into the default configuration, do we have any documentation about profile behavior about this?

zhoujinsong (commented):
@klion26 Thanks for the contribution, I left some comments.

pom.xml
klion26: the configuration in profile will overwrite these in the build configuration.  the priority for the configuration is: `configuration in profile(after enabled)` > `build configuration`.

can ref the resource[1][2] here.
[1] https://maven.apache.org/guides/introduction/introduction-to-profiles.html?spm=5176.28103460.0.0.6077572cRnkc6p#profile-inheritance
[2] https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Project_Inheritance

klion26:
After an offline discussion with @zhongqishang  we reached to consense that: add a new profile to enable the `no git dir` check, and make the check disabled default, because this has a better experience with new user which download the code with zip format.
 
The code has been changed to the above change, please have a look when you're free thanks. @zhoujinsong @zhongqishang 

normal build `mvn396 clean package  -pl '!amoro-mixed-format/amoro-mixed-format-trino' -DskipTests  -Pskip-dashboard-build`
WeChatWorkScreenshot_fee3c520-79b2-46ab-903f-55c62c380d30

build after the check enabled `mvn396 clean package  -pl '!amoro-mixed-format/amoro-mixed-format-trino' -DskipTests  -Pskip-dashboard-build -Pfail-on-no-git-dir`
WeChatWorkScreenshot_9c9ce7ce-80a0-433c-b27d-311dc3782d68

zhongqishang (approved):
IMO, not failing by default is a good choice. This PR also provides the option to fail.

LGTM.

klion26:
The pr changed to using a profile with configuration inspired by @zhoujinsong 

The result can be found below

WeChatWorkScreenshot_32715e4b-c883-4d0e-822c-dc5e0295fb52

WeChatWorkScreenshot_8cf631f0-fd12-46ab-9070-5e0d2bde34ed

zhoujinsong:
Thanks for the contribution! @klion26

klion26:
@zhoujinsong @zhongqishang thanks for the review and merging!

PR #3130, opened 2024-08-23
[AMORO-3110]extract error message in optimizers

### LINKED ISSUE #3110: [Bug]: Optimization state blocking in running

XBaith:

Close #3110 .

## Brief change log

-

## How was this patch tested?

## Documentation


baiyangtx (approved):
LGTM

PR #3164, opened 2024-09-03
[AMORO-3163] Support for configuring the number of remaining snapshots

### LINKED ISSUE #3163: [Feature]: Support for configuring the number of remaining snapshots
### Description

Currently, when we execute expired snapshots, we only consider the TTL as a factor. However, Iceberg supports setting a minimum number of snapshots to retain. I believe we should also make this configuration option available(eg. `snapshot.base.keep.min-count`). For developers and table users, these snapshots can be understood as the update logs of the table, making it easier for users to review the history of updates. If there are no updates for an extended period, Amoro may only retain a single snapshot after cleanup, which could result in a loss of information for users.

### Use case/motivation

_No response_

### Describe the solution

_No response_

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

XBaith:

Close #3163 .

## Brief change log

- Retain last n snapshots

## How was this patch tested?

## Documentation


tcodehuber (approved):
LGTM

PR #3179, opened 2024-09-06
[Improvement][AMORO-3178]: Change the default configuration of values.yaml to improve user experience

### LINKED ISSUE #3178: [Improvement]: Change the default configuration of values.yaml to improve user experience
### Search before asking

The current default of prometheusExporter in values.yaml is like this：
image

When the user wants to experience it, open the prometheusExporter configuration
like this: 

image

In fact, using helm deployment in this way will report an error.

image

### How should we improve?

Change the values.yaml default configuration like this：
image

### Subtasks

_No response_

### Code of Conduct

upczsh:

Close #3178 

## Brief change log

- as title

## How was this patch tested?

**Open the prometheus annotation in values ​​and helm deployment will report an error.**

image

**Apply this patch and open prometheus annotation for helm deployment.**

image

## Documentation


charts/amoro/templates/_pod.tpl
```diff
@@ -120,7 +120,7 @@ spark distribution package will be installed to here*/ -}}
 {{- if .Values.plugin.metricReporters }}
 {{- if .Values.plugin.metricReporters.prometheusExporter.enabled }}
 - name: prometheus
-  containerPort: {{ .Values.plugin.metricReporters.prometheusExporter.port }}
```
czy006, line 123: if have anothor properties,not just only port.maybe can use map set of this config

upczsh: Thank you @czy006 , is this what you said?
 

    prometheusExporter:

      name: prometheus-exporter

      enabled: true

      port: 7001

      properties:

        port: {{ .Values.plugin.metricReporters.prometheusExporter.port }}

czy006 (approved):

PR #3193, opened 2024-09-10
[AMORO-3192] Add the "clean-orphan-file.interval-time-minutes" parameter

### LINKED ISSUE #3192: [Improvement]: Add the "clean-orphan-file.interval-time-minutes" parameter.
### Search before asking

By adding the parameter "clean-orphan-file.interval-time-minutes" to control the frequency of cleaning orphan files, users can adjust the cleaning frequency as needed instead of a fixed 24-hour interval.

### How should we improve?

_No response_

### Subtasks

_No response_

### Code of Conduct

lintingbin:

Close #3192.

## Brief change log

- Add a parameter to replace the INTERVAL constant, which was previously fixed at 24 hours in the OrphanFilesCleaningExecutor class.

## How was this patch tested?

## Documentation

- If yes, how is the feature documented? docs


tcodehuber (commented):
Thanks for your work, but I think the default check interval should be less than the default min existing time.

lintingbin:
> Thanks for your work, but I think the default check interval should be less than the default min existing time.

Why is this necessary? The "min existing time" ensures that useful files are not mistakenly deleted, so setting it to two to three days is usually sufficient. However, cleaning orphan files is a performance-intensive task, so it's generally appropriate to configure the cleanup to occur once a week or once a month.

zhoujinsong:
@lintingbin Thanks for the contribution.

The scheduling interval for the current expiration task is usually set as an AMS configuration like `data-expiration.interval`, which applies to all tables. I think this meets the requirements, as we generally do not adjust different execution interval for different tables. What do you think?

BTW, we are improving the configuration value for time duration with unit support like 3d, we can do it for the new configuration.

zhoujinsong:
> Why is this necessary? The "min existing time" ensures that useful files are not mistakenly deleted, so setting it to two to three days is usually sufficient. However, cleaning orphan files is a performance-intensive task, so it's generally appropriate to configure the cleanup to occur once a week or once a month.

I agree with @lintingbin, there is no connection between the values of these two configurations.

lintingbin:
> @lintingbin Thanks for the contribution.
> 
> The scheduling interval for the current expiration task is usually set as an AMS configuration like `data-expiration.interval`, which applies to all tables. I think this meets the requirements, as we generally do not adjust different execution interval for different tables. What do you think?
> 
> BTW, we are improving the configuration value for time duration with unit support like 3d, we can do it for the new configuration.

The modifications have been completed according to the suggestions.

docs/user-guides/configurations.md
```diff
@@ -72,6 +72,7 @@ Data-cleaning configurations are applicable to both Iceberg Format and Mixed str
...
 | snapshot.keep.min-count                     | 1                | Minimum number of snapshots retained for table expiration                                                                                                                                                                                                             |
 | clean-orphan-file.enabled                   | false            | Enables periodically clean orphan files                                                                                                                                                                                                                               |
+| clean-orphan-file.interval                  | 1d               | The interval for Cleaning orphan files within a specified time                                                                                                                                                                                                        |
```
zhoujinsong, line 75: This change may need to be rollback

And you may want to add this new configuration example into `dist/src/main/amoro-bin/conf/config.yaml`

zhoujinsong (approved):
LGTM.

PR #3203, opened 2024-09-18
[Amoro-3196]: Fix scheduling hungry issue

### LINKED ISSUE #3196: [Bug]: catalog table filtering problem
### What happened?

We have created two catalogs, zj_catalog and error_catalog, during our usage.  Some tables in zj_catalog are in a pending state if no catalog selection is made.  However, when zj_catalog is selected, there are no corresponding tables in pending, and searching for the tables directly is also unsuccessful.  Moreover, these tables have not been planned or executed, and they are stuck in the pending state.
image
image

### Affects Versions

0.7.0

_No response_

_No response_

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

majin1102:

Close #3196.

The real issue is scheduling hunger problem
## Brief change log

The table runtime quotas are cached in the sorter once it has been calculated.
The correct logic is calculating quota on every single scheduling phase, and only cache quota result in a single phase.

So the fix is create a sorter for every schedule call.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/SchedulingPolicy.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/SchedulingPolicy.java
@@ -78,9 +78,9 @@ public class SchedulingPolicy {
 
   private Comparator<TableRuntime> createSorterByPolicy() {
     if (policyName.equalsIgnoreCase(QUOTA)) {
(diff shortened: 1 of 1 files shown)


zhoujinsong (approved):
LGTM.

Thanks for the contribution!

czy006 (approved):

PR #3204, opened 2024-09-18
[Improvement] Remove not necessary imports

chouchouji:
It is not necessary to import `defineProps` api about vue.

image


zhangmo8 (approved):
LG

chouchouji:
> LG

Thanks for your review!

PR #3217, opened 2024-09-23
[hotfix] ams-mysql-init.sql fix error init

mansonliwh:

In ams-mysql-init.sql, the SQL for creating the table_runtime table is incorrect, which will cause the service to fail to start when redeploying Amoro.


czy006 (approved):
LGTM

zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #3222, opened 2024-09-24
[Improvement] Support env.sh and add HADOOP_CONF_DIR and HIVE_CONF_DIR to CLASSPATH

baiyangtx:

HiveConf can load the existing hive-site.xml from the classpath during initialization.

So Amoro should support adding some directories to the classpath at startup.

In addition, user-defined environment variable configuration files need to be supported.

## Brief change log

- Add a env.sh in ${AMORO_CONF_DIR} 
- If exists env variables HADOOP_CONF_DIR and HIVE_CONF_DIR, add them to CLASSPATH
- Support add additional directories by env variable AMORO_ADDITION_CLASSPATH.
- Print all env variables and their value at startup.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
@@ -439,6 +439,8 @@ public class AmoroServiceContainer {
 
     private Map<String, Object> initEnvConfig() {
       LOG.info("initializing system env configuration...");
+      Map<String, String> envs = System.getenv();
+      envs.forEach((k, v) -> LOG.info("export {}={}", k, v));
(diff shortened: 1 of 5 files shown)


huyuanfeng2018:
If we can export hadoop's classpath, do we still need to package the hadoop client into amoro?

czy006 (approved):

czy006:
> If we can export hadoop's classpath, do we still need to package the hadoop client into amoro?

If it's configured, we really don't need it. But users have to manually deploy the dependencies of environments like Hadoop and Hive. This might become an optional replacement.

baiyangtx:
> If we can export hadoop's classpath, do we still need to package the hadoop client into amoro?

I think we can release binary packages without Hadoop/Hive dependencies

PR #3231, opened 2024-09-30
[AMORO-3218] Fix logic for consturct the result of optimizer tables api

### LINKED ISSUE #3218: [Bug]: optimizing Tab paging display error
### What happened?

image
No matter how many items are selected to be displayed per page, there will only be one page in the end.

### Affects Versions

master

_No response_

AMS

### How to reproduce

_No response_

### Relevant log output

_No response_

### Anything else

_No response_

### Code of Conduct

klion26:

Close #3218 .

## Brief change log

-

## How was this patch tested?

image

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
@@ -62,7 +62,9 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


klion26:
cc @huyuanfeng2018 @zhoujinsong @majin1102

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
```diff
@@ -471,6 +471,11 @@ public TableRuntime getRuntime(Long tableId) {
     return tableRuntimeMap.get(tableId);
   }
 
+  public int getRuntimeCount() {
```
huyuanfeng2018, line 474: What if we have filters?
For example, database is specified in search

klion26: @huyuanfeng2018 thanks for pointing this out, fix it.

huyuanfeng2018 (approved):
LGTM. @klion26  There are some conflicts that need to be resolved

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/OptimizerController.java
```diff
@@ -75,10 +76,15 @@ public void getOptimizerTables(Context ctx) {
     List<TableRuntime> tableRuntimes =
         tableRuntimeBeans.stream()
             .map(meta -> tableService.getRuntime(meta.getTableId()))
+            .filter(Objects::nonNull)
```
majin1102, line 79: Since we are using the in-memory table runtime list, why do we still need to query the ID list from the database and get list count from memory? 
It seems quite fragmented

amoro-ams/src/main/java/org/apache/amoro/server/table/TableManager.java
```diff
@@ -33,6 +35,9 @@ public interface TableManager extends TableRuntimeHandler {
 
   TableRuntime getRuntime(Long tableId);
 
+  /** Get the count of table runtimes associated to the given filter. */
+  int getRuntimeCount(@Nullable String dbFilter, @Nullable String tableFilter);
```
majin1102, line 39: I think instead of defining such an interface, it would be better to directly define a listRuntimes interface.

majin1102 (changes requested):
Thanks for your contributions

I left some comments, PTAL

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/OptimizerController.java
klion26: Yes, this is defensive programming, because we will process the information we get here(TableRuntime list) later and throw Exception in `OptimizingUtil#buildTableOptimizeInfo`), of course, we can also remove it here if needed

klion26:
@majin1102 I've addressed the comments, please take another look when you're free. thanks.

majin1102 (approved):
LGTM

klion26:
@huyuanfeng2018 @majin1102 thanks for the review and merging!

PR #3290, opened 2024-10-22
[ARORO-3289] Avoid calling getMixedTablePartitionSpecById in the scan loop

### LINKED ISSUE #3289: [Improvement]: Avoid calling getMixedTablePartitionSpecById in the loop
### Search before asking

In the Optimizing Plan phase, it is necessary to obtain the file's PartitionSpec. However, calling `getMixedTablePartitionSpecById` in a loop is a very expensive and unnecessary operation。

https://github.com/apache/amoro/blob/678b43c85347eb69d61e4f7ca016cb63d2ae56e4/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/plan/OptimizingEvaluator.java#L119-L124

Especially in environments where Kerberos is enabled, repeatedly calling `org.apache.iceberg.BaseTable#specs` can incur lock overhead.

- [ ] 

### How should we improve?

I want to place the code for obtaining PartitionSpec inside the TableFileScanHelper interface。
```java
...
  PartitionSpec partitionSpec = tableFileScanHelper.getSpec(fileScanResult.file().specId());
...
```

### Subtasks

_No response_

### Code of Conduct

7hong:

Close #3289 .

## How was this patch tested?

In my environment, there is a table with many partitions waiting to be merged. Each plan is very time-consuming. After optimization, the time-consuming is significantly reduced

image

### DIFF
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingEvaluator.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingEvaluator.java
@@ -32,7 +32,6 @@ import org.apache.amoro.shade.jackson2.com.fasterxml.jackson.annotation.JsonIgno
 import org.apache.amoro.table.KeyedTableSnapshot;
 import org.apache.amoro.table.MixedTable;
 import org.apache.amoro.table.TableSnapshot;
-import org.apache.amoro.utils.MixedTableUtil;
 import org.apache.iceberg.PartitionSpec;
 import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.SnapshotSummary;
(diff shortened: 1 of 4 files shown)


7hong:
@zhoujinsong @majin1102  Do you have time to review it? Thanks

baiyangtx:
image

Will the old code trigger a reload of TableMetadata?

It seems that if the reloading of TableMetadata is not triggered, the performance of the new code is the same as that of the old code.

7hong:
@baiyangtx 
Yes, the old code will refresh the TableMetadata. Especially when calling the `getUGI` method in an environment with Kerberos enabled, it will enter a synchronous blocking state.

czy006 (approved):
If there are Fire flame Graph or some unit test cases, it would be best.

7hong:
@czy006 
Flame graph before optimization：
image

Flame graph after optimization:

image

zhoujinsong (approved):
Based on the information provided by the flame graph, `Table#specs` does not reload table metadata. From its internal implementation, this process is controlled by the `shouldRefresh` variable. However, this call does trigger the Kerberos relogin check logic, which is unnecessary.

So, I am +1 with this improvement.

PR #3313, opened 2024-10-31
[AMORO-3312]Separate the shared iceberg-worker-pool for planning and committing

### LINKED ISSUE #3312: [Improvement]: Separate the shared iceberg-worker-pool for planning and committing
### Search before asking

Currently, the plan and commit operations share the same iceberg worker thread pool. When a large table with many small files is being planned, it can occupy a large number of threads. This may cause other tables that need to commit to remain in a committing state due to a lack of available worker threads. Therefore, it is necessary to separate them.

### How should we improve?

_No response_

### Subtasks

_No response_

### Code of Conduct

rfyu:
…

Close #3312.

## Brief change log

-

## How was this patch tested?

## Documentation


amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
```diff
@@ -449,10 +450,11 @@ private Map<String, Object> initEnvConfig() {
     private void setIcebergSystemProperties() {
       int workerThreadPoolSize =
           Math.max(
-              Runtime.getRuntime().availableProcessors(),
+              Runtime.getRuntime().availableProcessors() / 2,
```
XBaith, line 453: Why change the thread pool size to half the maximum available processors?

amoro-ams/src/main/java/org/apache/amoro/server/utils/IcebergThreadPools.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    LOG.info(
+        "init iceberg thread pool success, planningExecutor size:{},commitExecutor size:{}",
+        planningThreadPoolSize,
+        commitThreadPoolSize);
+  }
+
+  public static ExecutorService getPlanningExecutor() {
+    return Objects.requireNonNull(planningExecutor, "planningExecutor must not null");
```
XBaith, line 69: should we init thread pool if it is null

XBaith (commented):
Thanks for quick response!

XBaith:
Need to init `IcebergPools` in unit tests

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
rfyu: Thanks for the review. Because there used to be only one thread pool and now there are two. Of course, even if no changes are made, it won't cause significant issues.

rfyu:
> Need to init `IcebergPools` in unit tests

Thanks, I'll fix it.

zhoujinsong:
BTW, there are some changes that should be made to the mixed format tables.
I can open another PR to complete it after this change if you need it.

rfyu:
> @rfyu Thanks for the contribution!
> 
> AFAIK, there will be 3 iceberg working pools after the change:
> 
> * One for the optimizing plan
> * One for the optimizing commit
> * And One for other operations like web request
> 
> And they can be configured by different configurations. Am I right？

You are right.But currently, only the configurations related to the optimizing plan and commit operations have been added, the other operations are still using the default configurations.

rfyu:
> BTW, there are some changes that should be made to the mixed format tables. I can open another PR to complete it after this change if you need it.

Thank you for your help. I'm indeed not very familiar with the mixed format.

zhoujinsong:
> You are right.But currently, only the configurations related to the optimizing plan and commit operations have been added, the other operations are still using the default configurations.

If so, we may discuss the configuration names to make them more understandable. IMO, they can be:

- self-optimizing.plan-manifest-io-thread-count
- self-optimizing.commit-manifest-io-thread-count
- table-manifest-io.thread-count

HDYT?

rfyu:
> > You are right.But currently, only the configurations related to the optimizing plan and commit operations have been added, the other operations are still using the default configurations.
> 
> If so, we may discuss the configuration names to make them more understandable. IMO, they can be:
> 
> * self-optimizing.plan-manifest-io-thread-count
> * self-optimizing.commit-manifest-io-thread-count
> * table-manifest-io.thread-count
> 
> HDYT?

I agree.

zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #3339, opened 2024-11-21
[AMORO-3330] Reducing redundant 'databaseExists' in 'CommonUnifiedCatalog'

### LINKED ISSUE #3330: [Improvement]: Optimize excessive database access in CommonUnifiedCatalog
### Search before asking

Currently, some `databaseExists` checks in `CommonUnifiedCatalog` is redundant because the logic can be coverd by the subsequent logic(such as `CommonUnifiedCatalog#createDatabase`, `CommonUnifiedCatalog#dropDatabase`) , and the `databaseExists` checks will make a call to the external catalog(eg HiveMetaStore), which may be time-consuming.

### How should we improve?

Remove redundant 'databaseExists'  checks in `CommonUnifiedCatalog`  to improve performance.

### Subtasks

_No response_

### Code of Conduct

zhangwl9:

Close #3330
## Brief change log

1）the 'databaseExists' call in 'CommonUnifiedCatalog#createDatabase' and 'CommonUnifiedCatalog#dropDatabase' can be suppressed by the subsequent logic, so we delete redundant 'databaseExists' in those functions.
2) the 'databaseExists' in  'CommonUnifiedCatalog#listTables' can not be removed directly, because the 'HudiHadoopCatalog#listTables' and 'MixedCatalog#listTables' methods called in 'FormatCatalog#listTables' do not explicitly throw NoSuchDatabaseException. We add exception handling for NoSuchDatabaseException in these functions to eliminate the redundancy of 'databaseExists' in 'CommonUnifiedCatalog#listTables'.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-format-hudi/src/main/java/org/apache/amoro/formats/hudi/HudiHadoopCatalog.java
+++ b/amoro-format-hudi/src/main/java/org/apache/amoro/formats/hudi/HudiHadoopCatalog.java
@@ -21,6 +21,7 @@ package org.apache.amoro.formats.hudi;
 import org.apache.amoro.AmoroTable;
 import org.apache.amoro.DatabaseNotEmptyException;
 import org.apache.amoro.FormatCatalog;
+import org.apache.amoro.NoSuchDatabaseException;
 import org.apache.amoro.NoSuchTableException;
 import org.apache.amoro.properties.CatalogMetaProperties;
 import org.apache.amoro.shade.guava32.com.google.common.base.Preconditions;
@@ -155,7 +156,12 @@ public class HudiHadoopCatalog implements FormatCatalog {
         () -> {
           FileSystem fs = fs();
           Path databasePath = new Path(warehouse, database);
-          FileStatus[] items = fs.listStatus(databasePath);
+          FileStatus[] items;
+          try {
+            items = fs.listStatus(databasePath);
+          } catch (FileNotFoundException e) {
+            throw new NoSuchDatabaseException("Database: " + database + " is not exists", e);
+          }
           if (items == null || items.length == 0) {
             return Lists.newArrayList();
           }
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedCatalog.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedCatalog.java
@@ -20,6 +20,7 @@ package org.apache.amoro.formats.mixed;
 
 import org.apache.amoro.AmoroTable;
 import org.apache.amoro.FormatCatalog;
+import org.apache.amoro.NoSuchDatabaseException;
 import org.apache.amoro.TableFormat;
 import org.apache.amoro.mixed.MixedFormatCatalog;
 import org.apache.amoro.table.MixedTable;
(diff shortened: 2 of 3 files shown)


amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedCatalog.java
```diff
@@ -76,9 +77,13 @@ public AmoroTable<?> loadTable(String database, String table) {
...
   public List<String> listTables(String database) {
-    return catalog.listTables(database).stream()
-        .map(TableIdentifier::getTableName)
-        .collect(Collectors.toList());
+    try {
+      return catalog.listTables(database).stream()
+          .map(TableIdentifier::getTableName)
+          .collect(Collectors.toList());
+    } catch (NoSuchDatabaseException e) {
+      throw new NoSuchDatabaseException(e, database);
```
klion26, line 85: Why do we catch and retry the original exception here

amoro-format-hudi/src/main/java/org/apache/amoro/formats/hudi/HudiHadoopCatalog.java
```diff
@@ -155,6 +156,9 @@ public List<String> listTables(String database) {
         () -> {
           FileSystem fs = fs();
           Path databasePath = new Path(warehouse, database);
+          if (!fs.exists(databasePath)) {
```
klion26, line 159: As the Javadoc of `FormatCatalog#listTables` says that it will throw `NoSuchDatabaseException` when the database does not exist, I think adding this here is fine.

Can we use the following `fs.listStatus()` to check if the file exists?

klion26 (commented):
Thanks for the contribution, left some inline comments, please take a look when you're free.

amoro-format-hudi/src/main/java/org/apache/amoro/formats/hudi/HudiHadoopCatalog.java
zhangwl9: fs.listStatus(Path f) throws FileNotFoundException when the path does not exist，

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedCatalog.java
zhangwl9: The 'listTables' called in BasicMixedCatalog、BasicMixedIcebergCatalog、InternalMixedIcebergCatalog , etc., will throw other exceptions  such ad IllegalStateException not NoSuchDatabaseException.

zhangwl9: > Why do we catch and retry the original exception here

Thanks for your question. The 'listTables' called in BasicMixedCatalog、BasicMixedIcebergCatalog、InternalMixedIcebergCatalog , etc., will throw other exceptions  such as IllegalStateException not NoSuchDatabaseException.

zhangwl9: > Why do we catch and retry the original exception here

We can not catch this exception here. I have changed it.

zhoujinsong (approved):
LGTM.

Thanks for the contribution! @zhangwl9 
Thanks for the review! @klion26

PR #3343, opened 2024-11-25
[hotfix]Fix the incorrent paramater transmission of PageNum and PageSize

zhangwl9:

Close #xxx.

## Brief change log

We need to to transmit parameters of pageNum and pageSize in 'PageHelper#startPage' and the TableMetaMapper#selectTableRuntimesForOptimizerGroup,OptimizingMapper#selectOptimizingProcesses.

We correct the inconsistent problem.

## How was this patch tested?

## Documentation

### DIFF
--- a/LICENSE-binary
+++ b/LICENSE-binary
@@ -345,6 +345,11 @@ org.objenesis:objenesis
 org.mybatis:mybatis
 org.lz4:lz4-java
 org.jetbrains.kotlin:kotlin-*
+./charts/amoro/templates/common/_names.tpl
+./charts/amoro/templates/common/_capabilities.tpl
+./charts/amoro/templates/common/_tplvalues.tpl
+./charts/amoro/templates/common/_ingress.tpl
+./charts/amoro/templates/common/_labels.tpl
 
 The BSD License
 --------------
--- a/amoro-ams/src/main/java/org/apache/amoro/server/dashboard/MixedAndIcebergTableDescriptor.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/dashboard/MixedAndIcebergTableDescriptor.java
@@ -524,9 +524,7 @@ public class MixedAndIcebergTableDescriptor extends PersistentBase
                       tableIdentifier.getDatabase(),
                       tableIdentifier.getTableName(),
                       type,
-                      status,
-                      offset,
-                      limit));
+                      status));
       PageInfo<OptimizingProcessMeta> pageInfo = new PageInfo<>(processMetaList);
       total = (int) pageInfo.getTotal();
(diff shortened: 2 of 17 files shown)


zhangwl9:
@klion26  @czy006 Could you please take a look at this when you're free, thanks

klion26:
@zhangwl9 thanks for the contribution, the change here LGTM.

As there was some ut want to cover the logic, so could you please help to check
1. if the logic here in the master is right or not(if the code logic in the master branch can give the right result, maybe it is because the parameter in `PageHelper.startPage(pageNumber, limit, true)` -- line 517 has a higher priority than the parameter in the mapper)
2. if the logic here in the master is wrong, could you please please add more ut to cover the logic

baiyangtx (commented):
After introducing github.PageHelper, do we still need to pass the pageNumber and limit parameters to the mapper? 

It seems that these two parameters are not required in the process of constructing SQL.

zhangwl9:
> @zhangwl9 thanks for the contribution, the change here LGTM.
> 
> As there was some ut want to cover the logic, so could you please help to check
> 
> 1. if the logic here in the master is right or not(if the code logic in the master branch can give the right result, maybe it is because the parameter in `PageHelper.startPage(pageNumber, limit, true)` -- line 517 has a higher priority than the parameter in the mapper)
> 2. if the logic here in the master is wrong, could you please help to enhance the exist ut to cover the logic

Thank you very much for your advice. I have tested and found the following:

1) The parameter (pageNumber,limit) in PageHelper.startPage(pageNumber, limit, true) -- line 517 has a higher priority than the parameter(pageNum, pageSize) in the mapper, as shown in figure 
image


2) When building the mapper's SQL statement, if the parameters pageNum and pageSize are removed, amoro displays normally, as shown in the figure.
image

Therefore, pageNum and pageSize are not needed during SQL construction. I have already fixed it.

 3) Two unit tests involving mappers: TestIcebergServerTableDescriptor#testOptimizingPorcess and TestDefaultOptimizingService#testGetRuntimes can both cover the results of PageHelper and PageInfo, therefore, no new unit tests need to be added.

zhangwl9:
> After introducing github.PageHelper, do we still need to pass the pageNumber and limit parameters to the mapper?
> 
> It seems that these two parameters are not required in the process of constructing SQL.
@baiyangtx  Thank you very much for your advice. You are right. I have fix it.

klion26 (approved):
Thanks for the fix, LGTM

klion26:
@zhangwl9 Seems there are some tests failed, could you please help to check it, thanks.

czy006 (approved):
LGTM

PR #3345, opened 2024-11-27
[Improvement] expose 'http-server.rest-auth-type' variable in Helm chart

JiHyunSong:

Close #xxx.

## Brief change log

- export http-server.rest-auth-type variable in helm chart value.

## How was this patch tested?

## Documentation

### DIFF
--- a/charts/amoro/templates/amoro-configmap.yaml
+++ b/charts/amoro/templates/amoro-configmap.yaml
@@ -52,7 +52,7 @@ data:
           bind-port: {{ .Values.server.optimizing.port }}
 
       http-server:
-        rest-auth-type: {{ .Values.amoroConf.ams.restAuthType }}
+        rest-auth-type: {{ .Values.server.rest.restAuthType }}
         bind-port: {{ .Values.server.rest.port }}
 
       refresh-external-catalogs:
--- a/charts/amoro/templates/amoro-configmap.yaml
+++ b/charts/amoro/templates/amoro-configmap.yaml
@@ -52,6 +52,7 @@ data:
           bind-port: {{ .Values.server.optimizing.port }}
 
       http-server:
+        rest-auth-type: {{ .Values.amoroConf.ams.restAuthType }}
         bind-port: {{ .Values.server.rest.port }}
 
       refresh-external-catalogs:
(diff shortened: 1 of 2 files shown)


baiyangtx (approved):
LGTM

charts/amoro/values.yaml
```diff
@@ -142,6 +142,7 @@ amoroConf:
   ams:
     adminUsername: admin
     adminPassword: admin
+    restAuthType: token
```
XBaith, line 145: It should under `server.rest`

charts/amoro/templates/amoro-configmap.yaml
```diff
@@ -52,6 +52,7 @@ data:
           bind-port: {{ .Values.server.optimizing.port }}
 
       http-server:
+        rest-auth-type: {{ .Values.amoroConf.ams.restAuthType }}
```
XBaith, line 55: it is a config in AMS REST Serivce, so `rest-auth-type: {{ .Values.server.rest.restAuthType }}` is better option?

JiHyunSong: @XBaith 
I currently see the `server.rest` variables in this Helm chart as being more related to Kubernetes service or ingress configurations rather than application settings. On the other hand, I consider the configurations under `ams.` to be application-specific, which is why I placed them there. I also think this might be a part of the chart worth refactoring in the future..

Since there are differing opinions on this, I wanted to share my perspective. Could you please provide some guidance on this PR?

XBaith: The reason I prefer moving them under `server.rest` is that we already have the property `HTTP_SERVER_REST_AUTH_TYPE` configured as part of the HTTP REST service. It might be better to align with the original design. What do you think?

XBaith: Hi @baiyangtx, please let me know if you have any questions.
I will merge this if there is no more comments.

zhoujinsong (approved):
LGTM.

zhoujinsong:
Thanks for the contribution! @JiHyunSong 
Thanks for the review! @baiyangtx @XBaith

PR #3358, opened 2024-12-09
[AMORO-1604] Change default metric mode for eventTimeField

### LINKED ISSUE #1604: [Improvement]: When manually setting table.event-time-field to a field, it should widen its data file mertic length
### Search before asking

The current default length of the iceberg table for metadata metric is 'write.metadata.metrics.default' = 'truncate(16)'. meaning that data larger than 16 will have a loss of precision.
When we set table.event-time-field we often choose a time field in the format yyyy-MM-dd HH:mm:ss, obviously this string is longer than 16, then there will be a loss of precision when parsing the upperbounds and lowerbounds of the datafile.
From the user's point of view, most of the cases will not be able to change 'write.metadata.metrics.default' to a larger value or to 'full', so it is necessary to help the user to set this configuration.

### How should we improve?

AMS should set global default configuration such as 'write.metadata.metrics.default' to 'truncate(32)' when creating a table, or just for the fields set in the table.event-time-field configuration mertric changes

### Subtasks

_No response_

### Code of Conduct

MarigWeizhi:

Close #1604.

## Brief change log

- When creating a table, if the user sets `table.event-time-field`, `write.metadata.metrics.column.xxx` will be set to `truncate(32)` to prevent content like yyyy-MM-dd HH:mm:ss from being truncated by the default `truncate(16)`

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/internal/InternalIcebergCreator.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/internal/InternalIcebergCreator.java
@@ -28,13 +28,17 @@ import org.apache.amoro.properties.MetaTableProperties;
 import org.apache.amoro.server.table.TableMetadata;
 import org.apache.amoro.server.utils.InternalTableUtil;
 import org.apache.amoro.shade.guava32.com.google.common.base.Preconditions;
+import org.apache.amoro.table.TableProperties;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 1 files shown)


zhoujinsong:
Thanks for the contribution!

The `InternalIcebergCreator` works only for internal catalogs.
We can control the property settings for tables in the internal catalog or mixed format tables, but hard to control Iceberg tables in the external catalogs.

PR #3363, opened 2024-12-10
[hotfix] Fix drop database not working in terminal

MarigWeizhi:

Close #xxx.

## Brief change log

- Added the `dropNamespace` to support cascade or non-cascade in `SparkUnifiedCatalogBase` for spark 3.3
- spark 3.2 will use false by default

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-format-mixed/amoro-mixed-spark/v3.3/amoro-mixed-spark-3.3/src/main/java/org/apache/amoro/spark/SparkUnifiedCatalog.java
+++ b/amoro-format-mixed/amoro-mixed-spark/v3.3/amoro-mixed-spark-3.3/src/main/java/org/apache/amoro/spark/SparkUnifiedCatalog.java
@@ -19,6 +19,7 @@
 package org.apache.amoro.spark;
 
 import org.apache.amoro.TableFormat;
+import org.apache.amoro.TableIDWithFormat;
 import org.apache.spark.sql.catalyst.analysis.NoSuchFunctionException;
 import org.apache.spark.sql.catalyst.analysis.NoSuchNamespaceException;
 import org.apache.spark.sql.catalyst.analysis.NoSuchTableException;
@@ -30,6 +31,8 @@ import org.apache.spark.sql.connector.catalog.Table;
 import org.apache.spark.sql.connector.catalog.TableCatalog;
 import org.apache.spark.sql.connector.catalog.functions.UnboundFunction;
 import org.apache.spark.sql.connector.iceberg.catalog.ProcedureCatalog;
+
+import java.util.List;
 
 public class SparkUnifiedCatalog extends SparkUnifiedCatalogBase
     implements TableCatalog, SupportsNamespaces, ProcedureCatalog, FunctionCatalog {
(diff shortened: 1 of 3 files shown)


huyuanfeng2018 (commented):
Should we add a test to guard against future regression?

MarigWeizhi:
> Should we add a test to guard against future regression?

Added test, PTAL

amoro-format-mixed/amoro-mixed-spark/v3.3/amoro-mixed-spark-3.3/src/main/java/org/apache/amoro/spark/SparkUnifiedCatalog.java
```diff
@@ -86,7 +89,20 @@ public UnboundFunction loadFunction(Identifier ident) throws NoSuchFunctionExcep
   @Override
   public boolean dropNamespace(String[] namespace, boolean cascade)
       throws NoSuchNamespaceException, NonEmptyNamespaceException {
-    return false;
+    String database = namespaceToDatabase(namespace);
+    if (!unifiedCatalog.databaseExists(database)) {
+      throw new NoSuchNamespaceException(namespace);
+    }
+    List<TableIDWithFormat> tables = unifiedCatalog.listTables(database);
+    if (!tables.isEmpty() && !cascade) {
```
huyuanfeng2018, line 97: Why not drop all tables first and then drop the database?

I think should this logic be implemented directly by the parent (spark-common).
We already have concrete logic in  `SparkUnifiedCatalogBase.dropNamespace`.
so，remove Override is enough.

WDYT?

huyuanfeng2018 (commented):
Thanks for the PR!

amoro-format-mixed/amoro-mixed-spark/v3.3/amoro-mixed-spark-3.3/src/main/java/org/apache/amoro/spark/SparkUnifiedCatalog.java
zhoujinsong: I agreed that we may reuse some implementation in the parent class. But I think it is dangerous to delete all tables by default in Spark 3.2 implementation. So wee may:

- Add a method in the base class to support dropping namespace cascade or not.
- Drop namespace by default with false for Spark 3.2

MarigWeizhi:
Thanks for your review @zhoujinsong @huyuanfeng2018   
I updated the code, I implemented the method to support cascade or not in base instead of spark 3.2.

amoro-format-mixed/amoro-mixed-spark/v3.3/amoro-mixed-spark-3.3/src/main/java/org/apache/amoro/spark/SparkUnifiedCatalog.java
```diff
@@ -85,8 +85,8 @@ public UnboundFunction loadFunction(Identifier ident) throws NoSuchFunctionExcep
    */
   @Override
   public boolean dropNamespace(String[] namespace, boolean cascade)
```
huyuanfeng2018, line 87: Just delete this method, WDYT?

huyuanfeng2018 (commented):
just a nit comment

amoro-format-mixed/amoro-mixed-spark/v3.3/amoro-mixed-spark-3.3/src/main/java/org/apache/amoro/spark/SparkUnifiedCatalog.java
MarigWeizhi: deleted

huyuanfeng2018 (approved):
LGTM.

zhoujinsong:
Thanks for the contribution! @MarigWeizhi 
Thanks for the review! @huyuanfeng2018

PR #3383, opened 2025-01-02
[AMORO-3383] fix: unified catalog should support iceberg  native create_changelog_view procedure

Aireed:

The unified catalog does not support the Iceberg change metadata table syntax, which makes it impossible to support the create_changelog_view stored procedure.

```
spark-sql> CALL spark_catalog.system.create_changelog_view(
         >   table => 'iceberg_v7.iceberg_overwrite',
         >   options => map('start-snapshot-id','486064778759746866','end-snapshot-id', '1216062697090990172'),
         >   changelog_view => 'my_changelog_view_4'
         > );
ANTLR Tool version 4.9.3 used for code generation does not match the current runtime version 4.8ANTLR Runtime version 4.9.3 used for parser compilation does not match the current runtime version 4.8ANTLR Tool version 4.9.3 used for code generation does not match the current runtime version 4.8ANTLR Runtime version 4.9.3 used for parser compilation does not match the current runtime version 4.8
Error in query: spark_catalog requires a single-part namespace, but got [iceberg_v7, iceberg_overwrite]
```

## Brief change log

- add `changes`  subtablename  in  class **IcebergSparkFormat**

**Apart from this, I haven't found any other situations that need to be addressed. As the version upgrades, it will be necessary to align with the list of subTables supported by Iceberg.**

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

Thanks for the contriburion!

PR #3394, opened 2025-01-03
[AMORO-3393]: A stateless catalog manager implement

### LINKED ISSUE #3393: [Subtask]: A stateless CatalogManager implement.
### Search before asking

### Description

Refact and implement a stateless catalog management related interface 

### Parent issue

#3346

### Code of Conduct

baiyangtx:

Close #3393.

This is a subtask to implement a stateless RESTController. 

## Brief change log

- Rename the **CatalogService** interface to **CatalogManager**, indicating that the interface provides stateless services.
- Provide a stateless implementation called **DefaultCatalogManager**.
- The **TableService** interface will no longer inherit from **CatalogManager**.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
@@ -1,0 +1,150 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.amoro.server.catalog;
+
+import org.apache.amoro.api.CatalogMeta;
+import org.apache.amoro.config.Configurations;
+import org.apache.amoro.exception.AlreadyExistsException;
+import org.apache.amoro.exception.IllegalMetadataException;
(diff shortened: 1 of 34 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+    if (serverCatalog.isInternal()) {
+      return (InternalCatalog) serverCatalog;
+    }
+    throw new ObjectNotExistsException("Catalog " + catalogName + " is not internal catalog");
+  }
+
+  @Override
+  public List<ExternalCatalog> getExternalCatalogs() {
+    return listCatalogMetas().stream()
```
zhoujinsong, line 110: We can improve the performance of this method by reducing some database calls.
Now it will select all the catalog metadata and then select each one of them individually, which should be improved.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  @Override
+  public boolean catalogExist(String catalogName) {
+    return getCatalogMetaOptional(catalogName).isPresent();
+  }
+
+  @Override
+  public ServerCatalog getServerCatalog(String catalogName) {
+    Optional<CatalogMeta> catalogMeta = getCatalogMetaOptional(catalogName);
```
zhoujinsong, line 82: Catalog metadata changes are infrequent. Should we control the minimum query interval to avoid accessing the database too frequently?

zhoujinsong (commented):
@baiyangtx 
Thanks for the contribution!
The PR looks good to me in general.

But I have a concern about the performance, you may want to check.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
baiyangtx: Cache will lead to consistency issues, and unit tests will also fail. 

I have optimized this part, and now each call only accesses the database once.

amoro-ams/src/main/java/org/apache/amoro/server/catalog/DefaultCatalogManager.java
baiyangtx: fixed

zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #3396, opened 2025-01-06
[AMORO-3335] Add interface ConfigShade to support encryption of sensitive configuration items and provide a base64 encoding implementation

### LINKED ISSUE #3335: [Improvement]: Add support for using encrypted passwords in configurations
### Search before asking

Currently, the login password for the admin user and the connection password for the mysql/postgresql databases are set in plaintext in the configuration file, which may be a security risk. To avoid the use of plaintext passwords, we would like to add support for using encrypted passwords in the configuration file.

### How should we improve?

_No response_

### Subtasks

- [x] #3336
- [x] #3337
- [ ] #3417

### Code of Conduct

Jzjsnow:

When we start Amoro Management Service , we need to set configuration items in plaintext in the config file, including sensitive configurations such as admin-password and the passwords for connecting to databases (e.g., mysql, pg, etc.), which may be a security risk. To avoid the use of plaintext passwords, we provide an interface (`ConfigShade`) by implementing which developers can customize the decryption method themselves.

We also provide an implementation for base64 encoding first, not only as an example implementation of the interface, but also to solve the current problem of plaintext passwords.

Close #3335.

## Brief change log

- Add interface `ConfigShade` to the `amoro-common` module to provide the ability to customize decryption of sensitive options in the config file
- Provide an implementation `org.apache.amoro.config.shade.impl.Base64ConfigShade` for base64 encoding 

## How was this patch tested?

## Documentation

### How to use
Using the base64 implementation as an example, the following shows how to use a configuration file with sensitive items encrypted:

1. Add two new options `shade.identifier` and `shade.sensitive-keywords` to the `ams` part in `config.yaml` to specify the encryption algorithm and the encrypted sensitive keywords.
2. Replace the plaintext of the sensitive items specified in `shade.sensitive-keywords` with the encrypted ciphertext.
3. Restart the ams service.

Example config file (partial):
```
ams:
  admin-username: admin
  admin-password: YWRtaW4=
  server-bind-host: "0.0.0.0"
  server-expose-host: "127.0.0.1"

  shade:
    identifier: base64
    sensitive-keywords: admin-password;database.password

  database:
    type: mysql
    jdbc-driver-class: com.mysql.cj.jdbc.Driver
    url: jdbc:mysql://127.0.0.1:3306/amoro?useUnicode=true&characterEncoding=UTF8&autoReconnect=true&useAffectedRows=true&allowPublicKeyRetrieval=true&useSSL=false
    username: root
    password: cGFzc3dvcmQ=

   ...

```

### How to customize the encryption algorithm
To use a user-defined encryption algorithm, we expect the developer to provide a dependency package that implements the `ConfigShade` interface. 
```
/**
 * The interface that provides the ability to decrypt {@link
 * org.apache.amoro.config.Configurations}.
 */
public interface ConfigShade {
  /**
   * Initializes the custom instance using the service configurations.
   *
   * This method can be useful when decryption requires an external file (e.g. a key file)
   * defined in the service configs.
   */
  default void initialize(Configurations serviceConfig) throws Exception {}

  /**
   * The unique identifier of the current interface, used it to select the correct {@link
   * ConfigShade}.
   */
  String getIdentifier();

  /**
   * Decrypt the content.
   *
   * @param content The content to decrypt
   */
  String decrypt(String content);
}
```
In it, the method `getIdentifier()` can be called to get the unique identifier of the algorithm, which is used to configure the `shade.identifier`, and the method `decrypt(String content)` can be used to decrypt the input cipher text.

### DIFF
--- a/amoro-ams/src/test/java/org/apache/amoro/server/config/TestConfigShade.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/config/TestConfigShade.java
@@ -44,16 +44,17 @@ public class TestConfigShade {
   private static final String USERNAME = "admin";
   private static final String PASSWORD = "password";
 
-  private final String base64ConfigShadeIdentifier = new Base64ConfigShade().getIdentifier();
+  private static final String BASE64_CONFIG_SHADE_IDENTIFIER =
+      new Base64ConfigShade().getIdentifier();
 
   @Test
   public void testDecryptOptions() {
     String encryptUsername = getBase64EncodedText(USERNAME);
     String encryptPassword = getBase64EncodedText(PASSWORD);
     String decryptUsername =
-        ConfigShadeUtils.decryptOption(base64ConfigShadeIdentifier, encryptUsername);
+        ConfigShadeUtils.decryptOption(BASE64_CONFIG_SHADE_IDENTIFIER, encryptUsername);
     String decryptPassword =
-        ConfigShadeUtils.decryptOption(base64ConfigShadeIdentifier, encryptPassword);
+        ConfigShadeUtils.decryptOption(BASE64_CONFIG_SHADE_IDENTIFIER, encryptPassword);
     Assertions.assertEquals(decryptUsername, USERNAME);
     Assertions.assertEquals(decryptPassword, PASSWORD);
   }
--- a/amoro-ams/src/test/java/org/apache/amoro/server/config/TestConfigShade.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/config/TestConfigShade.java
@@ -51,6 +51,9 @@ public class TestConfigShade {
   public void testDecryptOptions() {
     String encryptUsername = getBase64EncodedText(USERNAME);
     String encryptPassword = getBase64EncodedText(PASSWORD);
+    Assertions.assertEquals(encryptUsername, "YWRtaW4=");
+    Assertions.assertEquals(encryptPassword, "cGFzc3dvcmQ=");
+
     String decryptUsername =
         ConfigShadeUtils.decryptOption(BASE64_CONFIG_SHADE_IDENTIFIER, encryptUsername);
     String decryptPassword =
@@ -110,7 +113,7 @@ public class TestConfigShade {
           .put("database.jdbc-driver-class", "com.mysql.cj.jdbc.Driver")
           .put(
               "database.url",
-              "jdbc:mysql://127.0.0.1:3306/amoro-jzjsnow?useUnicode=true&characterEncoding=UTF8&autoReconnect=true&useAffectedRows=true&allowPublicKeyRetrieval=true&useSSL=false")
+              "jdbc:mysql://127.0.0.1:3306/amoro?useUnicode=true&characterEncoding=UTF8&autoReconnect=true&useAffectedRows=true&allowPublicKeyRetrieval=true&useSSL=false")
           .put("database.username", "root")
           .put("database.password", "password")
           .build();
@@ -127,7 +130,7 @@ public class TestConfigShade {
           .put("database.jdbc-driver-class", "com.mysql.cj.jdbc.Driver")
           .put(
               "database.url",
-              "jdbc:mysql://127.0.0.1:3306/amoro-jzjsnow?useUnicode=true&characterEncoding=UTF8&autoReconnect=true&useAffectedRows=true&allowPublicKeyRetrieval=true&useSSL=false")
+              "jdbc:mysql://127.0.0.1:3306/amoro?useUnicode=true&characterEncoding=UTF8&autoReconnect=true&useAffectedRows=true&allowPublicKeyRetrieval=true&useSSL=false")
           .put("database.username", "root")
           .put("database.password", "password")
           .build();
--- a/amoro-ams/src/test/java/org/apache/amoro/server/config/TestConfigShade.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/config/TestConfigShade.java
@@ -20,6 +20,7 @@ package org.apache.amoro.server.config;
 
 import static org.apache.amoro.server.AmoroServiceContainer.expandConfigMap;
 
+import org.apache.amoro.config.shade.impl.Base64ConfigShade;
 import org.apache.amoro.config.shade.utils.ConfigShadeUtils;
 import org.apache.amoro.server.AmoroManagementConf;
 import org.apache.amoro.shade.guava32.com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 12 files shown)


amoro-common/src/main/java/org/apache/amoro/config/shade/ConfigShade.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.amoro.config.Configurations;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The interface that provides the ability to decrypt {@link
+ * org.apache.amoro.config.Configurations}.
+ */
+public interface ConfigShade {
+  Logger LOG = LoggerFactory.getLogger(ConfigShade.class);
```
klion26, line 30: seems the `LOG` here is no more used

amoro-common/src/main/java/org/apache/amoro/config/shade/utils/ConfigShadeUtils.java
```diff
@@ -0,0 +1,139 @@
...
+          .stringType()
+          .defaultValue("default")
+          .withDescription(
+              "The identifier of the encryption method for decryption. Defaults to \"default\", indicating no encryption");
+
+  public static final ConfigOption<List<String>> SHADE_SENSITIVE_KEYWORDS =
+      ConfigOptions.key("shade.sensitive-keywords")
+          .stringType()
+          .asList()
+          .defaultValues("admin-password", "database.password")
```
klion26, line 56: seems the the key here is only located in `ams` part in the configuration, not sure if this needs to be apply the whole config

amoro-common/src/main/java/org/apache/amoro/config/shade/utils/ConfigShadeUtils.java
```diff
@@ -0,0 +1,139 @@
...
+          .withDescription(
+              "A semicolon-separated list of keywords for the configuration items to be decrypted.");
+
+  static {
+    ServiceLoader<ConfigShade> serviceLoader = ServiceLoader.load(ConfigShade.class);
+    Iterator<ConfigShade> it = serviceLoader.iterator();
+    it.forEachRemaining(
+        configShade -> {
+          CONFIG_SHADES.put(configShade.getIdentifier(), configShade);
```
klion26, line 65: Do we need to log the relationship for the identifier and the configShade here? I assume the shade loaded here would not a big number.

klion26 (commented):
@Jzjsnow thanks for the contribution, I left some comments, please have a look when you're free.

amoro-common/src/main/java/org/apache/amoro/config/shade/utils/ConfigShadeUtils.java
```diff
@@ -0,0 +1,139 @@
...
+        (key, value) -> configShade.decrypt(value.toString());
+
+    Preconditions.checkArgument(
+        !configMap.isEmpty(), "Miss ams config! Please check the config file.");
+
+    for (String sensitiveOption : sensitiveOptions) {
+      try {
+        configMap.computeIfPresent(sensitiveOption, processFunction);
+      } catch (Exception e) {
+        LOG.error("Failed to decrypt sensitive option {}:", sensitiveOption, e);
```
klion26, line 112: do we need to throw exception here if no key found

amoro-common/src/main/java/org/apache/amoro/config/shade/utils/ConfigShadeUtils.java
```diff
@@ -0,0 +1,139 @@
...
+    it.forEachRemaining(
+        configShade -> {
+          CONFIG_SHADES.put(configShade.getIdentifier(), configShade);
+        });
+    LOG.info("Load config shade spi: {}", CONFIG_SHADES.keySet());
+  }
+
+  @VisibleForTesting
+  public static String decryptOption(String identifier, String content) {
+    ConfigShade configShade = CONFIG_SHADES.getOrDefault(identifier, DEFAULT_SHADE);
```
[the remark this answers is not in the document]
Jzjsnow: Yes, it is necessary to print the log. The method `decryptConfig` is called for normal use, and if the given identifier cannot be loaded, it will log ERROR and throw an IllegalStateException exception. Here, the method `decryptOption` is only used for unit testing.

amoro-ams/src/test/java/org/apache/amoro/server/config/TestConfigShade.java
```diff
@@ -0,0 +1,122 @@
...
+
+public class TestConfigShade {
+  private static final String USERNAME = "admin";
+  private static final String PASSWORD = "password";
+
+  @Test
+  public void testDecryptOptions() {
+    String encryptUsername = "YWRtaW4=";
+    String encryptPassword = "cGFzc3dvcmQ=";
+    String decryptUsername = ConfigShadeUtils.decryptOption("base64", encryptUsername);
```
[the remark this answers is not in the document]
Jzjsnow: I have changed the identifier string to a static final string with the value of `Base64ConfigShade.getIdentifier()`.

amoro-ams/src/test/java/org/apache/amoro/server/config/TestConfigShade.java
```diff
@@ -0,0 +1,122 @@
...
+import java.nio.file.Paths;
+import java.util.Map;
+
+public class TestConfigShade {
+  private static final String USERNAME = "admin";
+  private static final String PASSWORD = "password";
+
+  @Test
+  public void testDecryptOptions() {
+    String encryptUsername = "YWRtaW4=";
```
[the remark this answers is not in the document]
Jzjsnow: In the new commit, a `getBase64EncodedText` method is provided in the unit test to show the base64 encoding process.

amoro-common/src/main/java/org/apache/amoro/config/shade/utils/ConfigShadeUtils.java
Jzjsnow: I add the log to indicate the identifier of the `ConfigShade` implementation and the full name of the class.

amoro-common/src/main/java/org/apache/amoro/config/shade/utils/ConfigShadeUtils.java
Jzjsnow: Currently, the sensitive keywords are all in the `ams` part, so first I put the shade-related configurations under `ams` part and only decrypt the `ams` configurations . If in the future the containers also contain sensitive information, I think it may be necessary to put the shade-related configurations in a separate part, alongside the `ams` and `containers`.

czy006 (commented):
Thank you for your contribution. So, how should we handle this logic in the Helm Chart section

klion26 (commented):
@Jzjsnow Thanks for the contribution, LGTM. could you please file an issue to add the documentation, so that the users can know how to use this feature, thanks.

@czy006 could you please help to take another look at the k8s part.

Jzjsnow:
@czy006 Thanks for the heads up, I've changed the script in helm charts in the latest commit and verified it locally: 
Similar to the usage in `config.yaml`, users needs to fill in the options `amoroConf.shade.identifier` and `amoroConf.shade.sensitiveKeywords` in `charts/amoro/values.yaml` and replace the plaintext of the sensitive items specified in `sensitive-keywords` with the encrypted ciphertext. Then deploy it via the `helm install` command to make it work, as described in the official documentation.
If users want to use their customized encryption algorithm, they need to modify the amoro image prior to the above steps, as they need to upload their decryption dependency package to the image first.

zhoujinsong:
Hi，

Thanks for the contribution!

Can you help to add the configuration example into the `dist/src/main/amoro-bin/conf/config.yaml`?

Jzjsnow:
@zhoujinsong Thanks for the reminder! I've added the relevant parameters to the template `config.yaml` in the latest commit.

zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #3400, opened 2025-01-07
[improvement] encode for optimizerGroup value

ihadoop:

when optimizerGroup name has value like '/' , the search url will be  "http://127.0.0.1:1630/api/ams/v1/optimize/optimizerGroups/PROD/test/tables?dbSearchInput=&tableSearchInput=&page=1&pageSize=25". it will 404 for server. we should encode optimizerGroup.

when encode , it will be work well.
image

Close #xxx.

## Brief change log

-

## How was this patch tested?

## Documentation

 no

### DIFF
--- a/amoro-web/src/views/optimize/components/List.vue
+++ b/amoro-web/src/views/optimize/components/List.vue
@@ -91,7 +91,7 @@ async function getTableList() {
   try {
     loading.value = true
     const params = {
-      optimizerGroup: optimizerGroup.value || 'all',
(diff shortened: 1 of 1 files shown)


zhoujinsong:
@zhangmo8 Can you help to review this PR?

zhangmo8:
Except for special characters, In the path, encoding is a default behavior of the browser. If encoding is done manually, I'm not sure whether the backend can recognize it correctly.

If the back end can correctly retrieve the character `\t`. I think this change is great.

zhangmo8 (approved):
LG

PR #3409, opened 2025-01-17
[AMORO-3329] Add MemorySize for parsing memory value and unit

### LINKED ISSUE #3329: [Improvement]: Add MemorySize for memory value and unit
### Search before asking

Currently, we'll process memory value and unit manually(et resource configuration in optimizerGroup), and the unit only support `m`/`g`, if user configured with `mb`/`gb`, we'll get a wrong parsed value.

### How should we improve?

Support a `MemorySize` to parse the memory size with the unit, support the following values: `1`, `1k`, `1kb`, `1kibibytes`, etc.

**this new one can be compatible with the old one.**

### Subtasks

_No response_

### Code of Conduct

Jzjsnow:
Currently, when we use flink optimizer groups, the memory parameters `flink-conf.jobmanager.memory.process.size` and `flink-conf.taskmanager.memory.process.size` only support plain numeric strings or strings with `m/g` suffix. If we configure them with mb/gb suffix, we'll get wrong parsed results. And the default unit for plain numeric strings is `mb`, which is inconsistent with the default unit of byte in the flink native parameter and may cause inconvenience for users familiar with flink when using this parameter.

Close #3329.

## Brief change log

- Add a new class `MemorySize` to the `amoro-common` module for parsing flink's memory parameters, consistent with flink's native parsing logic.
- Update the values in the column `properties` of ams metadata tables `resource_group` and `resource` to automatically add the `MB` suffix to the flink jm and tm memory parameters that have been set and saved as plain numeric strings

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-common/src/main/java/org/apache/amoro/utils/MemorySize.java
+++ b/amoro-common/src/main/java/org/apache/amoro/utils/MemorySize.java
@@ -1,0 +1,410 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.amoro.utils;
+
+import static org.apache.amoro.shade.guava32.com.google.common.base.Preconditions.checkArgument;
+import static org.apache.amoro.utils.MemorySize.MemoryUnit.BYTES;
+import static org.apache.amoro.utils.MemorySize.MemoryUnit.GIGA_BYTES;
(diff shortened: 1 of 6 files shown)


amoro-common/src/main/java/org/apache/amoro/utils/MemorySize.java
```diff
@@ -0,0 +1,410 @@
...
+import java.util.stream.IntStream;
+
+/**
+ * MemorySize is a representation of a number of bytes, viewable in different units. Copied from
+ * Apache Flink.
+ *
+ * <h2>Parsing</h2>
+ *
+ * <p>The size can be parsed from a text expression. If the expression is a pure number, the value
+ * will be interpreted as bytes.
```
zhoujinsong, line 43: Could you add some information in the Java doc to tell where this file is copied from?
As I know, it was copied from the Flink project.

zhoujinsong (commented):

I think this is a good start to make our configuration value more readable.
We may improve our configuration format in other place, like the AMS configuration after this PR been merged.

I left a minor suggestion, you may want have a look!

amoro-common/src/main/java/org/apache/amoro/utils/MemorySize.java
Jzjsnow: Thanks for the suggestion, I've revised the Java doc in the latest commit to make the original source of the class more clear and explicit.

zhoujinsong (approved):
LGTM.

klion26 (approved):

This patch is back-compatible, but it contains some user behavior change, maybe we need to add this in the release notes, @Jzjsnow could you please create an issue to track the release notes update, thanks.

Apart the above, do we need to add some documentation about how to configure this in the documentation? if this is yes could you please add an issue to track it. thanks.

PR #3423, opened 2025-01-23
[AMORO-3418] Optimize ams configuration to support parsing of time and storage unit

### LINKED ISSUE #3418: [Improvement]: Optimize ams configuration to support parsing of time interval and storage related configuration items when both values and units are specified
### Search before asking

Currently, the ams configuration is inconsistent in the type of time interval related configuration items, some of which are of `Duration` type, such as `http-server.session-timeout`, and some of which are of `Long` type, such as `refresh-external-catalogs.interval`, which may be misused by the user.
In addition, storage-related parameters such as `thrift-server.max-message-size` only support `Long` type (default unit is byte). It would be more user-friendly to support user-specified units, as in the [Apache Flink configuration](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/deployment/config/#duration).

### How should we improve?

_No response_

### Subtasks

_No response_

### Code of Conduct

Jzjsnow:

The time-related configuration items in the current ams configuration are inconsistent in type, with both `Long` and `Duration` types. This results in users having to convert to milliseconds when using these `Long` types, which is not user friendly.

In addition, storage-related parameters such as thrift-server.max-message-size only support Long type (default unit is byte). It would be more user-friendly to support user-specified units, as in the [Apache Flink configuration](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/deployment/config/#duration).

Close #3418.

## Brief change log

- Adjust the data type of all time-related configuration items to `Duration`.
- Support for setting memory units for storage related items.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
@@ -81,10 +81,10 @@ public class AmoroManagementConf {
               "Sets the size of the worker pool. The worker pool limits the number of tasks concurrently processing "
                   + "manifests in the base table implementation across all concurrent commit operations.");
 
-  public static final ConfigOption<Long> REFRESH_EXTERNAL_CATALOGS_INTERVAL =
+  public static final ConfigOption<Duration> REFRESH_EXTERNAL_CATALOGS_INTERVAL =
       ConfigOptions.key("refresh-external-catalogs.interval")
-          .longType()
-          .defaultValue(3 * 60 * 1000L)
+          .durationType()
+          .defaultValue(Duration.ofMinutes(3))
           .withDescription("Interval to refresh the external catalog.");
 
   public static final ConfigOption<Integer> REFRESH_EXTERNAL_CATALOGS_THREAD_COUNT =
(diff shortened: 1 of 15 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
```diff
@@ -342,16 +343,16 @@ public class AmoroManagementConf {
...
+          .defaultValue(Duration.ofMinutes(1))
           .withDescription("Timeout duration for Optimizer heartbeat.");
 
-  public static final ConfigOption<Long> OPTIMIZER_TASK_ACK_TIMEOUT =
+  public static final ConfigOption<Duration> OPTIMIZER_TASK_ACK_TIMEOUT =
       ConfigOptions.key("optimizer.task-ack-timeout")
-          .longType()
-          .defaultValue(30000L)
+          .durationType()
+          .defaultValue(Duration.ofSeconds(30))
```
klion26, line 355: If a user sets this conf to `60000` before this commit, and he/she needs to adjust the config to `60`, am I understanding right? if it is true, can we still use millisecond as the unit in the current commit?

amoro-ams/src/test/java/org/apache/amoro/server/TestAmoroManagementConf.java
```diff
@@ -0,0 +1,134 @@
...
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.time.Duration;
+import java.util.Map;
+
+public class TestAmoroManagementConf {
+  private static final ConfigOption<Duration>[] TIME_RELATED_CONFIG_OPTIONS =
```
klion26, line 45: Do we need to add a test to cover the migration for the changed configs? so that we know this does not change the user behavior, or if the behavior is indeed needed, we need to add them to the release note so that the user can know it(maybe add a upgrade script is better)

klion26 (commented):
@Jzjsnow thanks for the contribution, left some comments, please take another look when you're free.

amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
```diff
@@ -226,11 +226,12 @@ public class AmoroManagementConf {
           .defaultValue(1261)
           .withDescription("Port that the optimizing service thrift server is bound to.");
 
-  public static final ConfigOption<Long> THRIFT_MAX_MESSAGE_SIZE =
+  public static final ConfigOption<String> THRIFT_MAX_MESSAGE_SIZE =
```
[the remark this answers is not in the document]
Jzjsnow: Thanks for the suggestion, previously conversion to MemorySize type was not supported in `ConfigHelpers`. In my latest commit, `ConfigOption` can now take the MemorySize generic type.

amoro-ams/src/test/java/org/apache/amoro/server/TestAmoroManagementConf.java
Jzjsnow: I have added a unit test to parse the old version of the config file in the latest commit. The parsing result is as expected after adding a unit suffix to the original values of the old config file, which might serve as a reference for the upgrade scripts.

zhoujinsong:
@Jzjsnow Thanks for the contribution!

The support for time and memory configuration units is a great improvement.
It seems that the new configuration ensures a certain level of compatibility with the old configurations, except for `terminal.session.timeout`. We can mention this in the new version's release notes.

Besides, there are some places we should upgrade to the new version example:
- dist/src/main/amoro-bin/conf/config.yaml
- charts/amoro/templates/common/amoro-configmap.yaml

amoro-ams/src/test/java/org/apache/amoro/server/TestAmoroManagementConf.java
```diff
@@ -0,0 +1,241 @@
...
+          .put(AmoroManagementConf.AUTO_CREATE_TAGS_INTERVAL.key(), "ms")
+          .put(AmoroManagementConf.REFRESH_TABLES_INTERVAL.key(), "ms")
+          .put(AmoroManagementConf.BLOCKER_TIMEOUT.key(), "ms")
+          .put(AmoroManagementConf.OPTIMIZER_HB_TIMEOUT.key(), "ms")
+          .put(AmoroManagementConf.OPTIMIZER_TASK_ACK_TIMEOUT.key(), "ms")
+          .put(AmoroManagementConf.OPTIMIZER_POLLING_TIMEOUT.key(), "ms")
+          .put(AmoroManagementConf.TERMINAL_SESSION_TIMEOUT.key(), "min")
+          .build();
+
+  private static final ConfigOption<String>[] STORAGE_RELATED_CONFIG_OPTIONS =
```
klion26, line 68: This needs to be updated to `ConfigOption<MemorySize>`

klion26 (approved):
The change LGTM % test variable, thanks for the contribution.

~~@zhoujinsong with the change to `TERMINAL_SESSION_TIMEOUT`, should we plan for an upgrade script or anything else?~~
It seems the last reply has answered my question. just saw it after send my comment.

PR #3426, opened 2025-01-24
[AMORO-3424] Add the self-optimizing.partition-filter parameter

### LINKED ISSUE #3424: [Feature]: Add the self-optimizing.partition-filter parameter.
### Description

Specify the partition range for Amoro optimization by adding self-optimizing.partition-filter, similar to the function of the where parameter in Spark's rewrite_data_files procedures.

### Use case/motivation

Currently, when Amoro optimizes Iceberg tables, it defaults to optimizing data from all partitions. However, this can bring about the following issues in practical use:

High cost of optimizing historical data: Some tables' historical data may not conform to Amoro's optimization rules, and optimizing all historical data can lead to resource waste and performance degradation.

Conflict between concurrent writing and optimization: Some historical partitions may have data repair operations involving deletions, which can take a relatively long duration. In such cases, it's preferable for Amoro to skip the optimization of these partitions to avoid conflicts.

### Describe the solution

Add the self-optimizing.partition-filter parameter.

### Subtasks

- [ ] Add the self-optimizing.partition-filter parameter. @lintingbin 

### Related issues

_No response_

### Code of Conduct

lintingbin:

Close #3424 .

## Brief change log

- Add the self-optimizing.partition-filter parameter.

## How was this patch tested?

## Documentation

- If yes, how is the feature documented? docs


lintingbin:
Here is my current simple implementation code. Please provide some feedback, and I will continue to improve it.

zhoujinsong:
Thanks for the contribution!

The Spark core is a somewhat heavy lib to add to the common module.
[JSqlParser](https://github.com/JSQLParser/JSqlParser) is a lightweight lib to parse SQL, it may help.

lintingbin:
> Thanks for the contribution!
> 
> The Spark core is a somewhat heavy lib to add to the common module. [JSqlParser](https://github.com/JSQLParser/JSqlParser) is a lightweight lib to parse SQL, it may help.

Okay, I'll research JSqlParser.

zhoujinsong (commented):
@lintingbin Thanks for the work!

I left some comments, PTAL.

lintingbin:
> @lintingbin Thanks for the work!
> 
> I left some comments, PTAL.

It has been resolved. Please review it again.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
```diff
@@ -115,6 +115,9 @@ private TableProperties() {}
       "self-optimizing.full.rewrite-all-files";
   public static final boolean SELF_OPTIMIZING_FULL_REWRITE_ALL_FILES_DEFAULT = true;
 
+  public static final String SELF_OPTIMIZING_FILTER = "self-optimizing.filter";
+  public static final String SELF_OPTIMIZING_FILTER_DEFAULT = null;
```
czy006, line 119: If it is set to null here, is it possible that NPE may occur during subsequent verification? Can we use **""** instead?

amoro-format-iceberg/src/main/java/org/apache/amoro/utils/ExpressionUtil.java
```diff
@@ -79,4 +108,169 @@ public static Expression convertPartitionDataToDataFilter(
...
+   * @return the Iceberg {@link Expression}
+   */
+  public static Expression convertSqlFilterToIcebergExpression(
+      String sqlFilter, List<Types.NestedField> tableColumns) {
+    if (sqlFilter == null) {
+      return Expressions.alwaysTrue();
+    }
+
+    try {
+      Select statement = (Select) CCJSqlParserUtil.parse("SELECT * FROM dummy WHERE " + sqlFilter);
```
czy006, line 126: Why do we need fixation here? **SELECT * FROM dummy**

czy006 (commented):
Thank you for your contribution. I commented on some of the doubts.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
lintingbin: image

It won't cause a NullPointerException (NPE) when used only on this side. Moreover, I noticed that in the documentation (configurations.md), many parameter default values are NULL. I haven't seen any default values set as "". I wonder how it would be represented in the documentation if the default value were "".

amoro-format-iceberg/src/main/java/org/apache/amoro/utils/ExpressionUtil.java
czy006: thank you for your explanation.

zhoujinsong:
Thanks for the work! @lintingbin

Issue #3436, opened 2025-02-12
[Bug]: In the MergeInto statement, if a subquery is used for the source table, it cannot correctly identify whether the condition contains the primary key.


200947912:
### What happened?

When using the MergeInto statement, if you execute the filter method after obtaining the references attribute from the Expression object, the Attribute attribute you get will be the column names in the on clause, and you won't be able to correctly obtain the specific columns in the alias table. 

Here is an example:

> merge into target t using (select id,col from source) source s 
> on t.id=s.id 
> when matched then update set a.col = b.col 
> when not matched then insert *;

In this way, I will get an UnsupportedOperationException. Exception message is "The condition {'a.id, 'b.id} is not allowed because it is not a primary key."

In org.apache.amoro.spark.sql.catalyst.analysis.ResolveMergeIntoMixedFormatTableReferences#checkConditionIsPrimaryKey if using **cond.references** then i'll get 'a.id with 'b.id,not id#1 or id#2.

### Affects Versions

0.7.1

### Relevant log output

```shell

```

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3439, opened 2025-02-19
[AMORO-3436] Fix #{3436} merge into statement cannot find primary keys

200947912:
in merge into statement cannot find mixed table primarykey(#3436)
fixe merge into statement when source table is subquery cannot find mixed table primary key.


czy006 (commented):
Thanks for your first contribution, you can use **mvn spotless:apply** to align the format

PR #3450, opened 2025-02-28
[Improvement]:  Refactored MemorySizeTest.testUnitConversion to Parameterized Test

Monilnarang:

- In the test testUnitConversion the same order of method calls (getBytes, getKibiBytes, getMebiBytes,  . . . ) are repeated multiple times with different inputs, making the test harder to maintain and extend.
- Multiple such clusters of the above method calls are cluttered together in one test, blurring the exact scope of this test.
- When a test fails, JUnit only shows which assertion failed, but not which specific input caused the failure.
- Adding new test cases requires copying and pasting the whole block instead of simply adding new data.

## Brief change log
Parameterized testUnitConversion
This reduces duplication, allows easy extension by simply adding new value sets, and makes debugging easier as it clearly indicates which test failed instead of requiring a search through individual assertions.

Test Run report before changes: 
Screenshot 2025-02-28 at 1 45 20 PM

Test Run report after changes: 
Screenshot 2025-02-28 at 1 45 50 PM

## How was this patch tested?

## Documentation

No

### DIFF
--- a/amoro-common/src/test/java/org/apache/amoro/utils/MemorySizeTest.java
+++ b/amoro-common/src/test/java/org/apache/amoro/utils/MemorySizeTest.java
@@ -26,46 +26,32 @@ import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
 
 import org.junit.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.stream.Stream;
 
 /** Tests for the {@link MemorySize} class. */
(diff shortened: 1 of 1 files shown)


klion26:
@Monilnarang thanks for the contribution, seems there are some checkstyle problems in the change, could you please have a look at it.

zhoujinsong:
You can run `mvn spotless:apply` command under the root folder of this project to fix the checkstyle issue.

klion26 (approved):

zhoujinsong:
Thanks for the work! @Monilnarang 
Thanks for the review! @klion26

PR #3453, opened 2025-03-05
[Improvement]: Refactored 4 tests in MemorySizeTest to use parameterized unit testing

Monilnarang:
Aim:

Improve the test code by avoiding code duplication, improving maintainability, and enhancing readability. By converting the test into a parameterized unit test, we reduce boilerplate code, make it easier to extend by simply adding new input values, and improve debugging by clearly identifying which test case fails instead of searching through individual assertions.

- The same method call to `getKibiBytes`, `getMebiBytes`, `getGibiBytes`, & `getTebiBytes`  are repeated multiple times with different inputs in the tests `testParseKibiBytes`, `testParseMebiBytes`, `testParseGibiBytes`, & `testParseTebiBytes` respectively making them harder to maintain and extend.
- When any of these test fails, JUnit only shows which type of assertion failed, but not which specific input caused the failure.

## Brief change log
Parameterized `testParseKibiBytes`, `testParseMebiBytes`, `testParseGibiBytes`, & `testParseTebiBytes`
This reduces duplication, allows easy extension by simply adding new value sets, and makes debugging easier as it clearly indicates which test failed instead of requiring a search through individual assertions.

Test reports after changes: 
Screenshot 2025-03-04 at 7 29 59 PM
Screenshot 2025-03-04 at 7 30 05 PM

 

## How was this patch tested?

## Documentation

No

### DIFF
--- a/amoro-common/src/test/java/org/apache/amoro/utils/MemorySizeTest.java
+++ b/amoro-common/src/test/java/org/apache/amoro/utils/MemorySizeTest.java
@@ -27,8 +27,8 @@ import static org.junit.Assert.fail;
 
 import org.junit.Test;
 import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.MethodSource;
 
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


zhoujinsong (approved):
LGTM.

Thanks for the contribution!

But there seem to be some conflicts with the master branch, you may need to fix

klion26:
@Monilnarang thanks for the contribution, and @zhoujinsong for the review!

PR #3455, opened 2025-03-05
[Improvement]: Refactored 2 tests in TestTreeNode

Monilnarang:
Aim:

Improve the test code by avoiding code duplication, improving maintainability, and enhancing readability. By converting the test into a parameterized unit test, we reduce boilerplate code, make it easier to extend by simply adding new input values, and improve debugging by clearly identifying which test case fails instead of searching through individual assertions.

- The same method call are repeated multiple times with different inputs in the tests `testParentNode` & `testNodeId` making them harder to maintain and extend.
- When any of these tests fails, JUnit only shows which type of assertion failed, but not which specific input caused the failure.

## Brief change log
Parameterized the tests: `testParentNode` & `testNodeId`. This reduces duplication, allows easy extension by simply adding new value sets, and makes debugging easier as it clearly indicates which test failed instead of requiring a search through individual assertions.

Test run report after change:
Screenshot 2025-03-04 at 8 40 28 PM

## How was this patch tested?

## Documentation

No

### DIFF
--- a/amoro-format-iceberg/src/test/java/org/apache/amoro/TestTreeNode.java
+++ b/amoro-format-iceberg/src/test/java/org/apache/amoro/TestTreeNode.java
@@ -21,6 +21,8 @@ package org.apache.amoro;
 import org.apache.amoro.data.DataTreeNode;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #3456, opened 2025-03-05
[Improvement]: Improved test testGetUriPath in TestFileUtil by retrofitting to parameterized test

Monilnarang:
- The same method call are repeated multiple times with different inputs in the test `testGetUriPath` making it harder to maintain and extend.
- When the test fails, JUnit only shows which type of assertion failed, but not which specific input caused the failure.

## Brief change log
Parameterized test `testGetUriPath`. This reduces duplication, allows easy extension by simply adding new value sets, and makes debugging easier as it clearly indicates which test failed instead of requiring a search through individual assertions.

Test run report after change:
Screenshot 2025-03-04 at 9 51 16 PM

## How was this patch tested?

## Documentation

No


zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #3458, opened 2025-03-05
[AMORO-3451] Fix kubernetes optimizer pod is frequently being killed

### LINKED ISSUE #3451: [Bug]: Kubernetes optimizer pod is frequently being killed
### What happened?

Kubernetes optimizer's JVM heap memory limit is the same as the pod memory limit, which is the cause of the issue.

### Affects Versions

master

Iceberg

AMS, Optimizer

### How to reproduce

- Create a Kubernetes optimizer
- Schedule some task to it

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

Marvel-Gu:

Close #3451.

## Brief change log

In ams KubernetesOptimizerContainer, limit the size of the JVM heap memory to 80% of the pod's memory limit.

## How was this patch tested?

## Documentation


Marvel-Gu:
I didn't format the code😂, I'll pay attention next time.

amoro-ams/src/main/java/org/apache/amoro/server/manager/KubernetesOptimizerContainer.java
```diff
@@ -148,18 +148,23 @@ public Map<String, Object> generatePodStartArgs(
...
 
     if (resource.getMemoryMb() > 0) {
       memory = resource.getMemoryMb();
     } else {
       memoryPerThread = Long.parseLong(checkAndGetProperty(groupProperties, MEMORY_PROPERTY));
       memory = memoryPerThread * resource.getThreadCount();
     }
+
+    // reserving 20% of total pod memory for JVM non-heap memory
+    jvmHeapMemory = (long) (memory * 0.8);
```
czy006, line 161: If we do not use off-heap memory in the code, I think it can be set to 0.9, which will not cause the container to be killed. If it is still frequently killed by OOM, we may need to check whether there is a memory leak in the code or the instantaneous Heap usage exceeds the container limit. We may be able to provide an external parameter for users to set, WDYT?

Marvel-Gu: Sounds great, I'll work on it.

zhoujinsong:
The default 80% configuration looks good to me. Maybe we can merge this PR now or do we have some other thing to validate before that? @Marvel-Gu @czy006

czy006:
> The default 80% configuration looks good to me. Maybe we can merge this PR now or do we have some other thing to validate before that? @Marvel-Gu @czy006

This is indeed a black magic in the container. We can merge this PR now, and we may have a better way to solve it in the future.

> #3451

Marvel-Gu:
> > The default 80% configuration looks good to me. Maybe we can merge this PR now or do we have some other thing to validate before that? @Marvel-Gu @czy006
> 
> This is indeed a black magic in the container. We can merge this PR now, and we may have a better way to solve it in the future.

This is the simplest and most effective solution I could think of for now. How do you think it could be improved?

charts/amoro/values.yaml
```diff
@@ -278,6 +278,7 @@ optimizer:
       #kube-config-path: "~/.kube/config"
       image: "apache/amoro:latest"
       pullPolicy: "IfNotPresent"
+      jvm.heap.ratio: "0.8"
   extra: []
```
czy006, line 282: I'm worried that our parameters will increase, can we put them into extra? Perhaps there is an additional map maintenance.

Marvel-Gu: Yes, putting it in extra is a better approach, I will modify it

zhoujinsong (approved):
LGTM.

PR #3462, opened 2025-03-11
[AMORO-3459][Improvement]: Obtain the master's service address based on service discovery, which requires writing the HTTP port to ZooKeeper

### LINKED ISSUE #3459: [Improvement]: Obtain the master's service address based on service discovery, which requires writing the HTTP port to ZooKeeper

czy006:
[AMORO-3459][Improvement]: Obtain the master's service address based on service discovery, which requires writing the HTTP port to ZooKeeper(#3459)

Close #3459.

## Brief change log

- Add AMS Rest Port in AmsServerInfo
- AMSROOT Path change **artic** to **amoro**

## How was this patch tested?

## Documentation


zhoujinsong:
@czy006 I'm afraid that this change might break client compatibility, causing older versions of the client to fail to retrieve information correctly. We may need to consider the risks associated with the version upgrade.

czy006:
> @czy006 I'm afraid that this change might break client compatibility, causing older versions of the client to fail to retrieve information correctly. We may need to consider the risks associated with the version upgrade.

This is indeed a destructive upgrade. We have changed the metadata and zk address. Could we explain this in the documentation? Or do we have to enforce compatibility? @zhoujinsong

baiyangtx:
> > @czy006 I'm afraid that this change might break client compatibility, causing older versions of the client to fail to retrieve information correctly. We may need to consider the risks associated with the version upgrade.
> 
> This is indeed a destructive upgrade. We have changed the metadata and zk address. Could we explain this in the documentation? Or do we have to enforce compatibility? @zhoujinsong

Considering that the data stored in zk is ephemeral, this upgrade only requires the optimizer, client, and ams to be upgraded together without involving data migration. Can it be included in version 0.8? 

@zhoujinsong

czy006:
> > > @czy006 I'm afraid that this change might break client compatibility, causing older versions of the client to fail to retrieve information correctly. We may need to consider the risks associated with the version upgrade.
> > 
> > 
> > This is indeed a destructive upgrade. We have changed the metadata and zk address. Could we explain this in the documentation? Or do we have to enforce compatibility? @zhoujinsong
> 
> Considering that the data stored in zk is ephemeral, this upgrade only requires the optimizer, client, and ams to be upgraded together without involving data migration. Can it be included in version 0.8?
> 
> @zhoujinsong

We need to remind users to stop AMS and Optimizer, but I believe that most users will do it

baiyangtx (approved):

PR #3518, opened 2025-04-15
[AMORO-3517]fix:MixedTableMaintainer would never expire a non-pk base…

### LINKED ISSUE #3517: [Bug]: MixedTableMaintainer would never expire a non-pk base table if it would not be batched written
### What happened?

Consider a scenario that a mixed table is initialized by a Spark batch job to fill history data and then is ingested with streaming data by Flink. The MixedTableMaintainer would never expire any snapshots of base table as it takes  'last-non-optimized snapshot time' into consideration to choose lower bound of snapshot to keep.  The  last-non-optimized snapshot  in base table is generated by data initialisation and would not been updated any longer.  Thus,  in my scenario, the snapshots of base table would not been expired eventually.

### Affects Versions

0.7

Mixed-Iceberg, Mixed-Hive

AMS

### How to reproduce

Refer to what happend

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

nicochen:

Close #3517.

## Brief change log

Modify MixedTableMaintainer.mustOlderThan() to judge whether the 'last non optimized snapshot time' should be considered as a factor to determine which snapshots to keep when expiring. 

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/maintainer/MixedTableMaintainer.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/maintainer/MixedTableMaintainer.java
@@ -460,13 +460,22 @@ public class MixedTableMaintainer implements TableMaintainer {
 
     @Override
     protected long mustOlderThan(TableRuntime tableRuntime, long now) {
+      DataExpirationConfig expiringDataConfig =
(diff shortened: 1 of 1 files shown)


Aireed (approved):
LGTM

Issue #3445, opened 2025-02-26
[Bug]: Some history records in database are not dropped as expected


zhoujinsong:
### What happened?

Some records created before 30 days existed in `table_optimizing_process/task_runtime/optimizing_task_quota` database tables with `self-optimizing.runtime-data-keep-days` configuration is set to 30.

### Affects Versions

master

### What table formats are you seeing the problem on?

Iceberg

### What engines are you seeing the problem on?

AMS

### How to reproduce

It seems the dropped tables left some records in the database will not be deleted any more.

### Relevant log output

```shell

```

### Are you willing to submit a PR?

- [ ] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ihadoop:
Is it means when table is dropped， it should clean all optimizing data immediately. 

In addition, we should remove  table history recorders in database  which cannot find table name in any catalog 。 
Could you confirm is it right? @zhoujinsong

zhoujinsong:
> Is it means when table is dropped， it should clean all optimizing data immediately.
> In addition, we should remove table history recorders in database which cannot find table name in any catalog 

Both yes for these two questions.

zhoujinsong:
The current clean method for the optimizing histories is scheduled separately for each table. I believe this design can be simplified, and this issue can be resolved simultaneously:

- Schedule the cleaning process for all tables together.
- Create a daily partition for history tables and change the clean SQL to drop partition.

Jzjsnow:
@ihadoop Hi, I would like to know how this issue is going, we are experiencing similar problems.

zhoujinsong:
It seems not to be fixed for now. Do you have time to address and fix it? @ihadoop

ihadoop:
I will submit pr next weekend

PR #3520, opened 2025-04-15
[AMORO-3252] Fix primary key duplicate exception when concurrently inserting table optimization process entries into database

### LINKED ISSUE #3252: [Bug]: Duplicate entry for key 'PRIMARY'  (processId)
### What happened?

Primary key conflicts occur, especially when starting  parallelized planning。

`processId` from:

https://github.com/apache/amoro/blob/4e7fc9b0eb35a6b8768e9ab79a99efed9622b0ea/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/plan/OptimizingPlanner.java#L80-L83

### Affects Versions

0.7.0

Iceberg

Optimizer

### How to reproduce

When starting parallel planning, the probability of occurrence is higher

### Relevant log output

```shell
2024-10-14 00:19:45,733 ERROR [plan-executor-thread-499] [org.apache.amoro.server.optimizing.OptimizingQueue] [] - Planning table xx.xx.xx(tableId=30058) failed
org.apache.amoro.server.exception.PersistenceException: org.apache.ibatis.exceptions.PersistenceException:
### Error updating database.  Cause: java.sql.SQLIntegrityConstraintViolationException: Duplicate entry '1728836383424' for key 'PRIMARY'
### The error may exist in org/apache/amoro/server/persistence/mapper/OptimizingMapper.java (best guess)
### The error may involve org.apache.amoro.server.persistence.mapper.OptimizingMapper.insertOptimizingProcess-Inline
### The error occurred while setting parameters
### SQL: INSERT INTO table_optimizing_process(table_id, catalog_name, db_name, table_name ,process_id, target_snapshot_id, target_change_snapshot_id, status, optimizing_type, plan_time, summary, from_sequence, to_sequence) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
### Cause: java.sql.SQLIntegrityConstraintViolationException: Duplicate entry '1728836383424' for key 'PRIMARY'
```

### Anything else

_No response_

### Code of Conduct

Jzjsnow:
Currently the primary keys of the tables `table_optimizing_process`, `task_runtime`, and `optimizing_task_quota` contain only `process_id` and not `table_id`, where `process_id` is created by the current timestamp, e.g. 1744702084683. When multiple tables are optimized concurrently, it is easy to have the same `process_id` at the same trigger time, resulting in a primary key duplication exception `org.apache.ibatis.exceptions.PersistenceException` during persistence.

Close #3252.

## Brief change log

-  Add table_id to the primary keys of  `table_optimizing_process`, `task_runtime`, and `optimizing_task_quota` respectively.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/resources/derby/ams-derby-init.sql
+++ b/amoro-ams/src/main/resources/derby/ams-derby-init.sql
@@ -140,7 +140,7 @@ CREATE TABLE table_optimizing_process (
     summary             CLOB(64m),
     from_sequence       CLOB(64m),
     to_sequence         CLOB(64m),
-    CONSTRAINT table_optimizing_process_pk PRIMARY KEY (process_id)
+    CONSTRAINT table_optimizing_process_pk PRIMARY KEY (process_id, table_id)
 );
 
 CREATE TABLE task_runtime (
@@ -160,7 +160,7 @@ CREATE TABLE task_runtime (
     rewrite_output  BLOB,
     metrics_summary CLOB,
     properties      CLOB,
-    CONSTRAINT task_runtime_pk PRIMARY KEY (process_id, task_id)
+    CONSTRAINT task_runtime_pk PRIMARY KEY (process_id, task_id, table_id)
 );
 
 CREATE TABLE optimizing_task_quota (
@@ -171,7 +171,7 @@ CREATE TABLE optimizing_task_quota (
     start_time      TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     end_time        TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     fail_reason     VARCHAR(4096),
-    CONSTRAINT optimizing_task_quota_pk PRIMARY KEY (process_id, task_id, retry_num)
+    CONSTRAINT optimizing_task_quota_pk PRIMARY KEY (process_id, task_id, retry_num, table_id)
 );
 
 CREATE TABLE api_tokens (
--- a/amoro-ams/src/main/resources/mysql/ams-mysql-init.sql
+++ b/amoro-ams/src/main/resources/mysql/ams-mysql-init.sql
@@ -153,7 +153,7 @@ CREATE TABLE `table_optimizing_process`
     `summary`                       mediumtext COMMENT 'Max change transaction id of these tasks',
     `from_sequence`                 mediumtext COMMENT 'from or min sequence of each partition',
     `to_sequence`                   mediumtext COMMENT 'to or max sequence of each partition',
-    PRIMARY KEY (`process_id`),
+    PRIMARY KEY (`process_id`, `table_id`),
     KEY  `table_index` (`table_id`, `plan_time`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT 'History of optimizing after each commit';
 
@@ -175,7 +175,7 @@ CREATE TABLE `task_runtime`
     `rewrite_output`            longblob DEFAULT NULL COMMENT 'rewrite files output',
     `metrics_summary`           text COMMENT 'metrics summary',
     `properties`                mediumtext COMMENT 'task properties',
-    PRIMARY KEY (`process_id`, `task_id`),
+    PRIMARY KEY (`process_id`, `task_id`, `table_id`),
     KEY  `table_index` (`table_id`, `process_id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COMMENT 'Optimize task basic information';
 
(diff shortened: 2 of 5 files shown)


czy006 (approved):
LGTM

klion26 (approved):
LGTM thanks for the contribution, lets see if there is any more comments from the community.

zhoujinsong:
Hi, Thanks for driving this.

After I listed all SQL related to `table_optimizing_process`, I found that the condition of some SQL does not contain `table_name`, which may cost poor performance, like: https://github.com/apache/amoro/blob/master/amoro-ams/src/main/java/org/apache/amoro/server/persistence/mapper/OptimizingMapper.java#L147

I think the `table_optimizing_process` might only need a globally unique ID as the primary key, without necessarily including `table_name`. However, the current ID generation rules do indeed carry a risk of duplication.

Currently, we use the`currentTimestamp` generation rule, which I understand is for easier cleanup operations. However, perhaps we should optimize the current cleanup logic, for example, by partitioning this table and using DROP PARTITION to perform cleanup more efficiently.
#3445 is following the cleaning improvement issue.

Jzjsnow:
> Hi, Thanks for driving this.
> 
> After I listed all SQL related to `table_optimizing_process`, I found that the condition of some SQL does not contain `table_name`, which may cause poor performance, like: https://github.com/apache/amoro/blob/master/amoro-ams/src/main/java/org/apache/amoro/server/persistence/mapper/OptimizingMapper.java#L147
> 
> I think the `table_optimizing_process` might only need a globally unique ID as the primary key, without necessarily including `table_name`. However, the current ID generation rules do indeed carry a risk of duplication.
> 
> Currently, we use the`currentTimestamp` generation rule, which I understand is for easier cleanup operations. However, perhaps we should optimize the current cleanup logic, for example, by partitioning this table and using DROP PARTITION to perform cleanup more efficiently. #3445 is following the cleaning improvement issue.

@zhoujinsong Thanks for the viewpoint. I think expiring all tables during cleanup optimization can greatly improve optimization performance as discussed in #3445. On the other hand, I think it is still necessary to keep the `tableid` in the query sql related to tables `table_optimizing_process`/`task_runtime` / `optimizing_task_quota` because we may still need to process the optimization information of a single table. 

For example, as mentioned in #3445, we'd better clean up all the optimization information of a table immediately when it is dropped, and in #3554 when we clear orphan table information during table service initializing, we still need tableid to filter out the optimizing entries related to the invalid tables.

Jzjsnow:
> I think the `table_optimizing_process` might only need a globally unique ID as the primary key, without necessarily including `table_name`. However, the current ID generation rules do indeed carry a risk of duplication.

Thanks to the tip about the process_id unique key, I have since researched the use of the `table_optimizing_process` table in depth. I found that `process_id` is critical to apply as a primary key, such as canceling the optimizing process (from the front end) and updating the taskRuntime etc.

I agree that using timestamps as `process_id` is better for cleanup process performance, and I think the solution to the current problem of having duplicate timestamps is clearer and more concise: the duplicate `process_id` occurs due to the highly concurrent execution of the `TableOptimizingProcess#planInternal()`. 

In the latest commit, i generate the `process_id` with a timestamp before the asynchronization to ensure the uniqueness of the `process_id`, PTAL.

czy006:
@Jzjsnow Hi,After resolving the conflict, we can continue with the review

Jzjsnow:
> @Jzjsnow Hi,After resolving the conflict, we can continue with the review

@czy006 Sure, I've resolved the conflict and rebased the changes. PTAL

klion26:
LGTM for the new change, as the change only changes the `planTime` generate logic, and the `planTime`, which was generated in a single thread, the change here avoids the duplication of `planTime` for different tables before.

I'm not sure if there's a chance of duplication of `planTime` in the multi-node scenario, but that can be addressed in a separate issue if it exists.

zhoujinsong:
I am not sure if the new change can resolve the conflicts of process ID generation.
Different groups will generate processes at the same time, which may generate the same process ID.

HDYT?

Jzjsnow:
@xxubai @zhoujinsong Thanks a lot for the feedback!

We’ve made the change to use Snowflake IDs instead of timestamps. This should fix the uniqueness processId issue between different groups mentioned above, especially when processes are generated at the same time.

Hope this clears things up. Please take a look when you are free.

xxubai (commented):
Thanks!
In Javascript, the maximum safe integer is `2^53 - 1`, but snowflake Id normally greater than the safe integer.
I'd like to compress the snowflake id generator to make it safe:
```
public class SnowflakeIdGenerator {
  // Number of bits allocated for the machine ID part
  private static final long MACHINE_ID_BITS = 5L;

  // Number of bits allocated for the sequence number part
  private static final long SEQUENCE_BITS = 8L;

  private long currentTime() {
    return System.currentTimeMillis() / 10;
  }

  public static long getMinSnowflakeId(long timestamp) {
    // if timestamp is milliseconds, compress it to 10ms
    if (timestamp < 10000000000L) {
      timestamp *= 100; // Convert seconds to 10ms
    } else {
      timestamp /= 10; // Convert milliseconds to 10ms
    }

    return (timestamp << TIMESTAMP_LEFT_SHIFT);
  }
}
```
Therefore, we don't need to make changes to the front.
Additionally, the ID generator can be implemented as a singleton class with a default machine ID, since AMS currently runs on a single node.

xxubai (approved):
Thanks @Jzjsnow

zhoujinsong (approved):
LGTM.

Thanks for the great work!

PR #3529, opened 2025-04-23
Fix CVE-2025-30065 and CVE-2025-24970

XBaith:

Bump `parquet-avro` to `1.15.1` and `netty-all` to `4.8.118.Final`

## Brief change log

- Fix critical and high CVE

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-format-iceberg/pom.xml
+++ b/amoro-format-iceberg/pom.xml
@@ -135,6 +135,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.apache.parquet</groupId>
+            <artifactId>parquet-jackson</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-auth</artifactId>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -117,7 +117,8 @@
         <cglib.version>2.2.2</cglib.version>
         <curator.version>5.7.0</curator.version>
         <mockito.version>4.11.0</mockito.version>
-        <parquet-avro.version>1.13.1</parquet-avro.version>
+        <parquet-jackson.version>1.13.1</parquet-jackson.version>
+        <parquet-avro.version>1.15.1</parquet-avro.version>
         <mysql-jdbc.version>8.0.33</mysql-jdbc.version>
         <orc-core.version>1.8.3</orc-core.version>
         <awssdk.version>2.24.12</awssdk.version>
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -372,6 +373,12 @@
                 <version>${parquet-avro.version}</version>
             </dependency>
 
+            <dependency>
```
zhoujinsong, line 376: It seems `parquet-avro` is dependent on `parquet-jackson` in runtime scope after pumping up to 1.15.1; this is the reason why we need it here.

amoro-format-iceberg/pom.xml
```diff
@@ -134,6 +134,11 @@
             <artifactId>parquet-avro</artifactId>
         </dependency>
 
+        <dependency>
```
zhoujinsong, line 137: Do we need to declare it a dependency here, as `amoro-format-iceberg` has already dependent on `amoro-common`?

amoro-format-mixed/amoro-mixed-trino/src/test/java/org/apache/amoro/trino/mixed/TestKeyedTable.java
```diff
@@ -62,7 +62,7 @@ public void testStats() {
         .matches(
             "VALUES "
                 + "('id', NULL, NULL, 0e0, NULL, '1', '4'), "
-                + "('name$name', 429e0, NULL, 0e0, NULL, NULL, NULL), "
+                + "('name$name', 461e0, NULL, 0e0, NULL, NULL, NULL), "
```
zhoujinsong, line 65: I have no idea why this change is needed.

zhoujinsong (commented):
@XBaith Thanks for driving this, I left some comments, PTAL.

pom.xml
XBaith: Actually, parquet-avro 1.15.1 contains parquet-jackson 1.15.1 which lead fail to complie the project with JDK8
I replace paruqet-jackson 1.15.1 with 1.13.1 to build the project correctly.

amoro-format-iceberg/pom.xml
XBaith: +1. the parquet-jackson and parquet-avro are included

pom.xml
zhoujinsong: Got it, we may remove this after upgrading the JDK version, maybe add a comment to add this TODO.

amoro-format-iceberg/pom.xml
zhoujinsong: Yes, so we may drop the dependencies here?

amoro-format-mixed/amoro-mixed-trino/src/test/java/org/apache/amoro/trino/mixed/TestKeyedTable.java
XBaith: `data_size` is an estimate and I'm not sure why it changes. I'm guessing it's the encoding strategy that causes the file size to change.

PR #3546, opened 2025-05-02
[AMORO-3545] Prevent empty tasks list during planning

### LINKED ISSUE #3545: [Bug]: AMS stops optimizing when some partitions have high cost but cannot be scheduled
### What happened?

In AbstractOptimizingPlanner, once the number of inputs exceeds maxInputSize, additional partition evaluators are dropped—yet at least one evaluator is always retained.

However, if the remaining evaluators only reference unschedulable tasks (for example, in IcebergPartitionPlan when a task has just a single data file), the planner will produce no output, even though the original evaluator list included valid tasks.

This happens because the logic is two staged and doesn't consider the task eligibility when pruning partitions.

### Affects Versions

master

Iceberg

AMS

### How to reproduce

_No response_

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

lrsb:
Fix #3545 

This change merges the logic in AbstractOptimizingPlanner to avoid having `actualPartitionPlans` with only unschedulable tasks.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
@@ -36,9 +36,9 @@ import org.apache.iceberg.util.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -151,27 +151,33 @@ public abstract class AbstractOptimizingPlanner extends AbstractOptimizingEvalua
       return cacheAndReturnTasks(Collections.emptyList());
     }
 
-    List<PartitionEvaluator> evaluators = new ArrayList<>(needOptimizingPlanMap.values());
+    LinkedList<PartitionEvaluator> evaluators = new LinkedList<>(needOptimizingPlanMap.values());
     // prioritize partitions with high cost to avoid starvation
     evaluators.sort(Comparator.comparing(PartitionEvaluator::getWeight, Comparator.reverseOrder()));
 
     double maxInputSize = maxInputSizePerThread * availableCore;
     actualPartitionPlans = Lists.newArrayList();
     long actualInputSize = 0;
-    for (PartitionEvaluator evaluator : evaluators) {
-      actualPartitionPlans.add((AbstractPartitionPlan) evaluator);
-      actualInputSize += evaluator.getCost();
-      if (actualInputSize > maxInputSize) {
-        break;
+    List<RewriteStageTask> plannedTasks = Lists.newArrayList();
+
+    while (plannedTasks.isEmpty() && !evaluators.isEmpty()) {
+      for (PartitionEvaluator evaluator = evaluators.poll();
+          evaluator != null;
+          evaluator = evaluators.poll()) {
+        actualPartitionPlans.add((AbstractPartitionPlan) evaluator);
+        actualInputSize += evaluator.getCost();
+        if (actualInputSize > maxInputSize) {
+          break;
+        }
+      }
+
+      double avgThreadCost = actualInputSize / availableCore;
+      for (AbstractPartitionPlan partitionPlan : actualPartitionPlans) {
+        plannedTasks.addAll(partitionPlan.splitTasks((int) (actualInputSize / avgThreadCost)));
       }
     }
 
-    double avgThreadCost = actualInputSize / availableCore;
-    List<RewriteStageTask> tasks = Lists.newArrayList();
-    for (AbstractPartitionPlan partitionPlan : actualPartitionPlans) {
-      tasks.addAll(partitionPlan.splitTasks((int) (actualInputSize / avgThreadCost)));
-    }
-    if (!tasks.isEmpty()) {
+    if (!plannedTasks.isEmpty()) {
       if (actualPartitionPlans.stream()
           .anyMatch(plan -> plan.getOptimizingType() == OptimizingType.FULL)) {
         optimizingType = OptimizingType.FULL;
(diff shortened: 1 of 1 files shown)


amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
```diff
@@ -158,19 +158,23 @@ public List<RewriteStageTask> planTasks() {
...
     long actualInputSize = 0;
+    double avgThreadCost = actualInputSize / availableCore;
+    List<RewriteStageTask> tasks = Lists.newArrayList();
+
     for (PartitionEvaluator evaluator : evaluators) {
-      actualPartitionPlans.add((AbstractPartitionPlan) evaluator);
-      actualInputSize += evaluator.getCost();
+      AbstractPartitionPlan actualPartitionPlan = (AbstractPartitionPlan) evaluator;
+      List<RewriteStageTask> splitTasks =
+          actualPartitionPlan.splitTasks((int) (actualInputSize / avgThreadCost));
```
klion26, line 167: It is ok to change this currently, because the parameter for `splitTasks` is not taken into consideration, but maybe in the future we will change the logic in `TaskSplitter#splitTasks`

Can we change the `actualInputSize / avgThreadCost` here to some other things like `evaluator.getCost() / avgThreadCostPre` -- but currently there is no `avgThreadCostPre` 

Maybe we can add a comment here to let others know we need to change this here? what do you think about this

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
```diff
@@ -158,19 +158,23 @@ public List<RewriteStageTask> planTasks() {
     double maxInputSize = maxInputSizePerThread * availableCore;
     actualPartitionPlans = Lists.newArrayList();
     long actualInputSize = 0;
+    double avgThreadCost = actualInputSize / availableCore;
```
klion26, line 161: I have a question about the change here.

Is the change here mainly to correct the log at the end of this function? is my understanding right?

Before the change, if the `tasks` is empty because there are no tasks after `actualPartitionPlan.splitTasks`, then there would be no actual optimization executed because we will check whether the `tasks` are empty or not in `AbstractOptimizingPlanner#isNecessary` and `OptimizingQueue#loadTaskRuntimes`.
- After the change, the `tasks` will be empty also, but the `actualInputSize` is the right value.

If this is the case, can we correct the calculation logic for `actualInputSize`?

klion26 (commented):
@lrsb thanks for the contribution, I left some inline comments, please take a look when you're free.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
lrsb: I see what you mean, will publish another review.

Regarding the comment I think it should be placed in the implementation classes.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
lrsb: > Is the change here mainly to correct the log at the end of this function?

No the issue is different:
- Let's say there's a PartitionEvaluator which output of splitTasks is empty (because, for example in Iceberg, it contains only one data file)
- Let's also say its cost is > maxInputSize

The result is that:
- actualPartitionPlans will contain just that evaluator even if there are more in the initial list
- Since split tasks output is empty nothing will be scheduled
- If this keeps happening we get into a starvation like situation where the first evaluator keeps getting scheduled but nothing is actually executed even if the initial list there are schedulable evaluators

> After the change, the tasks will be empty also, but the actualInputSize is the right value.

No, the output after the change will be populated with the other partition evaluators

klion26: Thanks for the explain.

klion26:
@lrsb thanks for the update. The change looks fine to me. Please refine the code style to pass the style check in CI.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
```diff
@@ -151,27 +151,31 @@ public List<RewriteStageTask> planTasks() {
       return cacheAndReturnTasks(Collections.emptyList());
     }
 
-    List<PartitionEvaluator> evaluators = new ArrayList<>(needOptimizingPlanMap.values());
+    LinkedList<PartitionEvaluator> evaluators = new LinkedList<>(needOptimizingPlanMap.values());
```
klion26, line 154: do you think we need to update the `needOptimizingPlanMap` after the plan

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
```diff
@@ -151,27 +151,33 @@ public List<RewriteStageTask> planTasks() {
...
```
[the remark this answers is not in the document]
lrsb: changed to priority queue since makes more sense given we are sorting too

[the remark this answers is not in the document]
xxubai: I think using a `Deque` is sufficient here, as a priority queue is not more efficient than a list in this case, especially during polling. Therefore, I don’t think using a priority queue for sorting is necessarily appropriate.

xxubai (approved):

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingPlanner.java
```diff
@@ -151,27 +151,33 @@ public List<RewriteStageTask> planTasks() {
...
-      if (actualInputSize > maxInputSize) {
-        break;
+    List<RewriteStageTask> plannedTasks = Lists.newArrayList();
+
+    while (plannedTasks.isEmpty() && !evaluators.isEmpty()) {
+      for (PartitionEvaluator evaluator = evaluators.poll();
+          evaluator != null;
+          evaluator = evaluators.poll()) {
+        actualPartitionPlans.add((AbstractPartitionPlan) evaluator);
+        actualInputSize += evaluator.getCost();
```
zhoujinsong, line 168: Nice catch!

But I am wondering if we should calculate the input size depending on the real tasks from plannedTasks rather than `evaluator#cost` if the cost is not true.

PR #3560, opened 2025-05-12
[AMORO-3559] Extract abstract class for table metrics

### LINKED ISSUE #3559: [Improvement]: Introduce abstract class TableMetrics for metric system
### Search before asking

Currently, we have multiple metric classes such as TableOptimizingMetrics, TableSummaryMetrics, and TableOrphanMetrics. Each of these classes is responsible for registering and managing specific metrics related to different aspects of table operations. However, there is a significant amount of duplicated code for registering and unregistering metrics. This duplication makes the code harder to maintain and extend.

### How should we improve?

To improve code maintainability and reduce redundancy, we can extract a common abstract class `TableMetrics`.

### Subtasks

_No response_

### Code of Conduct

Jzjsnow:

In the current implementation, each specific metric class (i.e. “TableOptimizingMetrics”, “TableOrphanFilesCleaningMetrics”, “TableSummaryMetrics") all contain similar logic and duplicate code for registering and unregistering metrics. This duplication makes the metrics more difficult to maintain and extend. To improve code maintainability and reduce redundancy, we introduce a generic abstract class `TableMetrics`.

Close #3559.

## Brief change log

- Create `TableMetrics` abstract class
- Update `TableOptimizingMetrics`, `TableOrphanFilesCleaningMetrics`, and `TableSummaryMetrics` to extend `TableMetrics`.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/TableMetrics.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/TableMetrics.java
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.amoro.server.table;
+
+import org.apache.amoro.ServerTableIdentifier;
+import org.apache.amoro.metrics.Metric;
+import org.apache.amoro.metrics.MetricDefine;
+import org.apache.amoro.metrics.MetricKey;
+import org.apache.amoro.server.metrics.MetricRegistry;
+import org.apache.amoro.shade.guava32.com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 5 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/table/TableMetrics.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.amoro.metrics.Metric;
+import org.apache.amoro.metrics.MetricDefine;
+import org.apache.amoro.metrics.MetricKey;
+import org.apache.amoro.server.metrics.MetricRegistry;
+import org.apache.amoro.shade.guava32.com.google.common.collect.ImmutableMap;
+import org.apache.amoro.shade.guava32.com.google.common.collect.Lists;
+
+import java.util.List;
+
+public abstract class TableMetrics {
```
zhoujinsong, line 31: How about renaming the class name to `AbstractTableMetrics`?

zhoujinsong (commented):
I left a minor suggestion, PTAL.

zhoujinsong:
Actually, I've noticed that other metrics like `OptimizerGroupMetrics` also share some common implementations, such as `register`/`unregister`. Perhaps we could later extract an `AbstractMetrics` implementation, but this can be addressed in a separate PR. 

HDYT?

amoro-ams/src/main/java/org/apache/amoro/server/table/TableMetrics.java
Jzjsnow: Thanks for the suggestion! AbstractTableMetrics is more intuitive and consistent. I’ve updated it accordingly.

Jzjsnow:
> Actually, I've noticed that other metrics like `OptimizerGroupMetrics` also share some common implementations, such as `register`/`unregister`. Perhaps we could later extract an `AbstractMetrics` implementation, but this can be addressed in a separate PR.
> 
> HDYT?

Good catch! I looked into it more closely and you're right — besides table-level metrics, both OptimizerGroup and ams service-level metrics also share similar registration logic. It makes sense to extract a common `AbstractMetrics` class. I’ll handle this refactor in a separate PR. Thanks for the suggestion!

zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #3577, opened 2025-05-23
Bump appVersion to 0.9-SNAPSHOT in Chart

xxubai:

Follow #3576.

## Brief change log

- Bump appVersion in Helm chart

## How was this patch tested?

## Documentation


charts/amoro/Chart.yaml
```diff
@@ -20,7 +20,7 @@ name: amoro
 description: A Helm chart for Amoro
 type: application
 version: 0.1.0
-appVersion: "0.7"
+appVersion: "0.9-SNAPSHOT"
```
klion26, line 23: Does this value need to be the latest stable version?

find that the history value for appVersion is: 0.5.0 -> 0.6.0 -> 0.7

xxubai: I think the chart version should follow the project version. If the current version is stable then it should be `0.8.0` or `0.8.0-incubating`.

zhoujinsong: Yes, we may make the `tools/change-version.sh` to change version here too.
Maybe improve this in another PR.

zhoujinsong (approved):
LGTM.

PR #3592, opened 2025-06-04
Optimize log output for SchedulingPolicy

lintingbin:

If the optimizer is in a relatively idle state, org.apache.amoro.server.optimizing.SchedulingPolicy will print a large number of logs. It is recommended to adjust the info level to debug.
image
The log below is sufficient to determine the scheduling strategy of a table.
image

## Brief change log

- Change the log output level from info to debug.

## How was this patch tested?

## Documentation


czy006:
+1, noice log with scheduler

Issue #3608, opened 2025-06-11
[Improvement]: Improve the stability of deleting expired files


xxubai:
### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/amoro/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Currently, when cleaning up snapshots, if the service crashes or restarts, the ongoing physical file deletion process will be interrupted and cannot resume. These service restarts are often caused by OOM issues, which may be related to an excessive number of file paths to delete (e.g., millions of expired historical files).

### How should we improve?

To address this issue, I propose the following suggestions:
1. Persist the state of expired snapshots so that the file deletion process can resume after a service restart.
2. Avoid expiring a large number of historical files all at once by performing the cleanup in batches.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #3630, opened 2025-06-19
[AMORO-3608] Using rolling batch files cleaner to reduce memory usage

xxubai:

Subtask of #3608 

## Brief change log

- As title

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/maintainer/IcebergTableMaintainer.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/maintainer/IcebergTableMaintainer.java
@@ -31,6 +31,7 @@ import org.apache.amoro.server.table.DefaultOptimizingState;
 import org.apache.amoro.server.table.DefaultTableRuntime;
 import org.apache.amoro.server.table.TableConfigurations;
 import org.apache.amoro.server.table.TableOrphanFilesCleaningMetrics;
+import org.apache.amoro.server.utils.ExpiredFileCleaner;
 import org.apache.amoro.server.utils.IcebergTableUtil;
 import org.apache.amoro.shade.guava32.com.google.common.annotations.VisibleForTesting;
 import org.apache.amoro.shade.guava32.com.google.common.base.Strings;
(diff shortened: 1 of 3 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+  public RollingFileCleaner(AuthenticatedFileIO fileIO, Set<String> excludeFiles) {
+    this.fileIO = fileIO;
+    this.excludeFiles =
+        excludeFiles != null
+            ? Sets.newConcurrentHashSet(excludeFiles)
+            : Sets.newConcurrentHashSet();
+  }
+
+  public void addFile(String filePath) {
+    if (excludeFiles.isEmpty()) {
```
zhoujinsong, line 72: Maybe we should collect the parent paths and try to delete them at last to avoid keeping themempty directorties here.

amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+      collectedFiles.add(filePath);
+      int fc = fileCounter.incrementAndGet();
+
+      if (fc % CLEANED_FILE_GROUP_SIZE == 0) {
```
zhoujinsong: It seems code 58-59 and 70-71 are duplicates; maybe we can move them out of the if-else code block.

amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java
xxubai: add remove empty dirs in `doCleanFiles`

zhoujinsong: Got it.

zhoujinsong:
It seems there are some CI issues that need to be fixed. @xxubai

PR #3639, opened 2025-06-27
[AMORO-3637] Complete task when optimizer get an exception in task execution

### LINKED ISSUE #3637: [Subtask]: Complete task when optimizer get an exception in task execution
### Search before asking

### Description
Currently, when the optimizer is encountered with an unexpected throws an exception during the execution of a task, it only logs “Optimizer executor[xx] got an unexpected error”, and doesn't call ams about the completion of the task. The task on ams is stuck in the ACK status and the optimization process is stuck at MINOR/MAJOR/FULL.

When the optimizer throws an exception during the execution of a task, it‘s better to notify ams the completion of the task.

### Parent issue

#3636

### Code of Conduct

Jzjsnow:

Close #3637.

## Brief change log

- Notify the ams of the completion of the task as long as the task is successfully acknowledged.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
@@ -241,10 +241,11 @@ public class DefaultOptimizingService extends StatedPersistentBase
   @Override
   public void completeTask(String authToken, OptimizingTaskResult taskResult) {
     LOG.info(
-        "Optimizer {} (threadId {}) complete task {}",
+        "Optimizer {} (threadId {}) complete task {}, result: {}",
         authToken,
         taskResult.getThreadId(),
-        taskResult.getTaskId());
+        taskResult.getTaskId(),
+        taskResult.getErrorMessage() == null ? "SUCCESS" : "FAIL");
     OptimizingQueue queue = getQueueByToken(authToken);
     OptimizerThread thread =
         getAuthenticatedOptimizer(authToken).getThread(taskResult.getThreadId());
--- a/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
@@ -241,7 +241,7 @@ public class DefaultOptimizingService extends StatedPersistentBase
   @Override
   public void completeTask(String authToken, OptimizingTaskResult taskResult) {
     LOG.info(
-        "Optimizer {} (threadId {}) complete task {}, result: {}",
+        "Optimizer {} (threadId {}) complete task {} (status: {})",
         authToken,
         taskResult.getThreadId(),
         taskResult.getTaskId(),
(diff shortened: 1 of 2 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -241,10 +241,11 @@ public void ackTask(String authToken, int threadId, OptimizingTaskId taskId) {
   @Override
   public void completeTask(String authToken, OptimizingTaskResult taskResult) {
     LOG.info(
-        "Optimizer {} (threadId {}) complete task {}",
+        "Optimizer {} (threadId {}) complete task {}, result: {}",
```
xxubai: The `taskOutput` in the taskResult is unreadable. Can we simplify the result information?

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
```diff
@@ -112,9 +132,10 @@ protected void completeTask(OptimizingTaskResult optimizingTaskResult) {
             return null;
           });
       LOG.info(
-          "Optimizer executor[{}] completed task[{}] to ams",
+          "Optimizer executor[{}] completed task[{}] to ams, taskResult: {}",
```
xxubai: Same above

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
Jzjsnow: Yep, `taskOutput` is not readable, and here we are more concerned about whether the task succeeded or not, which is determined by the `ErrorMessage` being null. However, the logged `"result: {}"` here may be ambiguous, I can change it to `"status: [SUCCESS|FAIL]"` for more clarity, HDYT?

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
```diff
@@ -50,8 +50,28 @@ public void start() {
       try {
         OptimizingTask task = pollTask();
         if (task != null && ackTask(task)) {
-          OptimizingTaskResult result = executeTask(task);
```
xxubai, line 53: Why do we need an additional try-catch block here, given that `executeTask(task)` already returns an error result when an exception occurs?

Jzjsnow: I remove the additional try-catch block in the new commit, and reuse the original try-catch block. The `completeTask(result)` method is now placed in the finally block to ensure that the task can complete in ams as long as the task is successfully acknowledged, regardless of whether an exception occurs during execution. PTAL

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -241,10 +241,11 @@ public void ackTask(String authToken, int threadId, OptimizingTaskId taskId) {
   @Override
   public void completeTask(String authToken, OptimizingTaskResult taskResult) {
     LOG.info(
-        "Optimizer {} (threadId {}) complete task {}",
+        "Optimizer {} (threadId {}) complete task {} (status: {})",
         authToken,
         taskResult.getThreadId(),
-        taskResult.getTaskId());
+        taskResult.getTaskId(),
+        taskResult.getErrorMessage() == null ? "SUCCESS" : "FAIL");
```
klion26, line 248: nit: we may reuse `TaskRuntime.Status.SUCCESS` and `TaskRuntime.Status.FAILED` instead of hard coding.

klion26 (approved):

PR #3657, opened 2025-07-08
Combine the functions of the executors to reduce the number of table loads

### LINKED ISSUE #3656: [Subtask]: Combine the functions of the following executors to reduce the number of table loads
### Search before asking

### Description

In this pr, I perform a functional merge of the following executors：DataExpiringExecutor, SnapshotsExpiringExecutor, OrphanFilesCleaningExecutor, DanglingDeleteExecutor

### Parent issue

#3655 

### Code of Conduct

zhangwl9:

Close #3656.

## Brief change log

- Combine the functions of the following four executors into one executor: DataExpiringExecutor, SnapshotsExpiringExecutor, OrphanFilesCleaningExecutor, DanglingDeleteExecutor
- Uniformly change the default value of the execution interval of the above executor to 4h

## How was this patch tested?

## Documentation


zhangwl9:
@xxubai @czy006 Could you please take a look at this when you're free, thanks

Aireed:
`Uniformly change the default value of the execution interval of the above executor to 1d`

Thanks for your contribution first, and I have a concern.  
It is inappropriate to unify the cleanup execution interval to 1 day for all tasks because features like expiresnapshot work in conjunction with table-level configurations (snapshot.keep.duration, default to 12hours). If the execution interval exceeds the table's configured retention period, the expected results will not be achieved.

WDYT

xxubai:
Could you explain why it's necessary to combine these four parts into one? As far as I know, they were originally a single executor and later split into multiple ones. Also, modifying configuration options is generally discouraged, as it can lead to incompatibility with old versions. 

If multiple maintenance tasks are handled by the same thread, a long-running or faulty cleanup on one table could block or delay the cleanup of other tables.

zhangwl9:
> `Uniformly change the default value of the execution interval of the above executor to 1d`
> 
> Thanks for your contribution first, and I have a concern. It is inappropriate to unify the cleanup execution interval to 1 day for all tasks because features like expiresnapshot work in conjunction with table-level configurations (snapshot.keep.duration, default to 12hours). If the execution interval exceeds the table's configured retention period, the expected results will not be achieved.
> 
> WDYT

@Aireed Yes，I have changed the default cleanup execution interval to 4 hour

zhangwl9:
> 表级配置（snapshot.keep.duration，默认为 12 小时）



> Could you explain why it's necessary to combine these four parts into one? As far as I know, they were originally a single executor and later split into multiple ones. Also, modifying configuration options is generally discouraged, as it can lead to incompatibility with old versions.你能解释为什么需要将这四个部分合并成一个吗？据我所知，它们原本是一个执行器，后来才拆分成多个。此外，修改配置选项通常是不被推荐的，因为它可能导致与旧版本的兼容性问题。
> 
> If multiple maintenance tasks are handled by the same thread, a long-running or faulty cleanup on one table could block or delay the cleanup of other tables.如果多个维护任务由同一个线程处理，一个表上的长时间运行或故障清理可能会阻塞或延迟其他表的清理。

@xxubai Before this pr, the execution of all four executors will call 'PeriodicTableScheduler#loadTable', which will result in the same table object being loaded multiple times, the metastore generating more connections, and the memory footprint spiking, leading to increased GC pressure, so I want to reduce the number of times 'PeriodicTableScheduler#loadTable' is loaded

zhangwl9:
@xxubai Do you know what the previous reason was for splitting the executor into the current 4? We found that with the current 4 executors, each executor loads the table once, which leads to more connections to the metastore, and when there are a lot of tables, it increases the memory consumption and loading latency

zhangwl9:
@xxubai  The current pr merges 4 executors into one, the parameters involved need to be merged as well, if we keep the parameters unchanged, it will become as shown below, and we need to add a comment to explain it, I'm not too sure if it's possible to use the below diagram, or not
image

zhangwl9:
@zhoujinsong @majin1102 Would you mind reviewing this pr for me in your spare time?

PR #3659, opened 2025-07-08
[AMORO-3658] Modify the execution interval of the executor from fixed to a range of random values.

### LINKED ISSUE #3658: [Subtask]: Modify the execution interval of the executor from fixed to a range of random values.
### Search before asking

### Description

 In this pr, I change the startup delay time of the executor to random values within a certain range.

### Parent issue

#3655

### Code of Conduct

zhangwl9:

Close #3658.

## Brief change log

Reduce the probability of each executor#executor() being called at the same time by changing each executor's getStartDelay() to a random value within a certain range to mitigate the dramatic increase in the number of metastore connections

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/DanglingDeleteFilesCleaningExecutor.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/DanglingDeleteFilesCleaningExecutor.java
@@ -27,6 +27,8 @@ import org.apache.amoro.server.table.TableService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.concurrent.ThreadLocalRandom;
+
 /** Clean table dangling delete files */
 public class DanglingDeleteFilesCleaningExecutor extends PeriodicTableScheduler {
 
@@ -53,6 +55,11 @@ public class DanglingDeleteFilesCleaningExecutor extends PeriodicTableScheduler 
   public void handleConfigChanged(
       DefaultTableRuntime tableRuntime, TableConfiguration originalConfig) {
     scheduleIfNecessary(tableRuntime, getStartDelay());
+  }
+
+  @Override
+  protected long getExecutorDelay() {
+    return ThreadLocalRandom.current().nextLong(INTERVAL);
   }
 
   @Override
--- a/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/PeriodicTableScheduler.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/PeriodicTableScheduler.java
@@ -154,8 +154,10 @@ public abstract class PeriodicTableScheduler extends RuntimeHandlerChain {
     logger.info("dispose thread pool for threads {}", getThreadName());
   }
 
+  protected abstract long getExecutorDelay();
+
   protected long getStartDelay() {
-    return START_DELAY;
+    return START_DELAY + getExecutorDelay();
   }
 
   protected AmoroTable<?> loadTable(DefaultTableRuntime tableRuntime) {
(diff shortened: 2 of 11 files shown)


zhangwl9:
@zhoujinsong  @shidayang  Could you please take a look at this when you're free, thanks

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/DanglingDeleteFilesCleaningExecutor.java
```diff
@@ -41,7 +43,7 @@ protected DanglingDeleteFilesCleaningExecutor(TableService tableService, int poo
 
   @Override
   protected long getNextExecutingTime(DefaultTableRuntime tableRuntime) {
-    return INTERVAL;
+    return INTERVAL + getExecutorDelay();
```
klion26: Since we have changed the `getStartDelay` to trigger the tables randomly every time, why do we need this change here?

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/DanglingDeleteFilesCleaningExecutor.java
```diff
@@ -55,6 +57,11 @@ public void handleConfigChanged(
     scheduleIfNecessary(tableRuntime, getStartDelay());
   }
 
+  @Override
+  protected long getExecutorDelay() {
+    return ThreadLocalRandom.current().nextLong(INTERVAL / 2);
```
klion26: Why do we need to to set th range to `INTERVAL / 2`

klion26 (commented):
Thanks for the contribution, I left some comments inline, please take another look when you're free, thanks.

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/DanglingDeleteFilesCleaningExecutor.java
zhangwl9: @klion26 I found that getStartDelay() is only called when the executor is started for the first time, and for subsequent cyclic execution of the executor, getNextExecutingTime() is called, intending that the trigger time of the executor at both the first start and the cyclic execution has a randomness to it, so I add the random value here as well

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/DanglingDeleteFilesCleaningExecutor.java
zhangwl9: @klion26  Because the default value of Interval is 1d, if the random range is [1d,2d], when interval=2d, the execution interval of executor is too long, and when there is a lot of cleanup data, it will lead to low execution frequency (2d once) and long execution time. Therefore, choose [1d,1.5d] to ensure that the trigger timing of executor is randomized and the execution interval is not too long.

zhangwl9:
With the current modification, although the subsequent execution interval of each executor is the same, the probability of executors being called at the same moment is greatly reduced because the start moment of each executor is a random range of values

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/DanglingDeleteFilesCleaningExecutor.java
zhangwl9: @klion26 fixup

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/DanglingDeleteFilesCleaningExecutor.java
zhangwl9: fixup

zhangwl9:
@Aireed  @baiyangtx @xxubai   Would you be able to review it for me in your spare time, thank you very much!

zhangwl9:
@Aireed I have rebased this pr

PR #3661, opened 2025-07-08
[AMORO-3660] Refactor InternalContainers to Containers and update related references

### LINKED ISSUE #3660: [Improvement]: Refactor InternalContainers to Containers
### Search before asking

In https://github.com/apache/amoro/pull/3486, an external resource container interface is introduced, which may be initialized in the same way as internal containers.

Therefore, it is necessary to unify the handling of both internal and external containers.

### How should we improve?

_No response_

### Subtasks

_No response_

### Code of Conduct

xxubai:

Close #3660.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/test/java/org/apache/amoro/server/AmsEnvironment.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/AmsEnvironment.java
@@ -28,6 +28,7 @@ import org.apache.amoro.mixed.CatalogLoader;
 import org.apache.amoro.mixed.MixedFormatCatalog;
 import org.apache.amoro.optimizer.standalone.StandaloneOptimizer;
 import org.apache.amoro.properties.CatalogMetaProperties;
+import org.apache.amoro.resource.ResourceContainer;
 import org.apache.amoro.resource.ResourceGroup;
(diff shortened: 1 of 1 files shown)


klion26:
Thanks for the contribution. Seems we need to rebase on master to fix the conflict.

zhoujinsong:
@majin1102 Can you help to review this change?

amoro-ams/src/test/java/org/apache/amoro/server/AmsEnvironment.java
```diff
@@ -310,7 +311,8 @@ public void stopOptimizer() {
         .listOptimizers()
         .forEach(
             resource -> {
-              InternalContainers.get(resource.getContainerName()).releaseResource(resource);
+              ((AbstractOptimizerContainer) Containers.get(resource.getContainerName()))
```
majin1102: I think this explicit conversion is a little unsafe. If we assure the Container got from the map always return AbstractOptimizerContainer, then we may need to modify Containers.get() return type. Otherwise I think a if instanceOf is suitable for this case.

majin1102 (approved):

PR #3664, opened 2025-07-10
Trim the leading and trailing whitespace in ResourceGroup properties

### LINKED ISSUE #3663: [Subtask]: Trim the leading and trailing whitespace in ResourceGroup properties
### Search before asking

### Description

Currently, the resourceGroup configuration items passed from the front-end may have leading and trailing spaces. If not handled, this may cause the configuration item to fail to take effect, or errors may occur when parsing the value (e.g., errors in the planInternal phase report in #3662 ). Therefore, we need to trim and merge duplicates for the resourceGroup properties.

### Parent issue

#3662 

### Code of Conduct

Jzjsnow:

Leading and trailing spaces in the resourceGroup properties may cause the configuration item to fail to take effect, or errors may occur when parsing the value.

Close #3663.

## Brief change log

- Trim and merge duplicates before persisting the resourceGroup properties

## How was this patch tested?

## Documentation


amoro-common/src/main/java/org/apache/amoro/resource/ResourceGroup.java
```diff
@@ -81,13 +84,24 @@ public String getContainer() {
...
-      this.properties.putAll(properties);
+      // Both keys and values will be trimmed of leading and trailing whitespace.
+      // Entries with empty (after trim) keys will be ignored.
+      // If duplicate keys exist after trimming, the latter value will overwrite the former.
+      this.properties.putAll(
+          properties.entrySet().stream()
+              .map(
+                  e ->
+                      new AbstractMap.SimpleEntry<>(
+                          e.getKey().trim(), e.getValue() == null ? null : e.getValue().trim()))
```
xxubai: Can you trim the properties in the controller? I think it would be better to sanitize the input values either in the frontend or within the controller layer.

Jzjsnow: Thanks for the suggestion, I have instead sanitised it in the controller and extracted the logic into PropertiesUtil for reuse.

xxubai (approved):

PR #3674, opened 2025-07-18
[Hotfix] Replace Jersey Guava imports with Amoro shaded Guava

zhongqishang:

Close #xxx.

## Brief change log

- Replace Jersey Guava imports with Amoro shaded Guava
- Add checkstyle rule

## How was this patch tested?

## Documentation

### DIFF
--- a/tools/maven/checkstyle.xml
+++ b/tools/maven/checkstyle.xml
@@ -120,6 +120,11 @@ This file is based on the checkstyle file of Apache Beam.
             <message key="import.illegal" value="{0}; Use amoro shaded guava instead."/>
         </module>
 
+        <module name="IllegalImport">
+            <property name="illegalPkgs" value="org.glassfish.jersey.internal.guava"/>
+            <message key="import.illegal" value="{0}; Use amoro shaded guava instead."/>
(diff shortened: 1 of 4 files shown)


tools/maven/checkstyle.xml
```diff
@@ -120,6 +120,11 @@ This file is based on the checkstyle file of Apache Beam.
             <message key="import.illegal" value="{0}; Use amoro shaded guava instead."/>
         </module>
 
+        <module name="IllegalImport">
```
klion26, line 123: Not sure if we need to update this with `Regex` property on?

https://checkstyle.sourceforge.io/checks/imports/illegalimport.html#Properties

zhongqishang: This package may have been imported accidentally. I think `IllegalImport` is enough. 

Of course, `Regex` can better detect the error introduction in the code. 
If need to check for incorrect references in code, it is best to use `Regex`.

klion26 (approved):
LGTM

zhoujinsong:
Thanks for the contribution! @zhongqishang

PR #3675, opened 2025-07-18
[AMORO-3673][Improvement]: Try to commit the optimizing process upon failure or cancellation

### LINKED ISSUE #3673: [Improvement]: Optimizing the submission of some successful self-optimizing tasks
### Search before asking

If some self-optimizing tasks in a table optimizing process fail or the process is canceled, the results of the successfully completed optimizing tasks will be discarded. In such cases, it is possible to submit the results of these completed tasks to avoid wasting optimizer resources.

### How should we improve?

The proposed changes are detailed in the AIP document: https://docs.google.com/document/d/143Wktb_5CmOAt-NHQoeGuP_9JUwIfc7r3_716wYTbC8/edit?usp=sharing

### Subtasks

_No response_

### Code of Conduct

cxxiii:

Close #3673 .

## Brief change log

- To prevent optimizer resource wastage, we propose submitting the results of successfully completed tasks in cases where the optimizing process fails because a task has exceeded the maximum number of retry attempts, or when the process is manually cancelled by the user.

## How was this patch tested?

## Documentation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,7 +134,7 @@
         <postgres-jdbc.version>42.7.2</postgres-jdbc.version>
         <derby-jdbc.version>10.14.2.0</derby-jdbc.version>
         <commons-dbcp2.version>2.9.0</commons-dbcp2.version>
-        <netty.version>4.1.118.Final</netty.version>
+        <netty.version>4.1.124.Final</netty.version>
         <javalin.version>4.6.8</javalin.version>
         <kyuubi-hive-jdbc-shaded.version>1.6.0-incubating</kyuubi-hive-jdbc-shaded.version>
         <rocksdb.version>7.10.2</rocksdb.version>
@@ -701,6 +701,12 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-handler</artifactId>
+                <version>${netty.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>io.netty</groupId>
+                <artifactId>netty-codec-http2</artifactId>
                 <version>${netty.version}</version>
             </dependency>
 


klion26:
@cxxiii could you please rebase the latest master, so that we can start to review this.

cxxiii:
> @cxxiii could you please rebase the latest master, so that we can start to review this.

Thank you and I have rebased the branch onto the latest master. Please proceed with the review.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingProcess.java
```diff
@@ -44,7 +44,11 @@ public interface OptimizingProcess {
 
   long getRunningQuotaTime(long calculatingStartTime, long calculatingEndTime);
 
+  boolean containSuccessTasks();
```
zhoujinsong: I am not sure if we need this public method, it seems it is only called inner the impelemenation.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingProcess.java
```diff
@@ -44,7 +44,11 @@ public interface OptimizingProcess {
...
+
   void commit();
 
+  void commitClosedProcess();
```
zhoujinsong: Can we reuse the `commit` method? I don't see the need to implement a method exclusively for the process closing case.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -507,14 +507,32 @@ private void acceptResult(TaskRuntime<?> taskRuntime) {
...
-            LOG.info(
-                "Task {} has reached the max execute retry count. Process {} failed.",
-                taskRuntime.getTaskId(),
-                processId);
-            this.failedReason = taskRuntime.getFailReason();
-            this.status = ProcessStatus.FAILED;
-            this.endTime = taskRuntime.getEndTime();
-            persistAndSetCompleted(false);
+            try {
+              buildCommit().commit();
```
zhoujinsong: I don't think we should perform the commit action directly here. Instead, we should delegate the commit process to the `OptimizingCommitExecutor`, just like how we handle normal commits currently.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingProcess.java
cxxiii: I think this public method is necessary, because we called this method through the interface in other classes.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingProcess.java
cxxiii: Your suggestion is good, I have made changes.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
cxxiii: Your suggestion is good, I have made changes.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
```diff
@@ -72,6 +72,10 @@ private TableProperties() {}
 
   public static final boolean ENABLE_SELF_OPTIMIZING_DEFAULT = true;
 
+  public static final String SELF_OPTIMIZING_COMMIT_ON_PARTIAL_SUCCESS =
+      "self-optimizing.commit-on-partial-success";
+  public static final boolean SELF_OPTIMIZING_COMMIT_ON_PARTIAL_SUCCESS_DEFAULT = true;
```
zhoujinsong: I suggest making the default value false , as it is a new feature.

zhoujinsong: And is `self-optimizing.allow-partial-commit` a proper name?

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/UnKeyedTableCommit.java
```diff
@@ -245,6 +256,17 @@ public void commit() throws OptimizingCommitException {
     }
   }
 
+  private boolean needRemove(
```
zhoujinsong: Perhaps we should proactively identify the delete files that need to be excluded in advance, rather than iterate through the task map each time to improve efficiency.

zhoujinsong (commented):
@cxxiii I left some comments, PTAL.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
cxxiii: Good suggestions.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/UnKeyedTableCommit.java
cxxiii: Ok, I will consider and modify it.

zhoujinsong (approved):
LGTM.

PR #3590, opened 2025-06-03
[AMORO-3587] Modify the default value of hive catalog clients to 20

lintingbin:
## Why are the changes needed?

Close #3587.

## Brief change log

- Fixed the default value of clients to 20 and set the clients parameter when creating an external Hive metastore catalog.
- Added description of the clients parameter in the docs/admin-guides/managing-catalogs.md document

## How was this patch tested?

- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? no
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #3678, opened 2025-07-22
[AMORO-3677] Fill catalogProperties in creating catalog rather than building

### LINKED ISSUE #3677: [Bug]: hive catalog clients properties is overridden by the default value
### What happened?

- Before creation

Image

- After creation
Image

### Affects Versions

master

Iceberg

AMS

### How to reproduce

1. create a external hive catalog and set clients to 15 in Properties.
2. save，clients is actually set to 20

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

MarigWeizhi:

Close #3677.

## Brief change log

- Set clients for hive catalog if not exists

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogBuilder.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogBuilder.java
@@ -23,8 +23,6 @@ import static org.apache.amoro.properties.CatalogMetaProperties.CATALOG_TYPE_CUS
 import static org.apache.amoro.properties.CatalogMetaProperties.CATALOG_TYPE_GLUE;
 import static org.apache.amoro.properties.CatalogMetaProperties.CATALOG_TYPE_HADOOP;
 import static org.apache.amoro.properties.CatalogMetaProperties.CATALOG_TYPE_HIVE;
-import static org.apache.amoro.properties.CatalogMetaProperties.CLIENT_POOL_SIZE;
-import static org.apache.amoro.properties.CatalogMetaProperties.CLIENT_POOL_SIZE_DEFAULT;
 
 import org.apache.amoro.TableFormat;
 import org.apache.amoro.api.CatalogMeta;
@@ -90,9 +88,6 @@ public class CatalogBuilder {
       case CATALOG_TYPE_HIVE:
         String amsUri = getAmsURI(serverConfiguration);
         catalogMeta.getCatalogProperties().put(CatalogMetaProperties.AMS_URI, amsUri);
-        catalogMeta
-            .getCatalogProperties()
-            .putIfAbsent(CLIENT_POOL_SIZE, String.valueOf(CLIENT_POOL_SIZE_DEFAULT));
         return new ExternalCatalog(catalogMeta);
       case CATALOG_TYPE_AMS:
         return new InternalCatalogImpl(catalogMeta, serverConfiguration);
--- a/amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogBuilder.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogBuilder.java
@@ -27,8 +27,6 @@ import static org.apache.amoro.properties.CatalogMetaProperties.CATALOG_TYPE_HIV
 import org.apache.amoro.TableFormat;
 import org.apache.amoro.api.CatalogMeta;
 import org.apache.amoro.config.Configurations;
-import org.apache.amoro.properties.CatalogMetaProperties;
-import org.apache.amoro.server.AmoroManagementConf;
 import org.apache.amoro.shade.guava32.com.google.common.base.Preconditions;
 import org.apache.amoro.shade.guava32.com.google.common.collect.ImmutableMap;
 import org.apache.amoro.shade.guava32.com.google.common.collect.Sets;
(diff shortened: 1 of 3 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogBuilder.java
```diff
@@ -90,9 +90,11 @@ public static ServerCatalog buildServerCatalog(
       case CATALOG_TYPE_HIVE:
         String amsUri = getAmsURI(serverConfiguration);
         catalogMeta.getCatalogProperties().put(CatalogMetaProperties.AMS_URI, amsUri);
-        catalogMeta
-            .getCatalogProperties()
-            .put(CLIENT_POOL_SIZE, String.valueOf(CLIENT_POOL_SIZE_DEFAULT));
+        if (!catalogMeta.getCatalogProperties().containsKey(CLIENT_POOL_SIZE)) {
```
klion26: If the logic here overrides the value the user set, does that mean we can safely remove this logic here?

klion26: Could we add some tests to cover this logic?

amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogBuilder.java
```diff
@@ -90,9 +90,11 @@ public static ServerCatalog buildServerCatalog(
       case CATALOG_TYPE_HIVE:
         String amsUri = getAmsURI(serverConfiguration);
         catalogMeta.getCatalogProperties().put(CatalogMetaProperties.AMS_URI, amsUri);
-        catalogMeta
-            .getCatalogProperties()
-            .put(CLIENT_POOL_SIZE, String.valueOf(CLIENT_POOL_SIZE_DEFAULT));
```
zhongqishang, line 95: Maybe just `put` -> `putIfAbsent`.

MarigWeizhi:
@klion26 @zhongqishang Thanks for the review, I updated the code and added a test.

klion26:
@MarigWeizhi Thanks for the follow-up, IIUC, we'll merge the property when constructing the catalog [here](https://github.com/apache/amoro/blob/02a2c1fb2c940a84c310e4e336bab0e25d0b20fa/amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/CatalogController.java#L416). Could you please help verify that it runs as  expected after we remove the logic here(line 93-95 of `amoro-ams/src/main/java/org/apache/amoro/server/catalog/CatalogBuilder.java`), which will update the property.

MarigWeizhi:
@klion26 Thank you for your reply.   
I think these three lines of code `CatalogBuilder.java (lines 93–95)` is to allow users to see the default value of `clients` in the Web UI even when it is not explicitly set, making it easier to view and modify.

I tested two case after removing these three lines in `CatalogBuilder.java` (lines 93–95):

1. **When creating a catalog without configuring `clients`**:
   - The catalog does **not** include the `clients` parameter.
   - As a result, the `clients` field is **not visible** in the Web UI.
   - Eventually, the default value of `20` is used at runtime, as defined here:  
     [CachedHiveClientPool.java#L54](https://github.com/apache/amoro/blob/master/amoro-common/src/main/java/org/apache/amoro/hive/CachedHiveClientPool.java#L54)

   Screenshot 1

2. **When creating a catalog with `clients=15`**:
   - The catalog correctly uses the configured value of `15`.
   - The `clients` parameter is displayed and editable in the Web UI.

   Screenshot 2

klion26:
@MarigWeizhi Thanks for the verification. Seems the result is as expected?
1. If the property was not set, then Amoro will use the default parameter, and the value was not displayed in the ui is as expected, as the user hasn't set it, but maybe we can log the value we used in the log or somewhere
2. If a user sets the property, then Amoro will use the user-set property

I prefer to handle the configuration in a single place, so that we can maintain it easily. Please let me know what you think about this? thanks.

MarigWeizhi:
@klion26 Thank you for your suggestion. I think it's a good idea to control the parameters in one place. I have updated the code and title of this PR.

zhoujinsong:
It seems this PR just rolls back the change produced by #3590.

How can we make sure the default client is 20 now?

MarigWeizhi:
> It seems this PR just rolls back the change produced by #3590.
> 
> How can we make sure the default client is 20 now?

@zhoujinsong 
This PR rolled back part of #3590, to avoid setting default values in two places.  The default value remains 20.  
If the user has not set the clients, a default value will be set at [CachedHiveClientPool.java#L54](https://github.com/apache/amoro/blob/master/amoro-common/src/main/java/org/apache/amoro/hive/CachedHiveClientPool.java#L54)
and default value is still 20 [CatalogMetaProperties.java#L67](https://github.com/apache/amoro/blob/master/amoro-common/src/main/java/org/apache/amoro/properties/CatalogMetaProperties.java#L67)

zhoujinsong:
It seems only the MixedHiveCatalog and HudiHiveCatalog are dependent on [CachedHiveClientPool.java#L54](https://github.com/apache/amoro/blob/master/amoro-common/src/main/java/org/apache/amoro/hive/CachedHiveClientPool.java#L54).

I am afraid the Iceberg Hive Catalog does not work with it.

zhoujinsong:
Besides, we may add the `clients` properties to the catalog meta in the creating catalog process rather than the build catalog method.

It may resolve the issue.

MarigWeizhi:
> Besides, we may add the `clients` properties to the catalog meta in the creating catalog process rather than the build catalog method.
> 
> It may resolve the issue.

@zhoujinsong 
Thank you for your suggestion. I updated the code, I set amsUri and clients when creating catalog . Please take a look when you have time.

zhoujinsong (approved):
LGTM.

PR #3701, opened 2025-07-31
[Improvement] Use fileIndex="nomax" to avoid log loss.

CatchYouIfICan:

[Improvement] Use fileIndex="nomax" to avoid log loss.

## Brief change log

-[Improvement] Use fileIndex="nomax" to avoid log loss.

## How was this patch tested?

## Documentation

### DIFF
--- a/dist/src/main/amoro-bin/conf/log4j2.xml
+++ b/dist/src/main/amoro-bin/conf/log4j2.xml
@@ -61,7 +61,7 @@
                 <!--Set the log base file size to trigger rolling log file updates if the size is exceeded-->
                 <SizeBasedTriggeringPolicy size="2048MB"/>
             </Policies>
-            <DefaultRolloverStrategy max="7">
+            <DefaultRolloverStrategy fileIndex="nomax">
                 <Delete basePath="${LOG_HOME}/history" maxDepth="1">
                     <IfFileName glob="*.log.gz"/>
                     <IfAny>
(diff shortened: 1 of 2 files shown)


dist/src/main/amoro-bin/conf/log4j2.xml
```diff
@@ -44,7 +44,7 @@
                 <!--Set the log base file size to trigger rolling log file updates if the size is exceeded-->
                 <SizeBasedTriggeringPolicy size="2048MB"/>
             </Policies>
-            <DefaultRolloverStrategy max="25"/>
+            <DefaultRolloverStrategy fileIndex="nomax"/>
```
zhoujinsong: We may not change here; we should limit the debug archive logs.

dist/src/main/amoro-bin/conf/log4j2.xml
```diff
@@ -86,7 +86,7 @@
                 <!--Set the log base file size to trigger rolling log file updates if the size is exceeded-->
                 <SizeBasedTriggeringPolicy size="64MB"/>
             </Policies>
-            <DefaultRolloverStrategy max="7"/>
+            <DefaultRolloverStrategy fileIndex="nomax"/>
```
zhoujinsong: Similar to the debug appender, we should not change the parameter here.

dist/src/main/amoro-bin/conf/log4j2.xml
```diff
@@ -99,7 +99,7 @@
             <Policies>
                 <SizeBasedTriggeringPolicy size="64MB"/>
             </Policies>
-            <DefaultRolloverStrategy max="7"/>
+            <DefaultRolloverStrategy fileIndex="nomax"/>
```
zhoujinsong: We should not change the parameter here, too

zhoujinsong (commented):
Hi, we should change the info appender only as it has a size-based rollover strategy.

CatchYouIfICan:
> Hi, we should change the info appender only as it has a size-based rollover strategy.

corrected

zhoujinsong (approved):
LGTM.

PR #3710, opened 2025-08-04
[Improvement] Add `.asf.yaml`

zhoujinsong:

Following the https://github.com/apache/amoro/discussions/3649, this PR add `.asf.yaml` to change the labels

## Brief change log

- Add `.asf.yaml` to change lables

## How was this patch tested?

## Documentation

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,47 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -0,0 +1,47 @@
...
+# See: https://github.com/apache/spark/blob/master/.asf.yaml
+
+github:
+  description: "Apache Amoro(incubating) is a Lakehouse management system built on open data lake formats."
+  homepage: https://amoro.apache.org/
+  labels:
+    - big-data
+    - data-lake
+    - lake-house
+    - data-ingestion
```
xxubai: I believe we donot have `data-ingestion` model in amoro. WDYT

.asf.yaml
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+#
+
+# See: https://github.com/apache/spark/blob/master/.asf.yaml
+
+github:
+  description: "Apache Amoro(incubating) is a Lakehouse management system built on open data lake formats."
+  homepage: https://amoro.apache.org/
+  labels:
+    - big-data
```
xxubai, line 24: I think it's a bit too broad.

.asf.yaml
zhoujinsong: Yes, it should be removed.

.asf.yaml
zhoujinsong: I think it expresses the category of this project.

zhoujinsong:
Thanks for reviewing! @xxubai

PR #3726, opened 2025-08-12
[AMORO-3676][Improvement]: use self-optimizing.min-plan-interval to skip some redundant evaluator

### LINKED ISSUE #3676: [Improvement]: use self-optimizing.min-plan-interval to skip some redundant evaluator
### Search before asking

The default interval for `tableRuntimeRefresh` is 1 minute.
In certain scenarios, we increase the optimize interval by setting `'self-optimizing.min-plan-interval'` to a value greater than 1 minute. 
As a result, when scheduling table optimization, the table might be skipped because the interval is shorter than `'self-optimizing.min-plan-interval'`. Consequently, the pendingInput generated in this round becomes useless.

Image

### How should we improve?

add filter to skip some redundant evalutor
Image

### Subtasks

_No response_

### Code of Conduct

Aireed:

Close #3676 .

## Brief change log

-  Before the evaluator, check if the plan interval is greater than the minimum value.

## How was this patch tested?

## Documentation


amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -63,8 +63,14 @@ protected long getNextExecutingTime(TableRuntime tableRuntime) {
 
   private void tryEvaluatingPendingInput(DefaultTableRuntime tableRuntime, MixedTable table) {
     DefaultOptimizingState optimizingState = tableRuntime.getOptimizingState();
+    // When planning tasks, tables may be ignored if their plan intervals are too small,
```
klion26: Can we combine `min-plan-interval` into the execute interval, or in the first place of `execute` in the current class?

I think we shouldn't schedule the evaluation if we'll abort it in the future.

Aireed:
> @Aireed thanks for your contribution. I left one inline comment. Please let me know what you think about it.

Thansk for your review.
TableRuntimeRefreshExecutor indeed involves more than just executing evaluations (evaluator); it also encompasses other logic, such as updating properties. Therefore, it cannot simply be regarded as an execute interval.

i discussed with @zhoujinsong  offline, we agreed that the evaluator should only be triggered when the optimize state is idle, as this makes it easier to understand

WDYT?

zhoujinsong:
> I discussed with @zhoujinsong offline, we agreed that the evaluator should only be triggered when the optimize state is idle, as this makes it easier to understand.

cc @Jzjsnow

Issue #3734, opened 2025-08-19
[Improvement]: Add operation/schemaId to Snapshot


xuzifu666:
### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/amoro/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Snapshot should provide users with relevant information such as operation and schema, which will bring useful improvements to visualization.

### How should we improve?

Add operation/schemaId information to the snapshot.

### Are you willing to submit PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

PR #3735, opened 2025-08-19
[AMORO-3734][Improvement]: Add operation/schemaId to Snapshot

xuzifu666:

Close https://github.com/apache/amoro/issues/3734.

## Brief change log

Snapshot should provide users with relevant information such as operation and schema, which will bring useful improvements to visualization.

### DIFF
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedSnapshot.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedSnapshot.java
@@ -42,10 +42,15 @@ public class MixedSnapshot implements TableSnapshot {
 
   @Override
   public long commitTime() {
-    Long changCommit =
-        Optional.ofNullable(changeSnapshot).map(Snapshot::timestampMillis).orElse(-1L);
-    Long baseCommit = Optional.ofNullable(baseSnapshot).map(Snapshot::timestampMillis).orElse(-1L);
-    return Longs.max(changCommit, baseCommit);
+    return Longs.max(getChangeCommit(), getBaseCommit());
+  }
+
+  public long getChangeCommit() {
+    return Optional.ofNullable(changeSnapshot).map(Snapshot::timestampMillis).orElse(-1L);
+  }
+
+  public long getBaseCommit() {
+    return Optional.ofNullable(baseSnapshot).map(Snapshot::timestampMillis).orElse(-1L);
   }
 
   public Snapshot getChangeSnapshot() {
@@ -67,5 +72,21 @@ public class MixedSnapshot implements TableSnapshot {
   @Override
   public String id() {
     return changeSnapshot + "_" + baseSnapshot;
+  }
+
+  @Override
+  public String operation() {
+    if (getChangeCommit() >= getBaseCommit()) {
+      return changeSnapshot.operation();
+    }
+    return baseSnapshot.operation();
+  }
+
+  @Override
+  public long schemaId() {
+    if (getChangeCommit() >= getBaseCommit()) {
+      return changeSnapshot.schemaId();
+    }
+    return baseSnapshot.schemaId();
   }
 }
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedSnapshot.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedSnapshot.java
@@ -45,11 +45,11 @@ public class MixedSnapshot implements TableSnapshot {
     return Longs.max(getChangeCommit(), getBaseCommit());
   }
 
-  public long getChangeCommit() {
+  private long getChangeCommit() {
     return Optional.ofNullable(changeSnapshot).map(Snapshot::timestampMillis).orElse(-1L);
   }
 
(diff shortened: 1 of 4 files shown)


amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedSnapshot.java
```diff
@@ -42,10 +42,15 @@ public long watermark() {
...
   @Override
   public long commitTime() {
-    Long changCommit =
-        Optional.ofNullable(changeSnapshot).map(Snapshot::timestampMillis).orElse(-1L);
-    Long baseCommit = Optional.ofNullable(baseSnapshot).map(Snapshot::timestampMillis).orElse(-1L);
-    return Longs.max(changCommit, baseCommit);
+    return Longs.max(getChangeCommit(), getBaseCommit());
+  }
+
+  public long getChangeCommit() {
```
xxubai: Why is this function declared as public?

If this method is only meant for internal usage, consider using private or protected to better encapsulate the implementation details.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedSnapshot.java
```diff
@@ -42,10 +42,15 @@ public long watermark() {
...
+    return Optional.ofNullable(changeSnapshot).map(Snapshot::timestampMillis).orElse(-1L);
+  }
+
+  public long getBaseCommit() {
```
xxubai: Same above

xxubai (commented):
Thanks for the contribution! 
Could you explain why we need to expose operation and schemaId? In which cases will they be used? It would be helpful if you could also provide some typical usage scenarios so we can better understand the design intent.

xuzifu666:
> Thanks for the contribution! Could you explain why we need to expose operation and schemaId? In which cases will they be used? It would be helpful if you could also provide some typical usage scenarios so we can better understand the design intent.

Thanks for your attention @xxubai . As I understand it, for example, if a user task is write-only, they need to retrieve the operation from the system table or snapshot file to determine whether their compaction task has taken effect. This needs to be obtained from the operation. Furthermore, during schema evolution, users also need to determine whether their field changes have taken effect based on the schemaId. This should be useful in actual visual monitoring or troubleshooting.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedSnapshot.java
xuzifu666: ditto

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/mixed/MixedSnapshot.java
xuzifu666: Thanks for reminder， changed it to private.

zhangwl9:
> > Thanks for the contribution! Could you explain why we need to expose operation and schemaId? In which cases will they be used? It would be helpful if you could also provide some typical usage scenarios so we can better understand the design intent.
> 
> Thanks for your attention @xxubai . As I understand it, for example, if a user task is write-only, they need to retrieve the operation from the system table or snapshot file to determine whether their compaction task has taken effect. This needs to be obtained from the operation. Furthermore, during schema evolution, users also need to determine whether their field changes have taken effect based on the schemaId. This should be useful in actual visual monitoring or troubleshooting.

Where will this operation and schemaId be displayed for the user to see, or does the user need to get it via restapi?

xuzifu666:
> > > Thanks for the contribution! Could you explain why we need to expose operation and schemaId? In which cases will they be used? It would be helpful if you could also provide some typical usage scenarios so we can better understand the design intent.
> > 
> > Thanks for your attention @xxubai . As I understand it, for example, if a user task is write-only, they need to retrieve the operation from the system table or snapshot file to determine whether their compaction task has taken effect. This needs to be obtained from the operation. Furthermore, during schema evolution, users also need to determine whether their field changes have taken effect based on the schemaId. This should be useful in actual visual monitoring or troubleshooting.
> 
> Where will this operation and schemaId be displayed for the user to see, or does the user need to get it via restapi?

Yes，when obtaining snapshot related information through the HTTP interface, it should be possible to obtain these newly added information.

xxubai (commented):
This change exposes two interfaces, which indeed provide more detailed snapshot information. 
However, I’m not entirely sure how these interfaces will be used in the future. Do you have any concrete plans for follow-up implementations? Maybe we can discuss this first.

xuzifu666:
> This change exposes two interfaces, which indeed provide more detailed snapshot information. However, I’m not entirely sure how these interfaces will be used in the future. Do you have any concrete plans for follow-up implementations? Maybe we can discuss this first.

As I mentioned earlier, the specific production value lies in the fact that during visual monitoring, users can obtain information such as the snapshot operation type and schema ID changes caused by schema evolution, which greatly facilitates monitoring and alerting. Other information can also be added later, such as the number of data entries and the number of changes in the snapshot, which will improve operational capabilities such as table management. @xxubai

zhoujinsong:
This change looks good to me. Maybe we can discuss how to display this information in the Web UI later.

PR #3743, opened 2025-08-21
[AMORO-3552] Optimize binary distribution NOTICE&LICENSE file 

### LINKED ISSUE #3552: [Improvement]: optimize  binary distribution NOTICE
### Search before asking

optimize  binary distribution NOTICE

### How should we improve?

optimize  binary distribution NOTICE

### Subtasks

_No response_

### Code of Conduct

zhoujinsong:

Close #3552.

## Brief change log

- as titled.

## How was this patch tested?

## Documentation

### DIFF
--- a/NOTICE-binary
+++ b/NOTICE-binary
@@ -42,6 +42,42 @@ Copyright 2005-2024 The Apache Software Foundation
 Apache Commons Lang
 Copyright 2001-2024 The Apache Software Foundation
 
+Apache Commons Pool
+Copyright 2001-2024 The Apache Software Foundation
+
+Apache Commons Math
+Copyright 2001-2024 The Apache Software Foundation
+
+Apache Commons Collections
+Copyright 2001-2024 The Apache Software Foundation
+
+Apache Commons Configuration
+Copyright 2001-2024 The Apache Software Foundation
+
+Apache Commons Text
+Copyright 2014-2024 The Apache Software Foundation
+
+Apache Commons Compress
+Copyright 2002-2024 The Apache Software Foundation
+
+Apache Commons IO
+Copyright 2002-2024 The Apache Software Foundation
+
+Apache Commons Net
+Copyright 2001-2024 The Apache Software Foundation
+
+Apache Commons BeanUtils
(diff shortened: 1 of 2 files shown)


NOTICE-binary
```diff
@@ -42,6 +42,42 @@ Copyright 2005-2024 The Apache Software Foundation
 Apache Commons Lang
 Copyright 2001-2024 The Apache Software Foundation
 
+Apache Commons Pool
```
klion26: IIUC, we don't need to mention the ASF project here?

```
- binary distribution NOTICE should content the NOTICE from ALv2
project (non ASF), for instance Jackson, AWS SDK, Netty, etc. That
would be great to double check that for next releases.
```

klion26: And we can remove the ASF project listed in the `NOTICE` file

```
- NOTICE is OK. Nit: the NOTICE can be lighter: no need to mention the
ASF projects here (Iceberg, Paimon, Flink)
```

klion26 (commented):
@zhoujinsong thanks for the contribution, left some inline comments, please let me know what you think about it.

NOTICE-binary
zhoujinsong: Yes, it is fine to exclude the ASF projects, but I still prefer to include it here, like many other projects did:

- https://github.com/apache/gravitino/blob/main/NOTICE.bin
- https://github.com/apache/spark/blob/master/NOTICE-binary

Besides, some projects did not like:
- https://github.com/apache/flink/blob/master/NOTICE

klion26: I'm fine with either way, but as the [asf website](https://infra.apache.org/licensing-howto.html#mod-notice) said, so maybe removing the ASF project here is better.

```
It is important to keep NOTICE as brief and simple as possible, as each addition places a burden on downstream consumers.

**Do not** add anything to NOTICE which is not legally required.
```

zhoujinsong: Make sense to me, I will remove the ASF projects.

zhoujinsong: @klion26 I have removed the ASF projects here, PTAL.

PR #3764, opened 2025-09-04
[Amoro-3762][Subtask] Change TableService to SPI mode to support expansion of other implementation methods.

wardlican:

Close #xxx.

## Brief change log
Solution Overview

Added a new SPI factory interface, TableServiceProvider (based on ServiceLoader).
Provides a default implementation, DefaultTableServiceProvider, and registers it in META-INF/services.
Added a new configuration option, TABLE_SERVICE_IMPL (string), with two supported syntaxes:
-- Specify the provider name (e.g., default)
-- Specify the implementation class name (e.g., org.apache.amoro.server.table.DefaultTableService)
Creates a TableService in the AmoroServiceContainer using a configuration-based loader method.
Retains the existing default behavior: Falling back to DefaultTableService when not configured.

-

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/TableService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/TableService.java
@@ -21,6 +21,7 @@ package org.apache.amoro.server.table;
 import org.apache.amoro.AmoroTable;
 import org.apache.amoro.ServerTableIdentifier;
 import org.apache.amoro.TableRuntime;
+import org.apache.amoro.server.catalog.ExternalCatalog;
 import org.apache.amoro.server.catalog.InternalCatalog;
 
 public interface TableService extends TableRuntimeHandler {
@@ -46,4 +47,16 @@ public interface TableService extends TableRuntimeHandler {
    * @return managed table.
    */
   AmoroTable<?> loadTable(ServerTableIdentifier identifier);
+
+  /** Explore and synchronize table runtimes from catalogs. Intended for periodic sync and tests. */
+  void exploreTableRuntimes();
+
+  /** Explore and synchronize a specific external catalog. */
+  void exploreExternalCatalog(ExternalCatalog externalCatalog);
+
+  /** Set or replace a runtime for testing or recovery scenarios. */
+  void setRuntime(DefaultTableRuntime tableRuntime);
+
+  /** Dispose a managed table and its runtime. */
+  void disposeTable(ServerTableIdentifier tableIdentifier);
 }
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
@@ -233,6 +233,13 @@ public class AmoroManagementConf {
           .defaultValue(1260)
           .withDescription("Port that the table service thrift server is bound to.");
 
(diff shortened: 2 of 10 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/table/TableService.java
```diff
@@ -46,4 +47,16 @@ default boolean contains(Long tableId) {
    * @return managed table.
    */
   AmoroTable<?> loadTable(ServerTableIdentifier identifier);
+
+  /** Explore and synchronize table runtimes from catalogs. Intended for periodic sync and tests. */
+  void exploreTableRuntimes();
```
baiyangtx, line 52: Who will call those methods?

baiyangtx (commented):
Why use the SPI approach to give TableService different implementations instead of directly adding multi-node scaling to the current TableService? Whether to deploy in multi-node mode should be controlled by a simple configuration flag; it shouldn’t require a completely separate TableService implementation.

wardlican:
> t TableService? Whether to deploy in multi-node mode should be controlled by a simple configuration flag; it shouldn’t require a completely separate TableService implementation.

To distinguish between the master-slave mode and the master-slave mode, the newly added master-slave mode should not affect the original logic. The master-slave mode will make more logical changes in TableServcie.

amoro-ams/src/main/java/org/apache/amoro/server/table/TableService.java
wardlican: This method is mentioned in the interface for compatibility with the original test cases in SPI mode, and secondly, this method also needs to be implemented in master-slave mode. @baiyangtx @zhoujinsong

baiyangtx:
> To distinguish between the master-slave mode and the master-slave mode, the newly added master-slave mode should not affect the original logic. The master-slave mode will make more logical changes in TableServcie.

Will the two implementations be merged in the future?
In the long run, the horizontally-scalable version is likely to become the default. If we do expect a single unified implementation eventually, we could simply switch between the two code paths with a config flag today instead of introducing a full SPI framework. After all, TableService is not intended to be opened up for third-party plug-ins down the road.

WDYT @zhoujinsong @Aireed

wardlican:
> > To distinguish between the master-slave mode and the master-slave mode, the newly added master-slave mode should not affect the original logic. The master-slave mode will make more logical changes in TableServcie.
> 
> Will the two implementations be merged in the future? In the long run, the horizontally-scalable version is likely to become the default. If we do expect a single unified implementation eventually, we could simply switch between the two code paths with a config flag today instead of introducing a full SPI framework. After all, TableService is not intended to be opened up for third-party plug-ins down the road.
> 
> WDYT @zhoujinsong @Aireed

Yes, the design here needs to decide whether to implement compatible logic through configuration in DefaultTableService.

zhoujinsong:
I'm also hesitant about whether it's necessary to add SPI extension. Perhaps it would be better if the existing implementation simply supports horizontal scaling, as the master-slave architecture will eventually be replaced by a multiple-master architecture.

wardlican:
> I'm also hesitant about whether it's necessary to add SPI extension. Perhaps it would be better if the existing implementation simply supports horizontal scaling, as the master-slave architecture will eventually be replaced by a multiple-master architecture.

OK, here i will change it to implement the master-slave mode logic in DefaultTableService through configuration.

PR #3768, opened 2025-09-05
[AMORO-3754]failed to failover when only one node is alive in ha mode #3754

### LINKED ISSUE #3754: [Bug]: failed to failover when only one node is alive in ha mode
### What happened?

Image

### Affects Versions

0.8.x

Iceberg

AMS

### How to reproduce

1. enable ha mode
2. keep only one ams instance 
3. Simulate ZooKeeper session timeout
4. may be ten tables are written, and optimizing is processing.
This is intermittent and does not occur consistently.

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

Aireed:

Close #3754 .

## Brief change log

- call stop regardless of whether the service is serving
- unregister metric key before registration

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
@@ -205,12 +205,13 @@ public class AmoroServiceContainer {
   }
 
   public void disposeOptimizingService() {
-    if (tableManagementServer != null && tableManagementServer.isServing()) {
-      LOG.info("Stopping table management server...");
+    if (tableManagementServer != null) {
+      LOG.info(
+          "Stopping table management server[serving:{}] ...", tableManagementServer.isServing());
       tableManagementServer.stop();
     }
-    if (optimizingServiceServer != null && optimizingServiceServer.isServing()) {
-      LOG.info("Stopping optimizing server...");
+    if (optimizingServiceServer != null) {
+      LOG.info("Stopping optimizing server[serving:{}] ...", optimizingServiceServer.isServing());
       optimizingServiceServer.stop();
     }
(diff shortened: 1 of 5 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/metrics/MetricRegistry.java
```diff
@@ -78,6 +78,7 @@ public <T extends Metric> MetricKey register(
         define.getName());
 
     MetricKey key = new MetricKey(define, tags);
+    unregister(key);
```
baiyangtx: I believe an `unregister` method should be provided and called by the metrics registrant during disposal, rather than performing the unregister action at registration time.

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
```diff
@@ -200,11 +200,11 @@ private void addHandlerChain(RuntimeHandlerChain chain) {
   }
 
   public void disposeOptimizingService() {
-    if (tableManagementServer != null && tableManagementServer.isServing()) {
+    if (tableManagementServer != null) {
```
zhoujinsong, line 208: It seems the codebase is out of date, comparing with the master branch:https://github.com/apache/amoro/blob/master/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java#L208

amoro-ams/src/main/java/org/apache/amoro/server/metrics/MetricRegistry.java
zhoujinsong: +1

PR #3776, opened 2025-09-09
[AMORO-3775] Add support for metric-based refresh event trigger in TableRuntimeRefreshExecutor

### LINKED ISSUE #3775: [Subtask]: Add support for pluggable refresh event in TableRuntimeRefreshExecutor and DefaultRefreshEvent
### Search before asking

### Description

Through SPI mode, make the refresh and evaluation process in TableRuntimeRefreshExecutor a pluggable refresh event, and define the existing periodic scanning and evaluation as the default refresh event.

### Parent issue

#3774 

### Code of Conduct

Jzjsnow:

Close #3775.

## Brief change log

Add support for MSE based refresh event:  
- Support for calculating partition filesize mean square error based on the loaded metadata.
- Filter partitions need to be optimized based on threshold and trigger pendingInput evaluation if necessary.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
@@ -20,7 +20,9 @@ package org.apache.amoro.server.scheduler.inline;
 
 import org.apache.amoro.AmoroTable;
 import org.apache.amoro.TableRuntime;
+import org.apache.amoro.config.OptimizingConfig;
 import org.apache.amoro.config.TableConfiguration;
+import org.apache.amoro.optimizing.evaluation.MetadataBasedEvaluationEvent;
 import org.apache.amoro.optimizing.plan.AbstractOptimizingEvaluator;
 import org.apache.amoro.process.ProcessStatus;
 import org.apache.amoro.server.optimizing.OptimizingProcess;
@@ -60,8 +62,19 @@ public class TableRuntimeRefreshExecutor extends PeriodicTableScheduler {
 
   private void tryEvaluatingPendingInput(DefaultTableRuntime tableRuntime, MixedTable table) {
     // only evaluate pending input when optimizing is enabled and in idle state
-    if (tableRuntime.getTableConfiguration().getOptimizingConfig().isEnabled()
+    OptimizingConfig optimizingConfig = tableRuntime.getOptimizingConfig();
+    if (optimizingConfig.isEnabled()
         && tableRuntime.getOptimizingStatus().equals(OptimizingStatus.IDLE)) {
+
+      if (optimizingConfig.isMetadataBasedTriggerEnabled()
+          && !MetadataBasedEvaluationEvent.isEvaluatingNecessary(
+              optimizingConfig, table, tableRuntime.getLastPlanTime())) {
+        logger.debug(
+            "{} optimizing is not necessary due to metadata based trigger",
+            tableRuntime.getTableIdentifier());
+        return;
+      }
+
       AbstractOptimizingEvaluator evaluator =
           IcebergTableUtil.createOptimizingEvaluator(tableRuntime, table, maxPendingPartitions);
       if (evaluator.isNecessary()) {
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
@@ -22,6 +22,7 @@ import org.apache.amoro.ServerTableIdentifier;
 import org.apache.amoro.config.OptimizingConfig;
 import org.apache.amoro.optimizing.HealthScoreInfo;
 import org.apache.amoro.optimizing.OptimizingType;
+import org.apache.amoro.optimizing.evaluation.MetadataBasedEvaluationEvent;
 import org.apache.amoro.shade.guava32.com.google.common.base.MoreObjects;
 import org.apache.amoro.shade.guava32.com.google.common.base.Preconditions;
 import org.apache.amoro.shade.guava32.com.google.common.collect.Sets;
@@ -47,6 +48,7 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
   protected final OptimizingConfig config;
   protected final long lastFullOptimizingTime;
   protected final long lastMinorOptimizingTime;
+  protected final long lastMajorOptimizingTime;
   protected final long fragmentSize;
   protected final long minTargetSize;
   protected final long planTime;
@@ -80,6 +82,9 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
   protected long posDeleteFileSize = 0L;
   protected long posDeleteFileRecords = 0L;
 
+  // mse stat
+  protected long fileSizeSquaredErrorSum = 0L;
+
   private long cost = -1;
   private Boolean necessary = null;
   private OptimizingType optimizingType = null;
@@ -91,7 +96,8 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
       Pair<Integer, StructLike> partition,
       long planTime,
       long lastMinorOptimizingTime,
-      long lastFullOptimizingTime) {
+      long lastFullOptimizingTime,
+      long lastMajorOptimizingTime) {
     this.identifier = identifier;
     this.config = config;
     this.partition = partition;
@@ -104,6 +110,7 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
     }
     this.planTime = planTime;
     this.lastMinorOptimizingTime = lastMinorOptimizingTime;
+    this.lastMajorOptimizingTime = lastMajorOptimizingTime;
     this.lastFullOptimizingTime = lastFullOptimizingTime;
     this.reachFullInterval =
         config.getFullTriggerInterval() >= 0
@@ -128,6 +135,11 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
     if (!config.isEnabled()) {
       return false;
     }
+    if (config.isMetadataBasedTriggerEnabled() && config.getEvaluationMseTolerance() > 0) {
+      // Update the file size squared error sum
+      updateFileSizeSquaredErrorSum(dataFile);
+    }
+
     if (isFragmentFile(dataFile)) {
       return addFragmentFile(dataFile, deletes);
     } else if (isUndersizedSegmentFile(dataFile)) {
@@ -206,6 +218,31 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
     }
 
     return false;
+  }
+
+  private void updateFileSizeSquaredErrorSum(DataFile dataFile) {
+    // Only accumulate squared error for files smaller than `minTargetSize`
+    // For files larger than or equal to minTargetSize, diffSize will be 0, contributing nothing to
+    // the error sum
+    long diffSize = minTargetSize - Math.min(dataFile.fileSizeInBytes(), minTargetSize);
+    if (diffSize <= 0) {
+      return;
+    }
+
+    // Prevent overflow for diffSize * diffSize by saturating to Long.MAX_VALUE
+    final long prod;
+    if (diffSize > Long.MAX_VALUE / diffSize) {
+      prod = Long.MAX_VALUE;
+    } else {
+      prod = diffSize * diffSize;
+    }
+
+    // Prevent overflow when adding to fileSizeSquaredErrorSum (saturate to Long.MAX_VALUE)
+    if (fileSizeSquaredErrorSum > Long.MAX_VALUE - prod) {
+      fileSizeSquaredErrorSum = Long.MAX_VALUE;
+    } else {
+      fileSizeSquaredErrorSum += prod;
+    }
   }
 
   protected boolean fileShouldFullOptimizing(DataFile dataFile, List<ContentFile<?>> deleteFiles) {
@@ -288,6 +325,19 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
   @Override
   public boolean isNecessary() {
     if (necessary == null) {
+      long lastPlanTime =
+          Math.max(
+              Math.max(lastMinorOptimizingTime, lastMajorOptimizingTime), lastFullOptimizingTime);
+      if (config.isMetadataBasedTriggerEnabled()
+          && !MetadataBasedEvaluationEvent.isReachFallbackInterval(config, lastPlanTime)) {
+        long fileCount = fragmentFileCount + undersizedSegmentFileCount;
+        if (!MetadataBasedEvaluationEvent.isPartitionPendingNecessary(
+            config, fileSizeSquaredErrorSum, fileCount)) {
+          LOG.debug("{} not necessary due to metadata-based evaluation, {}", name(), this);
+          necessary = false;
+          return false;
+        }
+      }
       if (isFullOptimizing()) {
         necessary = isFullNecessary();
       } else {
@@ -509,6 +559,10 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
     return posDeleteFileRecords;
   }
 
+  public long getFileSizeSquaredErrorSum() {
+    return fileSizeSquaredErrorSum;
+  }
+
   public static class Weight implements PartitionEvaluator.Weight {
 
     private final long cost;
@@ -532,6 +586,7 @@ public class CommonPartitionEvaluator implements PartitionEvaluator {
         .add("undersizedSegmentSize", minTargetSize)
         .add("planTime", planTime)
         .add("lastMinorOptimizeTime", lastMinorOptimizingTime)
+        .add("lastMajorOptimizeTime", lastMajorOptimizingTime)
         .add("lastFullOptimizeTime", lastFullOptimizingTime)
         .add("fragmentFileCount", fragmentFileCount)
         .add("fragmentFileSize", fragmentFileSize)
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
@@ -133,6 +133,19 @@ public class TableProperties {
       "self-optimizing.min-plan-interval";
   public static final long SELF_OPTIMIZING_MIN_PLAN_INTERVAL_DEFAULT = 60000;
 
+  /** metadata-based evaluation related properties */
+  public static final String SELF_OPTIMIZING_EVALUATION_FALLBACK_INTERVAL =
+      "self-optimizing.evaluation.fallback-interval"; // fallback evaluation interval in
+  // milliseconds
+
+  public static final int SELF_OPTIMIZING_EVALUATION_FALLBACK_INTERVAL_DEFAULT =
+      -1; // metadata-based evaluation not in effect
+
+  public static final String SELF_OPTIMIZING_EVALUATION_FILE_SIZE_MSE_TOLERANCE =
+      "self-optimizing.evaluation.file-size.mse-tolerance"; // Mean Squared Error (MSE) tolerance
+  // for file size evaluation
+  public static final long SELF_OPTIMIZING_EVALUATION_FILE_SIZE_MSE_TOLERANCE_DEFAULT = 0;
+
   /** table clean related properties */
   public static final String ENABLE_TABLE_EXPIRE = "table-expire.enabled";
 
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
@@ -554,6 +554,7 @@ public class DefaultTableRuntime extends AbstractTableRuntime
             currentChangeSnapshotId);
         state.setCurrentChangeSnapshotId(currentChangeSnapshotId);
         state.setCurrentSnapshotId(currentSnapshotId);
+        return true;
       }
     } else {
       long currentSnapshotId = doRefreshSnapshots((UnkeyedTable) table);
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurations.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurations.java
@@ -331,7 +331,17 @@ public class TableConfigurations {
             PropertyUtil.propertyAsLong(
                 properties,
                 TableProperties.SELF_OPTIMIZING_MIN_PLAN_INTERVAL,
-                TableProperties.SELF_OPTIMIZING_MIN_PLAN_INTERVAL_DEFAULT));
+                TableProperties.SELF_OPTIMIZING_MIN_PLAN_INTERVAL_DEFAULT))
+        .setEvaluationFallbackInterval(
+            PropertyUtil.propertyAsLong(
+                properties,
+                TableProperties.SELF_OPTIMIZING_EVALUATION_FALLBACK_INTERVAL,
+                TableProperties.SELF_OPTIMIZING_EVALUATION_FALLBACK_INTERVAL_DEFAULT))
+        .setEvaluationMseTolerance(
+            PropertyUtil.propertyAsLong(
+                properties,
+                TableProperties.SELF_OPTIMIZING_EVALUATION_FILE_SIZE_MSE_TOLERANCE,
+                TableProperties.SELF_OPTIMIZING_EVALUATION_FILE_SIZE_MSE_TOLERANCE_DEFAULT));
   }
 
   /**
--- a/amoro-ams/src/main/java/org/apache/amoro/server/utils/IcebergTableUtil.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/utils/IcebergTableUtil.java
@@ -228,15 +228,37 @@ public class IcebergTableUtil {
     OptimizingConfig config = tableRuntime.getOptimizingConfig();
     long lastMinor = tableRuntime.getLastMinorOptimizingTime();
     long lastFull = tableRuntime.getLastFullOptimizingTime();
+    long lastMajor = tableRuntime.getLastMajorOptimizingTime();
     if (TableFormat.ICEBERG.equals(table.format())) {
       return new IcebergOptimizerEvaluator(
-          identifier, config, table, snapshot, maxPendingPartitions, lastMinor, lastFull);
+          identifier,
+          config,
+          table,
+          snapshot,
+          maxPendingPartitions,
+          lastMinor,
+          lastFull,
+          lastMajor);
     } else if (TableFormat.MIXED_ICEBERG.equals(table.format())) {
       return new MixedIcebergOptimizingEvaluator(
-          identifier, config, table, snapshot, maxPendingPartitions, lastMinor, lastFull);
+          identifier,
+          config,
+          table,
+          snapshot,
+          maxPendingPartitions,
+          lastMinor,
+          lastFull,
+          lastMajor);
     } else if (TableFormat.MIXED_HIVE.equals(table.format())) {
       return new MixedHiveOptimizingEvaluator(
-          identifier, config, table, snapshot, maxPendingPartitions, lastMinor, lastFull);
+          identifier,
+          config,
+          table,
+          snapshot,
+          maxPendingPartitions,
+          lastMinor,
+          lastFull,
+          lastMajor);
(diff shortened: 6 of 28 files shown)


xxubai:
Can we move forward with this feature now? @Jzjsnow @klion26

Jzjsnow:
> Can we move forward with this feature now? @Jzjsnow @klion26

Sure, I've updated the branch and added the new evaluation criteria discussed earlier (see Step 1 for details). 

The current conditions for triggering pendingInput evaluation based on metrics are as follows:
Step 1: If the condition `delete file=0 && avg file size > target size * ratio` is met, the evaluation is considered unnecessary and will be skipped.
Step 2: Calculate detailed attributes for each partition in the table, including the sum of squared errors for file sizes. If this exceeds the file size tolerance threshold, the pendingInput requires evaluation.

Note that this update now supports MIX_ICEBERG tables, whereas previously only ICEBERG format was supported.

Please take a look when you are free. Looking forward to your feedback! @xxubai @zhoujinsong @klion26

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
```diff
@@ -133,6 +134,19 @@ private TableProperties() {}
       "self-optimizing.min-plan-interval";
   public static final long SELF_OPTIMIZING_MIN_PLAN_INTERVAL_DEFAULT = 60000;
 
+  /** metric-based evaluation related properties */
+  public static final String SELF_OPTIMIZING_EVALUATION_AVERAGE_FILE_SIZE_TOLERANCE =
+      "self-optimizing.evaluation.average-file-size.tolerance"; // the minimum tolerance value for
+  // the average
+  // partition file size (between 0 and (self-optimizing.target-size))
+  public static final MemorySize SELF_OPTIMIZING_EVALUATION_AVERAGE_FILE_SIZE_TOLERANCE_DEFAULT =
```
xxubai: Can we use **byte size** to unify the file size unit?

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/MixedAndIcebergTableDescriptor.java
```diff
@@ -451,19 +462,132 @@ public List<PartitionBaseInfo> getTablePartitions(AmoroTable<?> amoroTable) {
         getTableFilesInternal(amoroTable, null, null);
     try {
```
xxubai: Can use try catch with resource to close the io automaticly
```suggestion
        try (CloseableIterable<PartitionFileBaseInfo> tableFiles 
                 = getTableFilesInternal(amoroTable, null, null)) {
      for (PartitionFileBaseInfo fileInfo : tableFiles) {
        refreshPartitionBasicInfo(fileInfo, partitionBaseInfoHashMap);
      }
    } catch (IOException e) {
      LOG.warn("Failed to close the manifest reader.", e);
    }
```

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/MixedAndIcebergTableDescriptor.java
```diff
@@ -451,19 +462,132 @@ public List<PartitionBaseInfo> getTablePartitions(AmoroTable<?> amoroTable) {
...
+   * Get a list of partition information for a specific table that contains the square error sum of
+   * the partition file size. This method calculates statistics for each partition of a table,
+   * including file count, total file size, and the sum of squared errors of file sizes, for use in
+   * further operations such as table optimization.
+   */
+  public List<PartitionBaseInfo> getTablePartitionsWithDetailProperties(
+      MixedTable table, long minTargetSize) {
+    Map<String, PartitionBaseInfo> partitionBaseInfoHashMap = new HashMap<>();
+    CloseableIterable<PartitionFileBaseInfo> tableFiles = getTableFilesInternal(table, null, null);
+    try {
```
xxubai: You can simply use a try-with-resources statement.

amoro-ams/src/main/java/org/apache/amoro/server/refresh/event/MetricBasedRefreshEvent.java
```diff
@@ -0,0 +1,206 @@
...
+    // Perform periodic scheduling according to the fallback interval to avoid false positives or
+    // missed triggers based on metadata metric-driven evaluation
+    OptimizingConfig config = tableRuntime.getOptimizingConfig();
+    if (reachFallbackInterval(
+        tableRuntime.getLastPlanTime(), config.getEvaluationFallbackInterval())) {
+      logger.info("Maximum interval for evaluating table {} has reached.", table.id());
+      return true;
+    }
+
+    ExecutorService executorService = ThreadPools.getWorkerPool();
```
xxubai: Use a dedicated thread pool to avoid thread congestion.
```suggestion
       ExecutorService executorService = IcebergThreadPools.getPlanningExecutor();
```

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/MixedAndIcebergTableDescriptor.java
```diff
@@ -609,6 +733,11 @@ public List<OptimizingTaskInfo> getOptimizingTaskInfos(
   private CloseableIterable<PartitionFileBaseInfo> getTableFilesInternal(
       AmoroTable<?> amoroTable, String partition, Integer specId) {
     MixedTable mixedTable = getTable(amoroTable);
+    return getTableFilesInternal(mixedTable, partition, specId);
+  }
+
+  private CloseableIterable<PartitionFileBaseInfo> getTableFilesInternal(
```
xxubai: If I understand correctly, after the event-triggered evaluation, a full table scan is performed to collect partition information, which can be very expensive (especially for large tables with hundreds of thousands of files). Perhaps we can optimize this part when upgrading the Iceberg version and introducing **PartitionStatistics**.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TablePartitionDetailProperties.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.amoro.table;
+
+/** Detailed table partition properties list. */
+public class TablePartitionDetailProperties {
```
xxubai: Can we simply this name? such as `PartitionSummaryProperties`

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -115,7 +116,11 @@ public void execute(TableRuntime tableRuntime) {
                       != defaultTableRuntime.getCurrentChangeSnapshotId()))
           || (mixedTable.isUnkeyedTable()
               && lastOptimizedSnapshotId != defaultTableRuntime.getCurrentSnapshotId())) {
-        tryEvaluatingPendingInput(defaultTableRuntime, mixedTable);
+        if (!defaultTableRuntime.getOptimizingConfig().isEventBasedTriggerEnabled()
+            || MetricBasedRefreshEvent.isEvaluatingPendingInputNecessary(
```
xxubai: Will this cause an additional full table scan compared to before?
In addition, we should also check whether optimization is enabled, so it would be better to combine this with `tryEvaluatingPendingInput` to avoid extra overhead.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
Jzjsnow: Thanks for the suggestion. I have updated the variable to `self-optimising.evaluation.file-size.mse-tolerance` and specified bytes as the unit. This renders the variable name more clearly legible and aligns the unit with other size parameters.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/MixedAndIcebergTableDescriptor.java
Jzjsnow: Sure, I’ve optimized the logic in MixedAndIcebergTableDescriptor.getTablePartitions().

amoro-ams/src/main/java/org/apache/amoro/server/refresh/event/MetricBasedRefreshEvent.java
Jzjsnow: The latest commit removes the separate thread pool for manifest file MSE calculation since we now handle this during the evaluation scan phase instead.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/MixedAndIcebergTableDescriptor.java
Jzjsnow: You’re absolutely right. The latest commit moves partition file size MSE calculation to the evaluation scan phase, eliminating the need for an extra table scan. We’ll further optimize this with PartitionStatistics in future Iceberg upgrades to potentially avoid file-by-file traversal.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TablePartitionDetailProperties.java
Jzjsnow: This class has been removed in the latest commit—we now directly reuse similar variables from the evaluator instead.

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
Jzjsnow: Good call! In the latest commit, I've optimized the process to eliminate extra full table scans:
- Before `tryEvaluatingPendingInput`, we only judge based on the table's average file size and number of delete files.
- During `tryEvaluatingPendingInput`, MSE metrics are then calculated during the table scan.

We've also checked the metric-based trigger enabled in the `RuntimeRefeshExecutor` at the beginning,

Jzjsnow:
In the latest commit, we've revamped the logic of EventBasedTrigger with key adjustments and new configurations:

The EventBasedTrigger now includes two key parameters:
- FallbackInterval: The minimum interval for executing the original tryEvaluatingPendingInput logic. It prevents false positives or missed triggers from metadata metric-driven evaluation. Defaults to -1 (disabled); enabled when set to >=0.
- MseTolerance: The tolerance threshold for partition file size MSE (default: 0). Partitions with actual MSE below this threshold are considered unnecessary for optimization.

When enabled, the flow now:
- Determine if `tryEvaluatingPendingInput` needs to run:
  - Check if the FallbackInterval is met to trigger `tryEvaluatingPendingInput` directly.
  - Skip evaluation for empty tables.
  - Skip if the condition `delete file count = 0 && avg file size > target size * ratio` is satisfied (no need for pending input evaluation).
- Execute tryEvaluatingPendingInput if necessary:
  - Use the existing scan logic to retrieve partition file information.
  - Judge if each partition requires pending status: if the MSE threshold is met, further determine the optimization type (minor/major/full).
  - Update pendingInput related information.

Please take a look where you are free! @xxubai @klion26

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
```diff
@@ -133,6 +133,18 @@ private TableProperties() {}
...
+  /** metric-based evaluation related properties */
+  public static final String SELF_OPTIMIZING_EVALUATION_FALLBACK_INTERVAL =
+      "self-optimizing.evaluation.fallback-interval"; // fallback evaluation interval in
+  // milliseconds
+
+  public static final int SELF_OPTIMIZING_EVALUATION_FALLBACK_INTERVAL_DEFAULT =
+      -1; // event-based evaluation not in effect
+
+  public static final String SELF_OPTIMIZING_EVALUATION_FILE_SIZE_MSE_TOLERANCE =
+      "self-optimizing.evaluation.file-size.mse-tolerance";
```
turboFei: maybe add some comments for the MSE abbreviation

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -208,6 +217,11 @@ private boolean addTargetSizeReachedFile(DataFile dataFile, List<ContentFile<?>>
     return false;
   }
 
+  private void updateFileSizeSquaredErrorSum(DataFile dataFile) {
+    long diffSize = minTargetSize - Math.min(dataFile.fileSizeInBytes(), minTargetSize);
```
turboFei, line 227: maybe add comments that only accumulates squared error for files smaller than minTargetSize.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -208,6 +217,11 @@ private boolean addTargetSizeReachedFile(DataFile dataFile, List<ContentFile<?>>
...
+    fileSizeSquaredErrorSum += diffSize * diffSize;
```
turboFei: Is there a possibility of overflow here?

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+
+    void accept(Map<String, String> summary) {
+      deleteFileCnt +=
+          PropertyUtil.propertyAsInt(summary, SnapshotSummary.TOTAL_DELETE_FILES_PROP, 0);
+      dataFileCnt += PropertyUtil.propertyAsInt(summary, SnapshotSummary.TOTAL_DATA_FILES_PROP, 0);
+      fileSize +=
+          MemorySize.parse(
+                  PropertyUtil.propertyAsString(summary, SnapshotSummary.TOTAL_FILE_SIZE_PROP, "0"))
```
turboFei: PropertyUtil.propertyAsLong should be fine for TOTAL_FILE_SIZE_PROP

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -62,6 +63,16 @@ private void tryEvaluatingPendingInput(DefaultTableRuntime tableRuntime, MixedTa
     // only evaluate pending input when optimizing is enabled and in idle state
     if (tableRuntime.getTableConfiguration().getOptimizingConfig().isEnabled()
         && tableRuntime.getOptimizingStatus().equals(OptimizingStatus.IDLE)) {
+
+      if (tableRuntime.getTableConfiguration().getOptimizingConfig().isEventBasedTriggerEnabled()
+          && !MetricBasedEvaluationEvent.isEvaluatingNecessary(
+              tableRuntime.getOptimizingConfig(), table, tableRuntime.getLastPlanTime())) {
```
turboFei: what is the difference between 

`tableRuntime.getTableConfiguration().getOptimizingConfig()` and `tableRuntime.getOptimizingConfig()`, it looks confuse.

turboFei: Can we define variable likes `OptimizingConfig config = tableRuntime.getOptimizingConfig()` and reuse it to make it clear?

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
```diff
@@ -133,6 +133,18 @@ private TableProperties() {}
...
+  public static final long SELF_OPTIMIZING_EVALUATION_FILE_SIZE_MSE_TOLERANCE_DEFAULT = 0;
```
turboFei, line 147: I saw that, now 

if `SELF_OPTIMIZING_EVALUATION_FALLBACK_INTERVAL_DEFAULT > 0`, `isEventBasedTriggerEnabled` is true.

Should event-based evaluation be considered "enabled" if fallbackInterval ≥ 0 but evaluationMseTolerance == 0.

If not, can we give a default value for `mse-tolerance` or give some suggestions to determine the mse tolerance?

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,123 @@
...
+    return true;
+  }
+
+  public static boolean isPartitionPendingNecessary(
+      OptimizingConfig config, long partitionSquaredErrorSum, long partitionFileCount) {
+    long mseTolerance = config.getEvaluationMseTolerance();
+    return partitionFileCount > 1
+        && (mseTolerance == 0
+            || (double) partitionSquaredErrorSum / partitionFileCount
+                >= mseTolerance * mseTolerance);
```
turboFei: `>= (double) mseTolerance * mseTolerance`

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,123 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+public class MetricBasedEvaluationEvent {
+  private static final Logger logger = LoggerFactory.getLogger(MetricBasedEvaluationEvent.class);
+
+  public static boolean isReachFallbackInterval(OptimizingConfig config, long lastPlanTime) {
+    long fallbackInterval = config.getEvaluationFallbackInterval();
+    return fallbackInterval >= 0 && System.currentTimeMillis() - lastPlanTime > fallbackInterval;
```
turboFei: nit:
```
>= fallbackInterval
```

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  public static boolean isEvaluatingNecessary(
+      OptimizingConfig config, MixedTable table, long lastPlanTime) {
+    if (table.format() != TableFormat.ICEBERG && table.format() != TableFormat.MIXED_ICEBERG) {
+      logger.debug(
+          "MetricBasedRefreshEvent only support ICEBERG/MIXED_ICEBERG tables. Always return true for other table formats.");
```
turboFei: is it by design?

`MetricBasedRefreshEvent` or `MetricBasedEvaluationEvent`

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
Jzjsnow: Sure, I've added in the latest commit.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
Jzjsnow: Great point! I've updated the comments clarifying that the squared error accumulation is only applied to files with size smaller than `minTargetSize`.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
Jzjsnow: Thanks for catching this potential issue! I've fixed it by using Long.MAX_VALUE as the upper bound to prevent numeric overflow in the relevant logic.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
Jzjsnow: Thanks for pointing this out! This is indeed a typo in the log message and is fixed to align with the class name.

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
Jzjsnow: > what is the difference between
> 
> `tableRuntime.getTableConfiguration().getOptimizingConfig()` and `tableRuntime.getOptimizingConfig()`, it looks confuse.

This is indeed puzzling. I see that the instances obtained by both methods are consistent. We may consider optimizing this in the future.

Anyway, an `OptimizingConfig` variable is defined and reused to improve code clarity.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
Jzjsnow: Thanks for your thoughtful question! When `fallbackInterval ≥ 0 and evaluationMseTolerance == 0`, the MSE metric is not actually used for evaluation, but the pre-planning judgment logic is still activated. 

This logic includes two key steps which is in `MetricBasedEvaluationEvent#isEvaluatingNecessary()`:
a. Skip empty tables;
b. Skip the pending input evaluation process if the conditions `delete file=0 && avg file size > target size * ratio` are both met.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,119 @@
...
+    if (table.format() != TableFormat.ICEBERG && table.format() != TableFormat.MIXED_ICEBERG) {
+      logger.debug(
+          "MetricBasedEvaluationEvent only support ICEBERG/MIXED_ICEBERG tables. Always return true for other table formats.");
+      return true;
+    }
+    // Step 1: Perform periodic scheduling according to the fallback interval to avoid false
+    // positives or
+    // missed triggers based on metadata metric-driven evaluation
+    if (isReachFallbackInterval(config, lastPlanTime)) {
+      logger.info("Maximum interval for evaluating table {} has reached.", table.id());
```
xxubai: Set the log level to DEBUG. The evaluation process will be invoked frequently, generating a significant amount of log information

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,119 @@
...
+      return true;
+    }
+
+    // Step 2: Empty table should skip evaluating pending input
+    BasicTableStats basicStats = new BasicTableStats(table);
+    if (basicStats.dataFileCnt == 0) {
+      logger.info("Table {} contains no data files, skip evaluating pending input.", table.id());
```
xxubai: Same above

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    // Step 3: If the condition `delete file=0 && avg file size > target size * ratio` is satisfied,
+    // then evaluating the pending input is considered unnecessary and will be skipped.
+    long minTargetSize = (long) (config.getTargetSize() * config.getMinTargetSizeRatio());
+    double avgFileSize =
+        basicStats.dataFileCnt > 0 ? (double) basicStats.fileSize / basicStats.dataFileCnt : 0;
+
+    if (basicStats.deleteFileCnt == 0 && avgFileSize > minTargetSize) {
+      logger.info(
```
xxubai: Samve above

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,119 @@
...
+    return partitionFileCount > 1
+        && (mseTolerance == 0
+            || (double) partitionSquaredErrorSum / partitionFileCount
+                >= (double) mseTolerance * mseTolerance);
+  }
+
+  static class BasicTableStats {
+    int deleteFileCnt = 0;
+    int dataFileCnt = 0;
+    long fileSize = 0;
```
xxubai: nit: totalFileSize

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @VisibleForTesting
+    BasicTableStats() {}
+
+    BasicTableStats(MixedTable table) {
```
xxubai: nit: How about rename to `BasicFileStats`

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
```diff
@@ -288,6 +322,17 @@ private void addDelete(ContentFile<?> delete) {
   @Override
   public boolean isNecessary() {
     if (necessary == null) {
+      long lastPlanTime = Math.max(lastMinorOptimizingTime, lastFullOptimizingTime);
```
xxubai: Lack last major optimizing time?

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
```diff
@@ -0,0 +1,119 @@
...
+import org.apache.amoro.table.MixedTable;
+import org.apache.iceberg.Snapshot;
+import org.apache.iceberg.SnapshotSummary;
+import org.apache.iceberg.util.PropertyUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+public class MetricBasedEvaluationEvent {
```
xxubai: Why is it named MetricBasedEvaluationEvent when it doesn’t receive any metric events? Additionally, it is only bound to the Iceberg/Mixed Iceberg format, making the coupling too strong.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
Jzjsnow: Sure, all the log levels have been modified

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/CommonPartitionEvaluator.java
Jzjsnow: In the latest commit, the `lastMajorOptimizingTime` has been added to the partition evaluator alongside `lastMinorOptimizingTime` and `lastFullOptimizingTime`, enabling more comprehensive evaluations

xxubai:
LGTM. Also need to fix the unit tests

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/evaluation/MetricBasedEvaluationEvent.java
Jzjsnow: This naming might be misleading, so in the latest commit, I renamed it to MetadataBasedEvaluationEvent as we use the table's file statistics to trigger evaluation.

The reason for supporting only Iceberg/Mixed formats is that table file statistics for these formats can be directly retrieved from the table summary without planning, whereas other formats cannot. However, I've abstracted a TableStatsProvider interface, which may enable support for other formats in the future. PTAL

xxubai:
cc @klion26 @zhoujinsong @baiyangtx @Aireed @zhongqishang do you have any comments?

zhoujinsong:
Thanks for the great work! @Jzjsnow 
Thanks for the review! @xxubai @klion26 @turboFei

PR #3797, opened 2025-09-24
[AMORO-3796] Fixed UUID type in Iceberg tables with partitions & buckets

### LINKED ISSUE #3796: [Bug]: UUID Issue from native iceberg library
### What happened?

There is an error where Iceberg table that is bucketed (or partitioned) on the key with type UUID

### Affects Versions

0.8.0

Iceberg

Optimizer

### How to reproduce

1) Create an iceberg table that Amoro can access for optimisation purposes

2) Create an optimiser and see the logs of the Amoro executor pod, which starts to optimise the table

### Relevant log output

```shell

```

### Anything else

java.lang.IllegalStateException: Not an instance of java.util.UUID: [B@247ddda
	at org.apache.iceberg.data.GenericRecord.get(GenericRecord.java:123)
	at org.apache.iceberg.data.InternalRecordWrapper.get(InternalRecordWrapper.java:101)
	at org.apache.iceberg.Accessors$PositionAccessor.get(Accessors.java:71)
	at org.apache.iceberg.Accessors$PositionAccessor.get(Accessors.java:58)
	at org.apache.iceberg.StructTransform.wrap(StructTransform.java:78)
	at org.apache.iceberg.PartitionKey.wrap(PartitionKey.java:30)
	at org.apache.iceberg.PartitionKey.partition(PartitionKey.java:64)
	at org.apache.amoro.io.writer.GenericIcebergPartitionedFanoutWriter.partition(GenericIcebergPartitionedFanoutWriter.java:65)
	at org.apache.amoro.io.writer.GenericIcebergPartitionedFanoutWriter.partition(GenericIcebergPartitionedFanoutWriter.java:42)
	at org.apache.iceberg.io.PartitionedFanoutWriter.write(PartitionedFanoutWriter.java:52)
	at org.apache.amoro.optimizing.AbstractRewriteFilesExecutor.rewriterDataFiles(AbstractRewriteFilesExecutor.java:152)
	at org.apache.amoro.table.TableMetaStore.call(TableMetaStore.java:268)
	at org.apache.amoro.table.TableMetaStore.lambda$doAs$0(TableMetaStore.java:243)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:360)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1930)
	at org.apache.amoro.table.TableMetaStore.doAs(TableMetaStore.java:243)
	at org.apache.amoro.io.AuthenticatedHadoopFileIO.doAs(AuthenticatedHadoopFileIO.java:201)
	at org.apache.amoro.optimizing.AbstractRewriteFilesExecutor.execute(AbstractRewriteFilesExecutor.java:108)
	at org.apache.amoro.optimizing.AbstractRewriteFilesExecutor.execute(AbstractRewriteFilesExecutor.java:64)
	at org.apache.amoro.optimizer.common.OptimizerExecutor.executeTask(OptimizerExecutor.java:150)
	at org.apache.amoro.optimizer.common.OptimizerExecutor.executeTask(OptimizerExecutor.java:104)
	at org.apache.amoro.optimizer.common.OptimizerExecutor.start(OptimizerExecutor.java:53)
	at java.lang.Thread.run(Thread.java:750)

### Code of Conduct

CatOrLeader:
implement automatic replacement of UUID type columns with the FixedType[16]

There is an issue with Amoro's attempt to optimise tables in Iceberg when using partitions or buckets based on a UUID key. These changes FIX #3796 and do not corrupt any data

## Brief change log

- Added conversion of UUID type in the Iceberg table to the FixedType[16]
- Conversion is internal only so there is no side effect on the data in the table

## How was this patch tested?

- The patch was tested manually in the production environment of the company I currently work in – the new. jar file of amoro-format-iceberg (with the changes) replace the original one

- Optimisers work correctly, and table files are optimised successfully.

## Documentation

- If yes, how is the feature documented? Not applicable


xxubai (commented):
Thanks for raising this issue.

The UUID records in the data file are persisted as binary. If I understand correctly, the reader cannot directly cast binary to UUID.

You converted the UUID type to a fixed-length 16-byte binary in the schema, which makes sense. But I wonder if there is a more elegant way to fix this, ideally without intrusive changes to the writer code.

CatOrLeader:
@xxubai 

Hello! Thanks for your answer; I appreciate it a lot! About the issues you raised:

I was trying to find any other part of the code that can be fixed to resolve this issue; however, it is directly connected to the [original Iceberg library](https://github.com/apache/iceberg), for example:

* https://github.com/apache/iceberg/pull/14027

It is not fixed yet, but in my opinion, it is necessary to give users the ability to have UUID as a partition key even now because in some way it can be more optimised than other types.

Additionally, we can see a similar issue with the UUID in the libraries that use Iceberg:

* PyIceberg: https://github.com/apache/iceberg/issues/8248

I am open to any other possibility to fix this issue in the Amoro code; however, for now it seems like this dirty way of casting classes is the single possibility to resolve the issue on the Amoro code side

zhoujinsong:
Hi, maybe we should try to convert the byte array to UUID in https://github.com/apache/iceberg/blob/main/data/src/main/java/org/apache/iceberg/data/InternalRecordWrapper.java#L50.

CatOrLeader:
@zhoujinsong Hello! Yep, it is necessary to fix it in the `iceberg` library; however, it seems like we need to wait a lot of time while the bug is fixed. Additionally, when the `iceberg` library is updated, we need to bump the version.

However, it will be great if there exists a process that will keep my changes at least while a new Iceberg version with corrected code will be included in the `amoro`

xxubai:
> @zhoujinsong Hello! Yep, it is necessary to fix it in the `iceberg` library; however, it seems like we need to wait a lot of time while the bug is fixed. Additionally, when the `iceberg` library is updated, we need to bump the version.

Hi @CatOrLeader . Thanks for the contribution!

I directly tried modifying InternalRecordWrapper to add the corresponding converter in order to fix this issue (https://github.com/apache/iceberg/pull/14208/files#diff-3be4a9b849bc3dad71d9224f6d8f1b70ac5d6e36db9a1e608273c5df3e127a7c). 

I think we could follow @zhoujinsong 's suggestion and create a copy class inside Amoro to replace the current InternalRecordWrapper(Such as `src/main/java/org/apache/iceberg/data/IcebergRecordWrapper.java`).
Later, once this issue is fixed upstream, we could revert to the native class. What do you think?

CatOrLeader:
@xxubai Hi! I agree with you and @zhoujinsong that this method is more elegant and should be applied. Thank you for the expertise!

I will reflect the conclusions in the code and tag you again when it will be ready and tested in out environment. Is it ok?

Also, thank you for contributing to the `iceberg` library to fix this issue; I appreciate it

xxubai (approved):
LGTM

zhoujinsong:
Thanks for the work! @CatOrLeader 
Thanks for the review! @xxubai

PR #3802, opened 2025-09-26
[AMORO-3781] Save the last completion time for each cleanup operation performed on each optimization table.

### LINKED ISSUE #3781: [Feature]: Save the last completion time of the executor cleanup
### Description

The Amoro cleanup executor triggers its first execution when AMS starts, and subsequently runs periodically according to its respective interval.
The following scenario exists:
      When the Amoro cleanup executor has just finished running a few minutes ago, the AMS service is manually restarted, causing the cleanup executor to run again.

### Use case/motivation
Clearing executors should not be triggered frequently before and after ams service restart.

### Describe the solution

After the AMS service restarts, whether the executor for optimizing tables cleans up immediately depends on whether the interval since the last completion time exceeds a fixed execution interval.
 

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

zhangwl9:

Close #3781.

## Brief change log

To optimize table cleanup operations (Dangling-Delete_file cleanup, data expiration, orphan_file cleanup,snapshotsFile cleanup), the last cleanup time is saved. This ensures that after an AMS service restart, if a cleanup operation was completed just before the restart, the next cleanup operation will only be performed after exceeding the respective execution interval.
The execution flow is as follows:

- During the initialization of the ams service, retrieve the last execution time for each cleanup operation of every table to be optimized from the table_runtime_state table.

- Check whether the time interval since the last cleanup exceeds the cleanup operation execution interval.

- For optimization tables exceeding the execution interval, execute the cleanup operation.

- Otherwise, schedule for the next round of execution.

- Since each table's cleanup operation takes varying durations, after a table's cleanup completes, update the cleanup completion time in the table_runtime_state table.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/PeriodicTableScheduler.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/scheduler/PeriodicTableScheduler.java
@@ -127,8 +127,7 @@ public abstract class PeriodicTableScheduler extends RuntimeHandlerChain {
       if (isExecutable(tableRuntime)) {
         execute(tableRuntime);
         // Different tables take different amounts of time to execute the end of execute(),
-        // so you need to perform the save or update lastCleanEndTime operation separately for each
-        // table.
+        // so you need to perform the update operation separately for each table.
         persistUpdatingCleanupTime(tableRuntime);
       }
     } finally {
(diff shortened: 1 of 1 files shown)


zhangwl9:
I have completed the logic modification to save cleanup times to the table_runtime_state table. Could you please review the code at your convenience? Thank you very much. @czy006 @baiyangtx

zhoujinsong:

PR #3841, opened 2025-10-23
[AMORO-3608] Enhance RollingFileCleaner to preserve full URI for remote filesystems

xxubai:

Fix #3630

## Brief change log

- as title

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java
@@ -59,24 +59,7 @@ public class RollingFileCleaner {
     }
 
     collectedFiles.add(filePath);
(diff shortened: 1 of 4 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java
```diff
@@ -59,7 +59,24 @@ public void addFile(String filePath) {
     }
 
     collectedFiles.add(filePath);
-    String parentDir = new Path(URI.create(filePath).getPath()).getParent().toString();
+    // Preserve the full URI including scheme (e.g., s3://, s3a://, hdfs://) for S3 and other remote
```
zhoujinsong: Thanks a lot for contributing!

Do you think we should add a util method for it in `TableFileUtil`?

zhoujinsong (approved):
LGTM.

PR #3849, opened 2025-10-31
[AMORO-3852][BUILD] Enhance build system

turboFei:

build/mvn is introduced for auto-downloading Apache Maven if no suitable candidate mvn in the PATH, the suitable mvn means the version of mvn must be equal or great than maven.version defined in pom.xml

Same solution likes spark, kyuubi, celeborn

## Brief change log

- Introduce build/mvn script

## How was this patch tested?

## Documentation


pan3793:
[Maven Wrapper](https://maven.apache.org/tools/wrapper/) might be a better alternative as of today, especially for developers who use Microsoft Windows

PR #3856, opened 2025-10-31
[Bug]: Amoro optimization can result in the input files and the merge…

### LINKED ISSUE #3855: [Bug]: Amoro optimization can result in the input files and the merged output files having the same number of files, and this can cause the merge to fail and keep triggering the merge task.
### What happened?

Amoro optimization can result in the input files and the merged output files having the same number of files, and this can cause the merge to fail and keep triggering the merge task.
As shown below
Image

Image

### Affects Versions

0.7.1

Iceberg

Spark

### How to reproduce

_No response_

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

wardlican:
Amoro optimization can result in the input files and the merged output files having the same number of files, and this can cause the merge to fail and keep triggering the merge task.

Close #3855 

## Brief change log

1. For undersizedSegmentFiles that result in only one file after bin-packing:
2. If rewriting pos delete is not required, simply add them to rewriteDataFiles.
3. During the second bin-packing, they can be merged with fragment files or other files.

-

## How was this patch tested?

## Documentation


wardlican:
Please help review whether this repair plan is feasible.

Fix Results:

- When the total size of the input files is less than `targetSize`, return `Long.MAX_VALUE` (existing logic).
- When the total size of the input files is greater than or equal to `targetSize` but the average file size is less than `targetSize`, also return `Long.MAX_VALUE` (new logic).
- Only return `targetSize` when the average file size is greater than or equal to `targetSize`.

This ensures that:

multi small files (even if the total size is greater than or equal to 128MB, but the average file size is less than 128MB) can be merged.

- Avoiding the generation of output files with the same number of input files.
- Solving the file merging problem at the task execution level.
- The fix is ​​complete; small files should now be merged correctly.

amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/IcebergRewriteExecutor.java
```diff
@@ -101,12 +101,25 @@ protected TaskWriter<Record> dataWriter() {
...
+    }
+    long inputSize = Arrays.stream(rewrittenDataFiles).mapToLong(DataFile::fileSizeInBytes).sum();
+    // When the input files' total size is below targetSize, remove the output file size limit to
     // avoid outputting multiple files.
     // For more details, please refer to: https://github.com/apache/amoro/issues/3645
-    return inputSize < targetSize ? Long.MAX_VALUE : targetSize;
+    if (inputSize < targetSize) {
+      return Long.MAX_VALUE;
+    }
+    // Even if total size >= targetSize, if average file size is small (less than targetSize),
```
klion26: Will this output a single file with too big file size?

xxubai (commented):
@wardlican 
I have the same concern: could simply setting it to the maximum value result in a single file being excessively large? Could we perhaps refer to Spark's estimation logic？
Refer to  https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/actions/SizeBasedFileRewritePlanner.java#L199-L206

wardlican:
> @wardlican I have the same concern: could simply setting it to the maximum value result in a single file being excessively large? Could we perhaps refer to Spark's estimation logic？ Refer to https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/actions/SizeBasedFileRewritePlanner.java#L199-L206

Okay, I'll look into whether we can use the design from Iceberg for this.

wardlican:
Referencing Spark/Iceberg's SizeBasedFileRewritePlanner, the number of output files is determined intelligently:
| Condition      | Decision |
| ----------- | ----------- |
| Remainder > minFileSize (default 0.75 * targetSize)      |  Retain the remainder file (round up)       |
| avgFileSizeWithoutRemainder < 1.1 * targetSize    | Distribute the remainder to other files (rounding down)        |
| other   | Retain the remainder file (round up)        |

wardlican:
> o whether we can use the design



> @wardlican I have the same concern: could simply setting it to the maximum value result in a single file being excessively large? Could we perhaps refer to Spark's estimation logic？ Refer to https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/actions/SizeBasedFileRewritePlanner.java#L199-L206

The current implementation is based on the implementation of SizeBasedFileRewritePlanner; please help review it. @xxubai

xxubai:
> The current implementation is based on the implementation of SizeBasedFileRewritePlanner; please help review it. @xxubai

Can you test this implementation in your enviroment? It is really helpful for validating. Otherwise we should add a unit test to cover this case

wardlican:
> > The current implementation is based on the implementation of SizeBasedFileRewritePlanner; please help review it. @xxubai
> 
> Can you test this implementation in your enviroment? It is really helpful for validating. Otherwise we should add a unit test to cover this case

Okay, I will verify it in my local environment and add some unit tests.

wardlican:
> > The current implementation is based on the implementation of SizeBasedFileRewritePlanner; please help review it. @xxubai
> 
> Can you test this implementation in your enviroment? It is really helpful for validating. Otherwise we should add a unit test to cover this case

Unit tests have been added for the targetSize() method, covering the following scenarios:

- **testTargetSizeWithSmallInputFiles** - Merging multiple small files
        **Scenario**: 5 files of 10MB each (Total 50MB < 128MB).
        **Expected**: Returns inputSize (50MB), merging them into a single file.

- **testTargetSizeWithSingleSmallFile** - Single small file
        **Scenario**: 1 file of 10MB.
        **Expected**: Returns targetSize (128MB).

- **testTargetSizeWithExactTargetSize** - Input size equals targetSize
        **Scenario**: Input size is exactly 128MB.
        **Expected**: Returns ≥targetSize.

- **testTargetSizeWithLargeRemainder** - Large remainder case
        **Scenario**: 225MB input, remainder is 97MB (which is > 96MB minFileSize).
        **Expected**: Rounds up to calculate an appropriate split size.

- **testTargetSizeWithSmallRemainderDistributed** - Distributable small remainder
        **Scenario**: 256MB input (exactly 2x targetSize, no remainder).
        **Expected**: Rounds down and distributes the size evenly.

- **testTargetSizeWithMultipleSmallFiles** - Multiple small files exceeding targetSize
        **Scenario**: 10 files of 20MB each (Total 200MB > 128MB).
        **Expected**: Calculates an appropriate split size.

- **testTargetSizeWithVeryLargeInput** - Very large input
        **Scenario**: 500MB input.
        **Expected**: Capped at maxFileSize (192MB, calculated as targetSize×1.5).

- **testTargetSizeWithCustomMinTargetSizeRatio** - Custom ratio
        **Scenario**: Custom min-target-size-ratio set to 0.8.
        **Expected**: Performs calculations based on the custom ratio.

- **testTargetSizeBoundaryConditions** - Boundary conditions
        **Scenario**: Input size is near boundary values.
        **Expected**: Correctly handles edge cases and boundary conditions.

PR #3927, opened 2025-11-12
[Subtask]: Modify DefaultTableService to be compatible with master-slave mode.

### LINKED ISSUE #3923: [Subtask]: Modify DefaultTableService to be compatible with master-slave mode
### Search before asking

### Description

In DefaultTable Service, it is compatible with master-slave mode. In master-slave mode, it is necessary to load tables from catalog, determine the addition and deletion of tables, complete the allocation of tables to bucketId, and the logic of ams nodes loading the tables they are responsible for based on bucketId.

### Parent issue

### Code of Conduct

wardlican:
In DefaultTable Service, it is compatible with master-slave mode. In master-slave mode, it is necessary to load tables from catalog, determine the addition and deletion of tables, complete the allocation of tables to bucketId, and the logic of ams nodes loading the tables they are responsible for based on bucketId.

Close #3923 .

## Brief change log

- Add a bucketTableSyncScheduler to DefaultTableService for table loading logic in master-slave mode.
- The history of adding upgrades and switching to master-slave mode is that the allocation table is initialized with SQL for bucket allocation.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
@@ -253,13 +253,7 @@ public class AmoroServiceContainer {
 
     BucketAssignStore bucketAssignStore = null;
     if (IS_MASTER_SLAVE_MODE && haContainer != null) {
-      try {
-        bucketAssignStore = BucketAssignStoreFactory.create(haContainer, serviceConfig);
-      } catch (Exception e) {
-        LOG.error(
-            "Failed to create BucketAssignStore; master-slave bucket assignment and per-node table loading are disabled",
-            e);
-      }
+      bucketAssignStore = BucketAssignStoreFactory.create(haContainer, serviceConfig);
     }
 
     // In master-slave mode, create AmsAssignService for bucket assignment (shares BucketAssignStore
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
@@ -536,8 +536,11 @@ public class DefaultTableService extends PersistentBase implements TableService 
 
   @VisibleForTesting
   void exploreTableRuntimes() {
-    // In master-slave mode, only leader node should explore table runtimes
-    if (isMasterSlaveMode && haContainer != null && !haContainer.hasLeadership()) {
+    // In master-slave mode (fully wired), only leader node should explore table runtimes
+    if (isMasterSlaveMode
+        && haContainer != null
+        && bucketAssignStore != null
+        && !haContainer.hasLeadership()) {
       LOG.debug("Not the leader node in master-slave mode, skip exploring table runtimes");
       return;
     }
@@ -794,12 +797,12 @@ public class DefaultTableService extends PersistentBase implements TableService 
     meta.setGroupName(configuration.getOptimizingConfig().getOptimizerGroup());
     meta.setTableSummary(new TableSummary());
 
-    // In master-slave mode, assign bucketId to the table if it's not assigned yet.
+    // In master-slave mode (fully wired), assign bucketId to the table if it's not assigned yet.
     // Only leader node should assign bucketIds; follower may still persist the table with null
     // bucketId (e.g. onTableCreated on follower), and leader will assign later via exploration.
     String assignedBucketId = null;
-    if (isMasterSlaveMode) {
-      if (haContainer != null && haContainer.hasLeadership()) {
+    if (isMasterSlaveMode && haContainer != null && bucketAssignStore != null) {
+      if (haContainer.hasLeadership()) {
         TableRuntimeMeta existingMeta =
             getAs(
                 TableRuntimeMapper.class,
@@ -847,7 +850,10 @@ public class DefaultTableService extends PersistentBase implements TableService 
 
     doAs(TableRuntimeMapper.class, mapper -> mapper.insertRuntime(meta));
 
-    if (isMasterSlaveMode) {
+    // Only skip local runtime creation when master-slave mode is fully wired (bucketAssignStore
+    // is non-null). When bucketAssignStore is null (e.g. 3-arg test constructor), fall through
+    // and create the runtime in memory as in non-master-slave mode.
+    if (isMasterSlaveMode && haContainer != null && bucketAssignStore != null) {
       return true;
     }
 


wardlican:
Please perform CR on these implementations.

amoro-common/src/main/java/org/apache/amoro/ServerTableIdentifier.java
```diff
@@ -30,6 +30,7 @@ public class ServerTableIdentifier {
   private String database;
   private String tableName;
   private TableFormat format;
+  private String bucketId;
```
czy006: need to get and set methods?

amoro-ams/src/main/resources/mysql/upgrade.sql
```diff
@@ -161,3 +161,15 @@ ALTER TABLE `table_process`
 ADD COLUMN `external_process_identifier` varchar(256) DEFAULT NULL COMMENT 'Table optimizing external process identifier',
 ADD COLUMN `retry_number` int(11) NOT NULL DEFAULT 0 COMMENT 'Retry times',
 ADD COLUMN `process_parameters` mediumtext COMMENT 'Table process parameters';
+
+-- ADD table bucket_assignments for storing assigned info
+CREATE TABLE IF NOT EXISTS bucket_assignments (
```
czy006, line 166: #4121 is merge this sql

amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
```diff
@@ -62,6 +62,34 @@ public class AmoroManagementConf {
               "This setting controls whether to enable the AMS horizontal scaling feature, "
                   + "which is currently under development and testing.");
 
+  public static final ConfigOption<Integer> BUCKET_ID_TOTAL_COUNT =
```
czy006: this is have in other pr

amoro-ams/src/main/resources/mysql/upgrade.sql
wardlican: PR #4121 only added the `init` sql; in this PR, I'm taking the opportunity to address that omission as well.

amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
wardlican: Yes, I will remove this duplicate configuration item.

amoro-common/src/main/java/org/apache/amoro/ServerTableIdentifier.java
wardlican: I will remove this attribute; it is not needed here for the time being.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
```diff
@@ -495,7 +821,78 @@ private boolean triggerTableAdded(
...
+                mapper -> mapper.selectRuntime(serverTableIdentifier.getId()));
+        if (existingMeta != null) {
+          // Runtime already exists (e.g. inserted by follower with null bucketId)
+          if (existingMeta.getBucketId() != null) {
+            return true; // already assigned
+          }
+          assignedBucketId = assignBucketIdForTable();
+          if (assignedBucketId != null) {
+            existingMeta.setBucketId(assignedBucketId);
+            doAs(TableRuntimeMapper.class, mapper -> mapper.updateRuntime(existingMeta));
```
xxubai: Backfilling bucket_id for an existing runtime uses updateRuntime(existingMeta), which updates the full row, not just bucket_id. 
That can overwrite concurrent changes to status/config/summary with stale values read earlier. 
This should be a dedicated `UPDATE ... SET bucket_id = ? WHERE table_id = ? AND bucket_id IS NULL` instead.

amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
```diff
@@ -62,6 +62,13 @@ public class AmoroManagementConf {
               "This setting controls whether to enable the AMS horizontal scaling feature, "
                   + "which is currently under development and testing.");
 
+  public static final ConfigOption<Duration> BUCKET_TABLE_SYNC_INTERVAL =
+      ConfigOptions.key("bucket-table-sync.interval")
```
xxubai: The new config naming is a bit misleading. `bucket-table-sync.interval` is HA/master-slave-specific, but it is exposed as a top-level AMS config. `ha.bucket-table-sync.interval` would better reflect its scope and align with the rest of the HA settings.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
```diff
@@ -87,19 +94,46 @@ public class DefaultTableService extends PersistentBase implements TableService
...
+  private final boolean isMasterSlaveMode;
   private RuntimeHandlerChain headHandler;
   private ExecutorService tableExplorerExecutors;
 
+  // Master-slave mode related fields
+  private ScheduledExecutorService bucketTableSyncScheduler;
+  private volatile List<String> assignedBucketIds = new ArrayList<>();
+  private final long bucketTableSyncInterval;
+  // Lock for bucketId assignment to prevent concurrent assignment conflicts
+  private final Object bucketIdAssignmentLock = new Object();
```
xxubai: The current bucketIdAssignmentLock only serializes assignment inside a single JVM, so it cannot guarantee correctness across AMS nodes, while still making the whole DB-read + heap-build path single-threaded locally. 

I’d suggest narrowing the lock to local pending-counter bookkeeping only, and moving correctness to persistence-layer atomicity instead, e.g. a dedicated `UPDATE table_runtime SET bucket_id = ? WHERE table_id = ? AND bucket_id IS NULL` path. 

Longer term, bucket selection itself should be coordinated by the shared HA/store layer rather than local in-memory state.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
```diff
@@ -259,8 +341,234 @@ public void dispose() {
...
+        // Use min-heap to find bucketId with minimum table count
+        // Create new entries instead of using Map.Entry directly to avoid reference issues
+        PriorityQueue<Map.Entry<String, Integer>> minHeap =
+            new PriorityQueue<>(
+                Comparator.<Map.Entry<String, Integer>>comparingInt(Map.Entry::getValue)
+                    .thenComparing(Map.Entry::getKey)); // Use bucketId as tie-breaker
+
+        // Create independent entries for the heap to avoid reference issues
+        for (Map.Entry<String, Integer> entry : bucketTableCount.entrySet()) {
+          minHeap.offer(new java.util.AbstractMap.SimpleEntry<>(entry.getKey(), entry.getValue()));
```
xxubai: nit:

```suggestion
          minHeap.offer(new SimpleEntry<>(entry.getKey(), entry.getValue()));
```

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
```diff
@@ -87,19 +94,46 @@ public class DefaultTableService extends PersistentBase implements TableService
...
+  private final Object bucketIdAssignmentLock = new Object();
+  // Local cache of bucketId assignments to track pending assignments before they're saved to DB
+  // This ensures concurrent table creation gets different bucketIds even before DB is updated
+  private final Map<String, Integer> pendingBucketIdCounts = new ConcurrentHashMap<>();
+
   public DefaultTableService(
       Configurations configuration,
       CatalogManager catalogManager,
       TableRuntimeFactory tableRuntimeFactory) {
+    this(configuration, catalogManager, tableRuntimeFactory, null, null);
```
xxubai, line 114: Master-slave mode is not actually wired into `DefaultTableService`. `AmoroServiceContainer` still constructs the service with the old 3-arg constructor, so haContainer and bucketAssignStore stay null. 

As a result, `use-master-slave-mode=true` enables part of the new logic, but table loading/sync still falls back to the old path while `triggerTableAdded()` stops creating in-memory runtimes. 

This can leave newly created tables persisted but missing from tableRuntimeMap, and later scans may repeatedly try to insert them again.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
wardlican: Okay, I will add an `updateBucketIdIfNull` method, specifically for updating the `bucketId`.

amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
wardlican: Okay, it does indeed need to be placed under `ha` here.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
wardlican: Yes, there really is no need to add this lock here.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
wardlican: Okay, I will fix this issue.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
wardlican: Okay, I will optimize the implementation here to prevent this issue.

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
```diff
@@ -250,12 +250,23 @@ public void startOptimizingService() throws Exception {
...
     defaultRuntimeFactory.initialize(processFactories);
-    // In master-slave mode, create AmsAssignService for bucket assignment
+
+    BucketAssignStore bucketAssignStore = null;
     if (IS_MASTER_SLAVE_MODE && haContainer != null) {
       try {
-        // Create and start AmsAssignService for bucket assignment
-        // The factory will handle different HA types (ZK, database, etc.)
-        amsAssignService = new AmsAssignService(haContainer, serviceConfig);
+        bucketAssignStore = BucketAssignStoreFactory.create(haContainer, serviceConfig);
```
xxubai: If I understand correctly, when BucketAssignStore initialization fails, the service does not fully fall back to non-master-slave behavior. `initialize()` falls back to loading all table runtimes because bucketAssignStore is null, but other paths still keep master-slave semantics, for example only the leader runs `exploreTableRuntimes()` and `triggerTableAdded()` may still return early without creating local runtimes. This can leave the in-memory table runtime state inconsistent with the expected runtime state derived from the database.

wardlican: There is indeed a risk associated with this bug; I will fix it.

PR #3937, opened 2025-11-14
[Subtask]: In master-slave mode, each AMS should automatically senses the optimizer. 

### LINKED ISSUE #3929: [Subtask]: In master-slave mode, each AMS should automatically senses the optimizer.
### Search before asking

### Description

In master-slave mode, each AMS automatically detects the optimizer. Only the master node maintains the optimizer's heartbeat.

### Parent issue

### Code of Conduct

wardlican:
In master-slave mode, each optimizer needs to communicate with each ams node. Therefore, the registration mode of the optimizer needs to be modified in master-slave mode. The optimizer only reports heartbeats to the master node. Only the master node will maintain whether the optimizer's heartbeat has timed out. The slave node loads optimizer information from the database to detect whether there are any new or deleted optimizers.

Close #3929 .

## Brief change log
- In master-slave mode, only the master node is responsible for maintaining the addition and expiration of optimizers. 
- Slave nodes periodically detect the addition and deletion of optimizers from the database.

## How was this patch tested?

## Documentation


wardlican:
Please perform CR on these implementations.

amoro-ams/src/main/java/org/apache/amoro/server/ha/DataBaseHighAvailabilityContainer.java
```diff
@@ -301,6 +346,40 @@ private void onLeaderLost() {
...
+    }
+    try {
+      long currentTime = System.currentTimeMillis();
+      List<HaLeaseMeta> leases =
+          getAs(
+              HaLeaseMapper.class,
+              mapper -> mapper.selectLeasesByService(clusterName, OPTIMIZING_SERVICE));
+      for (HaLeaseMeta lease : leases) {
+        // Only include nodes with valid (non-expired) leases
+        if (lease.getLeaseExpireTs() != null && lease.getLeaseExpireTs() > currentTime) {
```
czy006: Checking only the lease expiration time may return nodes that have just expired but have not yet been cleared.

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
```diff
@@ -79,10 +189,54 @@ public void start() {
...
+      task = callAuthenticatedAms(amsUrl, (client, token) -> client.pollTask(token, threadId));
+      if (task != null) {
+        LOG.info(
+            "Optimizer executor[{}] polled task[{}] from AMS {}",
+            threadId,
+            task.getTaskId(),
+            amsUrl);
+      }
+    } catch (TException exception) {
+      LOG.error(
```
czy006, line 150: can we retry to get task? if have exception,maybe lost the task?

amoro-ams/src/main/java/org/apache/amoro/server/AmsAssignService.java
```diff
@@ -0,0 +1,488 @@
...
+    } else {
+      throw new IllegalArgumentException(
+          "Unsupported HA container type: " + haContainer.getClass().getName());
+    }
+  }
+
+  /**
+   * Start the assignment service. Only works in master-slave mode and when current node is leader.
+   */
+  public void start() {
```
czy006, line 99: These operations are non-transactional and appear to operate without using locks. Could this lead to inconsistent bucket allocations during Leader switchover?

czy006 (commented):
Thank you for your contribution; I spent some time reviewing it.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -610,6 +634,96 @@ protected void processTask(OptimizerKeepingTask keepingTask) {
...
+            dbOptimizersByToken.size(),
+            tokensToAdd.size(),
+            tokensToRemove.size(),
+            authOptimizers.size());
+      } catch (Exception e) {
+        LOG.error("Failed to load optimizers from database", e);
+      }
+    }
+
+    private void registerOptimizerWithoutPersist(OptimizerInstance optimizer) {
```
xxubai, line 705: nit: 
```suggestion
    private void syncOptimizerFromDatabase(OptimizerInstance optimizer) {
```

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -610,6 +634,96 @@ protected void processTask(OptimizerKeepingTask keepingTask) {
...
+          }
+        }
+
+        for (String token : tokensToRemove) {
+          removeOptimizerFromLocal(token);
+          LOG.debug("Removed optimizer {} (not in database)", token);
+        }
+
+        LOG.info(
+            "Synced optimizers from database: total={}, added={}, removed={}, current={}",
```
xxubai, line 695: Downgrade to `LOG.debug` when there are no changes, and only emit `LOG.info` when optimizers are actually added or removed.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
wardlican: The purpose of the `registerOptimizerWithoutPersist` method is to register optimizer information into local memory—without performing any persistence operations—when a follower node synchronizes this data from the database in a primary-replica configuration (since the data already resides in the database). Consequently, renaming this method to `syncOptimizerFromDatabase` would not be entirely appropriate, given the specific implementation logic involved.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
wardlican: Okay, I'll adjust the log level here to debug.

czy006:
I think the safer fix here is to separate optimizer lifecycle management from local task recovery.

At the moment, the follower path only syncs optimizer add/remove from DB, but it no longer performs the local queue recovery that used to happen in `OptimizerKeeper.processTask()`. As a result, follower-local `SCHEDULED/ACKED` tasks may stay stuck after the master deletes an optimizer from DB.

My suggestion is:
1. keep optimizer expiry / DB deletion as a leader-only responsibility
2. extract the local task recovery logic (`collectTasks(...)` + `retryTask(...)`) into a separate helper
3. invoke that helper not only in the leader path, but also in follower `onFollowerTick()` after syncing optimizers from DB

Importantly, I would not limit the fix to “retry tasks for removed tokens only”, because the existing recovery predicate also handles:
- stale token tasks
- SCHEDULED tasks that exceeded ack timeout
- ACKED tasks that exceeded execute timeout

So the follower should periodically recover local tasks for all local queues based on the current `authOptimizers` snapshot, instead of only updating the optimizer map.

I don't know how serious this problem is.The impact is on the resumption of operations under special circumstances.
WDYT? @xxubai @wardlican

zhoujinsong (approved):
LGTM overall.

Thanks a lot for pushing forward this feature!

czy006:
After our offline discussion, we confirmed that there will be a retry mechanism at this stage, and it will not be affected by this. This is an improvement plan, not a necessary condition.

PR #3941, opened 2025-11-14
[AMORO-3940] Flink config load will cause ClassCastException when dir…

### LINKED ISSUE #3940: [Bug]: Flink config load will cause ClassCastException when directly use Yaml load.
### What happened?

Since private Map<String, String> loadFlinkConfig() returns Map<String, String> ,
This line will cause some config with non-string value throws ClassCastException. 

### Affects Versions

master/0.8

_No response_

_No response_

### How to reproduce

_No response_

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

simonsssu:
Close #3940


zhoujinsong (approved):
The change looks good to me.

PR #3950, opened 2025-11-17
[AMORO-3928] Modify the optimizer to support obtaining tasks from each AMS node for processing.

### LINKED ISSUE #3928: [Subtask]: Modify the optimizer to support obtaining tasks from each AMS node for processing.
### Search before asking

### Description

In master-slave mode, the optimizer needs to support pulling optimization tasks from each ams node for processing.

### Parent issue

### Code of Conduct

wardlican:
In master-slave mode, the optimizer needs to support pulling optimization tasks from each ams node for processing.

Close #3928.

## Brief change log
1. In master-slave mode, the optimizer can pull tasks from each currently active AMS.
2. In master-slave mode, the AMS automatically detects the addition and timeout of the optimizer from the database.
-

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/test/java/org/apache/amoro/server/TestOptimizerGroupKeeper.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/TestOptimizerGroupKeeper.java
@@ -49,6 +49,7 @@ import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 @RunWith(Parameterized.class)
 public class TestOptimizerGroupKeeper extends AMSTableTestBase {
@@ -80,7 +81,7 @@ public class TestOptimizerGroupKeeper extends AMSTableTestBase {
   @Before
   public void prepare() throws Exception {
     optimizerRegistrar = registerInfo -> optimizingService().authenticate(registerInfo);
-    setupMockContainer();
+    setupMockContainer(() -> currentGroupName);
   }
 
   @After
@@ -124,9 +125,10 @@ public class TestOptimizerGroupKeeper extends AMSTableTestBase {
   }
 
   /** Setup mock container and inject it into Containers using reflection. */
-  private void setupMockContainer() throws Exception {
+  private void setupMockContainer(Supplier<String> targetGroupNameSupplier) throws Exception {
     MockOptimizerContainer mockContainer =
-        new MockOptimizerContainer(resourceAvailable, scaleOutCallCount, optimizerRegistrar);
+        new MockOptimizerContainer(
+            resourceAvailable, scaleOutCallCount, optimizerRegistrar, targetGroupNameSupplier);
 
     // Use reflection to set isInitialized to true
     DynFields.UnboundField<Boolean> initializedField =
@@ -339,14 +341,17 @@ public class TestOptimizerGroupKeeper extends AMSTableTestBase {
     private final AtomicBoolean resourceAvailable;
     private final AtomicInteger scaleOutCallCount;
     private final Function<OptimizerRegisterInfo, String> optimizerRegistrar;
+    private final Supplier<String> targetGroupNameSupplier;
 
     public MockOptimizerContainer(
         AtomicBoolean resourceAvailable,
         AtomicInteger scaleOutCallCount,
-        Function<OptimizerRegisterInfo, String> optimizerRegistrar) {
+        Function<OptimizerRegisterInfo, String> optimizerRegistrar,
+        Supplier<String> targetGroupNameSupplier) {
       this.resourceAvailable = resourceAvailable;
       this.scaleOutCallCount = scaleOutCallCount;
       this.optimizerRegistrar = optimizerRegistrar;
+      this.targetGroupNameSupplier = targetGroupNameSupplier;
     }
 
     @Override
@@ -354,6 +359,12 @@ public class TestOptimizerGroupKeeper extends AMSTableTestBase {
 
     @Override
     protected Map<String, String> doScaleOut(Resource resource) {
+      String targetGroup = targetGroupNameSupplier != null ? targetGroupNameSupplier.get() : null;
+      if (targetGroup != null && !targetGroup.equals(resource.getGroupName())) {
+        // Stale task from a previously leaked group; silently ignore to prevent cross-test
+        // contamination of scaleOutCallCount via the shared static Containers registry.
+        return Maps.newHashMap();
+      }
       scaleOutCallCount.incrementAndGet();
       if (!resourceAvailable.get()) {
         throw new RuntimeException("No resources available");
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,8 @@ amoro-web/node/
 public
 resources
 !javadoc/resources
+!amoro-ams/src/main/resources/
+!amoro-ams/src/main/resources/**
 
 !dist/src/main/amoro-bin/bin/
 !dist/src/main/amoro-bin/conf/
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Amoro support multiple processing engines for Mixed format as below:
 
 | Processing Engine | Version                | Batch Read  | Batch Write | Batch Overwrite | Streaming Read | Streaming Write | Create Table | Alter Table |
 |-------------------|------------------------|-------------|-------------|-----------------|----------------|-----------------|--------------|-------------|
-| Flink             | 1.17.x, 1.18.x         |  &#x2714;   |   &#x2714;   |       &#x2716;   |      &#x2714;   |       &#x2714;   |    &#x2714;   |   &#x2716;   |
+| Flink             | 1.17.x, 1.18.x, 1.19.x |  &#x2714;   |   &#x2714;   |       &#x2716;   |      &#x2714;   |       &#x2714;   |    &#x2714;   |   &#x2716;   |
 | Spark             | 3.3, 3.4, 3.5          |  &#x2714;   |   &#x2714;   |       &#x2714;   |      &#x2716;   |       &#x2716;   |    &#x2714;   |   &#x2714;   |
 | Hive              | 2.x, 3.x               |  &#x2714;  |   &#x2716;  |       &#x2714;  |      &#x2716;  |       &#x2716;  |    &#x2716;  |   &#x2714;  |
 | Trino             | 406                    |  &#x2714;  |   &#x2716;  |       &#x2714;  |      &#x2716;  |       &#x2716;  |    &#x2716;  |   &#x2714;  |
--- a/amoro-ams/pom.xml
+++ b/amoro-ams/pom.xml
@@ -219,6 +219,18 @@
         </dependency>
 
         <dependency>
+            <groupId>org.casbin</groupId>
+            <artifactId>jcasbin</artifactId>
+            <version>${jcasbin.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.20.0</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
         </dependency>
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
@@ -26,6 +26,7 @@ import org.apache.amoro.utils.MemorySize;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 
 public class AmoroManagementConf {
 
@@ -52,6 +53,59 @@ public class AmoroManagementConf {
           .stringType()
           .defaultValue("admin")
           .withDescription("The administrator password");
+
+  public static final ConfigOption<Boolean> AUTHORIZATION_ENABLED =
+      ConfigOptions.key("http-server.authorization.enabled")
+          .booleanType()
+          .defaultValue(false)
+          .withDescription("Whether to enable dashboard RBAC authorization.");
+
+  public static final ConfigOption<String> AUTHORIZATION_DEFAULT_ROLE =
+      ConfigOptions.key("http-server.authorization.default-role")
+          .stringType()
+          .noDefaultValue()
+          .withDescription(
+              "Optional default dashboard role for authenticated users without an LDAP role mapping.");
+
+  public static final ConfigOption<Boolean> AUTHORIZATION_LDAP_ROLE_MAPPING_ENABLED =
+      ConfigOptions.key("http-server.authorization.ldap-role-mapping.enabled")
+          .booleanType()
+          .defaultValue(false)
+          .withDescription("Whether to resolve dashboard roles from LDAP group membership.");
+
+  public static final ConfigOption<String> AUTHORIZATION_LDAP_ROLE_MAPPING_GROUP_MEMBER_ATTRIBUTE =
+      ConfigOptions.key("http-server.authorization.ldap-role-mapping.group-member-attribute")
+          .stringType()
+          .defaultValue("member")
+          .withDescription("LDAP group attribute that stores member references.");
+
+  public static final ConfigOption<String> AUTHORIZATION_LDAP_ROLE_MAPPING_USER_DN_PATTERN =
+      ConfigOptions.key("http-server.authorization.ldap-role-mapping.user-dn-pattern")
+          .stringType()
+          .noDefaultValue()
+          .withDescription(
+              "LDAP user DN pattern used to match group members. Use {0} as the username placeholder.");
+
+  public static final ConfigOption<List<Map<String, String>>>
+      AUTHORIZATION_LDAP_ROLE_MAPPING_GROUPS =
+          ConfigOptions.key("http-server.authorization.ldap-role-mapping.groups")
+              .mapType()
+              .asList()
+              .noDefaultValue()
+              .withDescription(
+                  "LDAP group-to-role mapping entries containing group-dn and role fields.");
+
+  public static final ConfigOption<String> AUTHORIZATION_LDAP_ROLE_MAPPING_BIND_DN =
+      ConfigOptions.key("http-server.authorization.ldap-role-mapping.bind-dn")
+          .stringType()
+          .defaultValue("")
+          .withDescription("Optional LDAP bind DN used when querying role-mapping groups.");
+
+  public static final ConfigOption<String> AUTHORIZATION_LDAP_ROLE_MAPPING_BIND_PASSWORD =
+      ConfigOptions.key("http-server.authorization.ldap-role-mapping.bind-password")
+          .stringType()
+          .defaultValue("")
+          .withDescription("Optional LDAP bind password used when querying role-mapping groups.");
 
   /** Enable master & slave mode, which supports horizontal scaling of AMS. */
   public static final ConfigOption<Boolean> USE_MASTER_SLAVE_MODE =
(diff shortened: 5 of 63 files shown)


wardlican:
Please help with the code review.

czy006:
After resolving any conflicts in the code, I will initiate the first review on my end cc @wardlican

czy006:
After resolving any conflicts in the code, I will initiate the first review on my end cc @wardlican

wardlican:
> After resolving any conflicts in the code, I will initiate the first review on my end cc @wardlican

Okay, I will resolve the conflicts based on the latest master branch and add support for a database-based storage solution for bucket allocation information, based on the latest HA (High Availability) scheme. @czy006

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/AbstractOptimizerOperator.java
```diff
@@ -86,25 +157,79 @@ private boolean shouldReturnNull(Throwable t) {
...
+   *
+   * @param amsUrl The AMS node URL to call
+   * @param operation The operation to execute
+   * @return The result of the operation
+   * @throws TException If the operation fails
+   */
+  protected <T> T callAuthenticatedAms(String amsUrl, AmsAuthenticatedCallOperation<T> operation)
       throws TException {
+    // Maximum retry time window for auth errors in master-slave mode (30 seconds)
+    long maxAuthRetryTimeWindow = TimeUnit.SECONDS.toMillis(90);
```
xxubai: 30 seconds or 90 seconds?

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/AbstractOptimizerOperator.java
```diff
@@ -86,25 +157,79 @@ private boolean shouldReturnNull(Throwable t) {
...
+    Long firstAuthErrorTime = null;
+
     while (isStarted()) {
       if (tokenIsReady()) {
         String token = getToken();
         try {
-          return operation.call(OptimizingClientPools.getClient(config.getAmsUrl()), token);
+          // Reset retry time on successful call
+          firstAuthErrorTime = null;
+          return operation.call(OptimizingClientPools.getClient(amsUrl), token);
```
xxubai, line 189: `firstAuthErrorTime` always be null in this pleace. I think it should reset after calling.
```suggestion
          T res = operation.call(OptimizingClientPools.getClient(amsUrl), token);
          // Reset retry time on successful call
          firstAuthErrorTime = null;
          
          return res;
```

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerToucher.java
```diff
@@ -32,7 +32,7 @@
 public class OptimizerToucher extends AbstractOptimizerOperator {
   private static final Logger LOG = LoggerFactory.getLogger(OptimizerToucher.class);
 
-  private TokenChangeListener tokenChangeListener;
+  private transient TokenChangeListener tokenChangeListener;
```
xxubai, line 35: Why set it to `transient`

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
```diff
@@ -49,6 +53,112 @@ public OptimizerExecutor(OptimizerConfig config, int threadId) {
   }
 
   public void start() {
+    // Check if in master-slave mode (node manager initialised for either ZK or DB HA)
+    boolean isMasterSlaveMode = getConfig().isMasterSlaveMode() && hasAmsNodeManager();
+
+    if (isMasterSlaveMode) {
+      // Master-slave mode: get node list and process tasks from each node
+      startMasterSlaveMode();
```
xxubai: If I understand correctly, single-node mode is a special case of multi-node mode. Could we refactor startMasterSlaveMode and startSingleNodeMode into a more generic implementation that supports multiple modes?

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
```diff
@@ -49,6 +53,112 @@ public OptimizerExecutor(OptimizerConfig config, int threadId) {
   }
 
   public void start() {
+    // Check if in master-slave mode (node manager initialised for either ZK or DB HA)
+    boolean isMasterSlaveMode = getConfig().isMasterSlaveMode() && hasAmsNodeManager();
```
xxubai: If the existence of `hasAmsNodeManager` already determines whether we use the multi-node task fetching strategy, do we still need the `master-slave-mode` configuration? 
If do not need, I recommand to remove this config property.

amoro-common/src/main/java/org/apache/amoro/OptimizerProperties.java
```diff
@@ -48,4 +48,5 @@ public class OptimizerProperties {
   public static final String OPTIMIZER_CACHE_MAX_TOTAL_SIZE_DEFAULT = "128mb";
   public static final String OPTIMIZER_CACHE_TIMEOUT = "cache-timeout";
   public static final String OPTIMIZER_CACHE_TIMEOUT_DEFAULT = "10min";
+  public static final String OPTIMIZER_MASTER_SLAVE_MODE = "master-slave-mode";
```
xxubai: It is a boolean property. Can you rename it to `OPTIMIZER_MASTER_SLAVE_MODE_ENABLED` or just remove it.

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
```diff
@@ -49,6 +53,112 @@ public OptimizerExecutor(OptimizerConfig config, int threadId) {
...
+      // Active-standby mode: use original logic
+      startSingleNodeMode();
+    }
+  }
+
+  /** Start in master-slave mode: get node list and process tasks from each AMS node. */
+  private void startMasterSlaveMode() {
+    // Dynamic polling interval control
+    long basePollInterval = TimeUnit.SECONDS.toMillis(1); // Base interval: 1 second
+    long maxPollInterval = TimeUnit.SECONDS.toMillis(30); // Max interval: 10 seconds
```
xxubai: Max interval: 10 or 30 seconds

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/AbstractOptimizerOperator.java
```diff
@@ -41,21 +43,90 @@ public class AbstractOptimizerOperator implements Serializable {
...
+      }
+    }
+    if (thriftAmsNodeManager != null) {
+      return thriftAmsNodeManager.getAllAmsUrls();
+    }
+    return Collections.singletonList(getConfig().getAmsUrl());
   }
 
   protected <T> T callAms(AmsCallOperation<T> operation) throws TException {
+    return callAms(getConfig().getAmsUrl(), operation);
```
xxubai, line 117: I have one concern:
This breaks authentication in multi-node mode: the optimizer still registers and heartbeats against only one AMS endpoint, but OptimizerExecutor now polls/acks/completes against every AMS node.

The flow becomes:
- the optimizer authenticates with AMS-A and gets a token
- it uses that same token to call AMS-B
- AMS-B does not know this token because it validates against its own local authOptimizers
- the request fails with `PluginRetryAuthException` and keep retry

Unless we add per-node registration/token propagation, cross-node task fetching will not work reliably.

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/AbstractOptimizerOperator.java
czy006: elapsedTime every time is near 0, maxAuthRetryTimeWindow is not to do it

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/AbstractOptimizerOperator.java
wardlican: The purpose of this window is to allow the optimizer to wait for the slave AMS node to complete its automatic registration (PLUGIN_RETRY_AUTH_ERROR_CODE) once a connection has been established. The process by which a slave node discovers its assigned buckets and completes registration is as follows:

- ZK Mode: Triggered by a ZK watch; typically completes within seconds.
- DB Mode: Discovered via polling based on the `node_heartbeat_ts` timestamp, with a polling interval of approximately 10–30 seconds.

Conclusion: A duration of 30 seconds is more appropriate.

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/AbstractOptimizerOperator.java
wardlican: Resetting the timer before every invocation causes the `elapsedTime >= maxAuthRetryTimeWindow` branch to never trigger, meaning `resetToken` is never executed. Consequently, the optimizer will retry indefinitely in master-slave mode and will be unable to recover in the event of a genuine authentication failure.

When the method returns successfully, `firstAuthErrorTime` acts as a local variable that is destroyed upon method exit; therefore, it requires no manual reset whatsoever. The correct fix is ​​simply to delete these two lines:

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerToucher.java
wardlican: The reason is that `AbstractOptimizerOperator` implements `Serializable`; consequently, `OptimizerToucher`—being a subclass—is also a serializable object. `TokenChangeListener` is an interface that is typically implemented by callers (such as the Flink or Spark optimizer) using lambda expressions or anonymous classes. If the `transient` keyword is omitted, a `NotSerializableException` will be thrown whenever `OptimizerToucher` is serialized (for instance, when Flink or Spark distributes tasks to executors).

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
wardlican: That's an excellent suggestion; I will optimize the implementation here.

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
wardlican: After switching to the unified pattern shown above, this issue no longer occurs.

amoro-common/src/main/java/org/apache/amoro/OptimizerProperties.java
wardlican: Okay, it will be renamed to OPTIMIZER_MASTER_SLAVE_MODE_ENABLED.

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerExecutor.java
wardlican: Remove the inner upper bound of `Math.min(backoffFactor, 3)` to allow the backoff to truly utilize `maxPollInterval`.
Following the fix, the actual progress of the backoff is as follows:

+1 | 2s
-- | --
+2 | 4s
+3 | 8s
+4 | 16s
+5 | 30s（maxPollInterval limit is in effect.）
+6 | 30s

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/AbstractOptimizerOperator.java
wardlican: Under the current design, the optimizer registers solely with the primary node to obtain a token. Each AMS node then automatically synchronizes the list of currently registered optimizers from the database; once this synchronization is complete, the optimizer is able to pull tasks from any of the other AMS nodes. Shifting to a model where each AMS node independently maintains tokens would introduce excessive complexity.

amoro-ams/src/test/java/org/apache/amoro/server/TestOptimizerGroupKeeper.java
```diff
@@ -80,7 +81,7 @@ public static Object[] parameters() {
   @Before
   public void prepare() throws Exception {
     optimizerRegistrar = registerInfo -> optimizingService().authenticate(registerInfo);
-    setupMockContainer();
+    setupMockContainer(() -> currentGroupName);
```
xxubai, line 84: Could you open a separate PR to fix this unit test?

wardlican: So, does this need to be reverted?

xxubai: Yeah...

xxubai: Or you can wait until the fix PR is merged and then rebase.

wardlican: Okay, I will reverted it and fix it in another PR. https://github.com/apache/amoro/issues/4151

amoro-ams/src/main/java/org/apache/amoro/server/DBBucketAssignStore.java
```diff
@@ -180,6 +180,25 @@ public void updateLastUpdateTime(AmsServerInfo nodeInfo) throws BucketAssignStor
     }
   }
 
+  @Override
+  public List<AmsServerInfo> getAliveNodes() throws BucketAssignStoreException {
+    try {
+      List<BucketAssignmentMeta> rows =
+          getAs(BucketAssignMapper.class, mapper -> mapper.selectAllByCluster(clusterName));
```
czy006, line 190: `getAliveNodes()` directly iterates through all `bucket_assignments` and returns them (only verifying the port), without filtering by `node_heartbeat_ts`. Could this result in the optimizer's node list containing offline nodes, triggering additional retries?

amoro-ams/src/main/resources/mysql/upgrade.sql
```diff
@@ -164,12 +164,13 @@ ADD COLUMN `process_parameters` mediumtext COMMENT 'Table process parameters';
 
 -- ADD table bucket_assignments for storing assigned info
 CREATE TABLE IF NOT EXISTS bucket_assignments (
-                                                  cluster_name       VARCHAR(64)   NOT NULL COMMENT 'AMS cluster name',
+    cluster_name       VARCHAR(64)   NOT NULL COMMENT 'AMS cluster name',
     node_key           VARCHAR(256) NOT NULL COMMENT 'Node key (host:thriftBindPort)',
     server_info_json   TEXT         NULL COMMENT 'JSON encoded AmsServerInfo',
     assignments_json   TEXT         NULL COMMENT 'JSON array of bucket IDs',
     last_update_time   BIGINT       NOT NULL DEFAULT 0 COMMENT 'Last update timestamp (ms since epoch)',
+    node_heartbeat_ts  BIGINT       NOT NULL DEFAULT 0 COMMENT 'Per-node heartbeat timestamp updated only by the owning node (ms since epoch)',
```
czy006: `ALTER TABLE ... ADD COLUMN (node_heartbeat_ts)`

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/AbstractOptimizerOperator.java
```diff
@@ -86,25 +157,77 @@ private boolean shouldReturnNull(Throwable t) {
...
+   * @throws TException If the operation fails
+   */
+  protected <T> T callAuthenticatedAms(String amsUrl, AmsAuthenticatedCallOperation<T> operation)
       throws TException {
+    // Maximum retry time window for auth errors in master-slave mode (30 seconds)
+    long maxAuthRetryTimeWindow = TimeUnit.SECONDS.toMillis(30);
+    Long firstAuthErrorTime = null;
+
     while (isStarted()) {
       if (tokenIsReady()) {
```
czy006, line 186: I suspect that if a node keeps failing, it will get stuck on that node for a long time, and subsequent nodes will stop polling.We can Enforce per-node retry budget (attempt count or retry window), then fail fast to continue with next node.

amoro-ams/src/main/java/org/apache/amoro/server/DBBucketAssignStore.java
wardlican: There is indeed a window-period issue here; I'll optimize it.

amoro-ams/src/main/resources/mysql/upgrade.sql
wardlican: Okay, I will separate this change.

amoro-common/src/main/java/org/apache/amoro/OptimizerProperties.java
```diff
@@ -48,5 +48,5 @@ public class OptimizerProperties {
   public static final String OPTIMIZER_CACHE_MAX_TOTAL_SIZE_DEFAULT = "128mb";
   public static final String OPTIMIZER_CACHE_TIMEOUT = "cache-timeout";
   public static final String OPTIMIZER_CACHE_TIMEOUT_DEFAULT = "10min";
-  public static final String OPTIMIZER_MASTER_SLAVE_MODE = "master-slave-mode";
+  public static final String OPTIMIZER_MASTER_SLAVE_MODE_ENABLED = "master-slave-mode";
```
xxubai: `master-slave-mode-enabled`

wardlican: okay,i will fix it.

amoro-optimizer/amoro-optimizer-common/src/main/java/org/apache/amoro/optimizer/common/OptimizerConfig.java
```diff
@@ -214,6 +220,14 @@ public void setCacheTimeout(String cacheTimeout) {
...
+  public boolean isMasterSlaveMode() {
+    return masterSlaveMode;
+  }
+
+  public void setMasterSlaveMode(boolean masterSlaveMode) {
+    this.masterSlaveMode = masterSlaveMode;
+  }
+
   @Override
   public String toString() {
```
xxubai, line 232: nit: missing field `masterSlaveMode` in `toString`

amoro-ams/src/main/resources/postgres/upgrade.sql
```diff
@@ -224,12 +224,15 @@ ADD COLUMN process_parameters text;
...
     assignments_json   TEXT         NULL,
     last_update_time   BIGINT       NOT NULL DEFAULT 0,
     PRIMARY KEY (cluster_name, node_key)
-    );
+);
 
 CREATE INDEX IF NOT EXISTS idx_bucket_assignments_cluster ON bucket_assignments (cluster_name);
+
+-- ADD node_heartbeat_ts to table bucket_assignments
+ALTER TABLE table_process ADD COLUMN node_heartbeat_ts BIGINT NOT NULL DEFAULT 0;
```
xxubai:
```suggestion
ALTER TABLE bucket_assignments ADD COLUMN node_heartbeat_ts BIGINT NOT NULL DEFAULT 0;
```

czy006:
LGTM.After the overall functionality is completed, we need to conduct a comprehensive test

PR #3905, opened 2025-11-06
[AMORO-3873] Support Bearer/JWT authentication

turboFei:
## Why are the changes needed?

Close #3873.

## Brief change log

- Support Bearer/JWT authentication. see https://www.jwt.io/introduction#what-is-json-web-token 

## How was this patch tested?

- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (yes / no)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #3953, opened 2025-11-17
[AMORO-3873][FOLLOWUP] Rename rest auth bearer type to JWT

turboFei:

Followup #3905 

I plan to support thrift authentication for amoro thrift client connection and look into the hive code.

For hive `metastore.catalog.servlet.auth`, its valid auth types are `"none", "simple", "jwt", "oauth2"`.
https://github.com/apache/hive/blob/e34e87faf32b0167f438e30344f0a04fe10f6575/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/conf/MetastoreConf.java#L1883-L1886
image

And for JWT and oauth2, they both use `Bearer` schema.
https://github.com/apache/hive/blob/e34e87faf32b0167f438e30344f0a04fe10f6575/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/ServletSecurity.java#L277-L296

image

So, I think it is better to rename the rest-auth type from bearer to jwt to narrow the scope for the interface.

Also I believe it is helpful to align the auth types we plan to support in the future.

## Brief change log

- rename the rest auth bearer type to jwt, it has not been released yet, no impaction.

## How was this patch tested?

## Documentation


amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
```diff
@@ -275,8 +275,7 @@ public class AmoroManagementConf {
       ConfigOptions.key("http-server.rest-auth-type")
           .stringType()
           .defaultValue("token")
-          .withDescription(
-              "The authentication used by REST APIs, token (default), basic or bearer.");
+          .withDescription("The authentication used by REST APIs, token (default), basic or jwt.");
```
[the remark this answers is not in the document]
xxubai: Do we have `token` implementation?

amoro-common/src/main/java/org/apache/amoro/authentication/TokenAuthenticationProvider.java
```diff
@@ -24,12 +24,11 @@
 
 public interface TokenAuthenticationProvider {
   /**
-   * TokenAuthenticationProvider is used by the Amoro server authentication layer to validate Bearer
-   * tokens, such as JWT (JSON Web Token), provided in client requests. If the token is invalid,
-   * expired, or fails signature verification, a {@link SignatureCheckException} should be thrown to
-   * deny access.
+   * TokenAuthenticationProvider is used by the Amoro server authentication layer to validate JSON
```
xxubai, line 27: Should we limit this to JWT only? I see dashboard will check the IP only currently

turboFei: the default jwt auth provider class name is empty now

amoro-common/src/main/java/org/apache/amoro/authentication/TokenAuthenticationProvider.java
```diff
@@ -24,12 +24,11 @@
...
+   * Web Token (JWT) provided in client requests. If the token is invalid, expired, or fails
```
turboFei, line 28: Already mentioned JWT here.
 @xxubai 

It should be Ok to keep the class name as TokenAuthenticationProvider.

For OAuth2 authentication, maybe we can also implement this interface.

https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/auth/oauth2/OAuth2Authenticator.java

https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/auth/oauth2/JWTAccessTokenAuthenticator.java

PR #3969, opened 2025-11-25
[Subtask]: Supports forwarding OpenAPI requests to the master node in master-slave mode.

### LINKED ISSUE #3963: [Subtask]: Supports forwarding OpenAPI requests to the master node in master-slave mode.
### Search before asking

### Description

Supports forwarding OpenAPI requests to the master node in master-slave mode.

### Parent issue

### Code of Conduct

wardlican:
In master-slave mode, requests from the slave node's OpenAPI will be automatically forwarded to the master node.

Close  #3963.

## Brief change log

Add the RequestForwarder class to intercept and forward requests in http.before.

## How was this patch tested?

## Documentation


wardlican:
Please help with the code review.

zhoujinsong:
Thanks for the contribution! After reviewing the current master branch architecture, I think this request forwarding mechanism may no longer be necessary. Here's the reasoning:

**Why forwarding is not needed:**

1. **All write operations go directly to DB** — `createCatalog`, `dropCatalog`, `updateCatalog`, and table creation all write to the database directly. No in-memory leader-exclusive state is involved. Any node can handle these requests independently.

2. **Optimizing status is fully persisted** — Every `OptimizingStatus` transition is written to `table_runtime.status_code` synchronously (via `StatedPersistentBase.invokeConsistency()`). `TableController` already reads status directly from DB via `TableRuntimeMapper.selectRuntime()`, not from in-memory `tableRuntimeMap`. Any node can serve these queries correctly.

3. **Optimizing process history is also DB-backed** — `getOptimizingProcesses` queries `optimizing_process` table directly. No forwarding needed.

4. **`DefaultTableService` handles cross-node table sync natively** — In master-slave mode, each node periodically syncs its assigned bucket tables via `syncBucketTables()`. The bucket routing is already handled at the data layer, not the HTTP layer.

The only purely in-memory state is the optimizer heartbeat tracking and in-flight task assignment inside `OptimizingQueue`, but these are operational monitoring concerns, not correctness-critical for API responses.

Given this, adding an HTTP proxy layer with circuit breakers, retry logic, and exception-as-control-flow introduces significant complexity without a clear benefit. A simpler approach would be to route Dashboard traffic through a load balancer.

Would be happy to discuss further if there are specific cases I missed.

wardlican:
> Thanks for the contribution! After reviewing the current master branch architecture, I think this request forwarding mechanism may no longer be necessary. Here's the reasoning:
> 
> **Why forwarding is not needed:**
> 
> 1. **All write operations go directly to DB** — `createCatalog`, `dropCatalog`, `updateCatalog`, and table creation all write to the database directly. No in-memory leader-exclusive state is involved. Any node can handle these requests independently.
> 2. **Optimizing status is fully persisted** — Every `OptimizingStatus` transition is written to `table_runtime.status_code` synchronously (via `StatedPersistentBase.invokeConsistency()`). `TableController` already reads status directly from DB via `TableRuntimeMapper.selectRuntime()`, not from in-memory `tableRuntimeMap`. Any node can serve these queries correctly.
> 3. **Optimizing process history is also DB-backed** — `getOptimizingProcesses` queries `optimizing_process` table directly. No forwarding needed.
> 4. **`DefaultTableService` handles cross-node table sync natively** — In master-slave mode, each node periodically syncs its assigned bucket tables via `syncBucketTables()`. The bucket routing is already handled at the data layer, not the HTTP layer.
> 
> The only purely in-memory state is the optimizer heartbeat tracking and in-flight task assignment inside `OptimizingQueue`, but these are operational monitoring concerns, not correctness-critical for API responses.
> 
> Given this, adding an HTTP proxy layer with circuit breakers, retry logic, and exception-as-control-flow introduces significant complexity without a clear benefit. A simpler approach would be to route Dashboard traffic through a load balancer.
> 
> Would be happy to discuss further if there are specific cases I missed.



> Thanks for the contribution! After reviewing the current master branch architecture, I think this request forwarding mechanism may no longer be necessary. Here's the reasoning:
> 
> **Why forwarding is not needed:**
> 
> 1. **All write operations go directly to DB** — `createCatalog`, `dropCatalog`, `updateCatalog`, and table creation all write to the database directly. No in-memory leader-exclusive state is involved. Any node can handle these requests independently.
> 2. **Optimizing status is fully persisted** — Every `OptimizingStatus` transition is written to `table_runtime.status_code` synchronously (via `StatedPersistentBase.invokeConsistency()`). `TableController` already reads status directly from DB via `TableRuntimeMapper.selectRuntime()`, not from in-memory `tableRuntimeMap`. Any node can serve these queries correctly.
> 3. **Optimizing process history is also DB-backed** — `getOptimizingProcesses` queries `optimizing_process` table directly. No forwarding needed.
> 4. **`DefaultTableService` handles cross-node table sync natively** — In master-slave mode, each node periodically syncs its assigned bucket tables via `syncBucketTables()`. The bucket routing is already handled at the data layer, not the HTTP layer.
> 
> The only purely in-memory state is the optimizer heartbeat tracking and in-flight task assignment inside `OptimizingQueue`, but these are operational monitoring concerns, not correctness-critical for API responses.
> 
> Given this, adding an HTTP proxy layer with circuit breakers, retry logic, and exception-as-control-flow introduces significant complexity without a clear benefit. A simpler approach would be to route Dashboard traffic through a load balancer.
> 
> Would be happy to discuss further if there are specific cases I missed.

Yes, all current state information is retrieved from the database, and any updated states are also synchronized back to it. Since every AMS node synchronizes its state information from the database, this feature is not required.

czy006:
This is an enhanced feature and does not affect our usage.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/DashboardServer.java
```diff
@@ -412,6 +416,41 @@ private EndpointGroup apiGroup() {
...
+
+        // Enhanced error handling for forwarding failures
+        String errorMessage = "Failed to forward request to leader node: " + e.getMessage();
+
+        // Check if circuit breaker is open
+        if (requestForwarder.isCircuitBreakerOpenForMonitoring()) {
+          errorMessage += " (Circuit breaker is open due to repeated failures)";
+        }
+
+        ctx.json(
```
czy006, line 445: In `DashboardServer.preHandleRequest`, the IOException branch writes `ctx.json(...)` but does not explicitly set `ctx.status(...)`.  
This can lead to response-status ambiguity (often 200 with error payload), which is unsafe for clients and retry logic.

Suggested fix:

Set explicit status before `ctx.json(...)`, preferably `HttpCode.SERVICE_UNAVAILABLE` (503) for forwarding failures.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/RequestForwarder.java
```diff
@@ -0,0 +1,521 @@
...
+    RequestConfig requestConfig =
+        RequestConfig.custom()
+            .setConnectTimeout(timeoutMs)
+            .setSocketTimeout(timeoutMs)
+            .setConnectionRequestTimeout(timeoutMs)
+            .build();
+    this.httpClient =
+        HttpClients.custom()
+            .setDefaultRequestConfig(requestConfig)
+            .setMaxConnTotal(100)
```
czy006: `ha.request-forwarder.max-connections` and `ha.request-forwarder.max-connections-per-route` are read and logged, but `RequestForwarder` still hardcodes `setMaxConnTotal(100)` and `setMaxConnPerRoute(20)`.

Suggested fix:
Pass both values into `RequestForwarder` constructor and use them when building the HTTP client.
Also add unit tests to verify configured values are actually used.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/RequestForwarder.java
```diff
@@ -0,0 +1,521 @@
...
+    String lowerHeaderName = headerName.toLowerCase();
+    // Skip headers that should not be forwarded or should be recalculated
+    return lowerHeaderName.equals("connection")
+        || lowerHeaderName.equals("keep-alive")
+        || lowerHeaderName.equals("transfer-encoding")
+        || lowerHeaderName.equals("content-length") // Let Javalin calculate this automatically
+        || lowerHeaderName.equals("content-encoding"); // Skip content-encoding to avoid issues
+  }
+
+  public void close() {
```
czy006, line 505: `RequestForwarder` exposes `close()`, but its instance is not retained/closed by `AmoroServiceContainer` during shutdown.  
This can leave HTTP client resources unmanaged.

Suggested fix:
Store `RequestForwarder` as a container field and close it in `disposeRestService()` / `dispose()`.

czy006:
In master-slave mode, startup uses `registerAndElect()` + `startOptimizingService()` and does not go through `waitLeaderShip()`.  
However, in ZK HA, leader REST endpoint data is written in `waitLeaderShip()`, while forwarding reads from `getLeaderNodeInfo()` (table-service master path).  
This creates a gap where followers cannot resolve leader REST info and forwarding fails even when the cluster is healthy.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/DashboardServer.java
wardlican: okay,I will fix it.

wardlican:
> In master-slave mode, startup uses `registerAndElect()` + `startOptimizingService()` and does not go through `waitLeaderShip()`. However, in ZK HA, leader REST endpoint data is written in `waitLeaderShip()`, while forwarding reads from `getLeaderNodeInfo()` (table-service master path). This creates a gap where followers cannot resolve leader REST info and forwarding fails even when the cluster is healthy.

It seems there was an issue while resolving conflicts during the merge process. I'll check and fix it.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/RequestForwarder.java
wardlican: okay,I will fix it.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/RequestForwarder.java
wardlican: okay,I will fix it.

czy006:
After offline confirmation, the node can currently handle the relevant requests. However, this is an outdated implementation, and the PR will be closed.

PR #3982, opened 2025-11-29
[AMORO-3981] Automatic generate amoro configuration docs 

turboFei:
## Why are the changes needed?

Close #3981.

Automatic generate amoro configuration docs.

## Brief change log

- Automatic generate amoro configuration docs

## How was this patch tested?

- [x] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [x] Add screenshots for manual tests if appropriate

```
(cd site/amoro-docs && hugo serve)
```
image

image

- [x] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (yes / no)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #3987, opened 2025-11-30
[AMORO-3981][FOLLOWUP] Refine the duration to string format 

turboFei:

Close #xxx.
In https://github.com/apache/amoro/pull/3982, I support to generate the conf docs automatically.

However, the default value looks not human readable.

All the duration default value is in ns time unit.
image

In this PR, we refine the Duration format to string.

## Brief change log

- Refine the duration format to string

## How was this patch tested?

- [] Add screenshots for manual tests if appropriate

## Documentation

### DIFF
--- a/docs/configuration/ams-config.md
+++ b/docs/configuration/ams-config.md
@@ -52,10 +52,10 @@ table td:last-child, table th:last-child { width: 40%; word-break: break-all; }
 | clean-dangling-delete-files.enabled | true | Enable dangling delete files cleaning. |
 | clean-dangling-delete-files.thread-count | 10 | The number of threads used for dangling delete files cleaning. |
 | clean-orphan-files.enabled | true | Enable orphan files cleaning. |
-| clean-orphan-files.interval | 86400000 ms | Interval for cleaning orphan files. |
+| clean-orphan-files.interval | 1 d | Interval for cleaning orphan files. |
 | clean-orphan-files.thread-count | 10 | The number of threads used for orphan files cleaning. |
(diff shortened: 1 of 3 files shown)


turboFei:
I will regenerate the docs after #3982 got merged.

turboFei:
cc @huyuanfeng2018

huyuanfeng2018 (commented):
I think need to refer to it here: `org.apache.amoro.config.ConfigHelpers#TimeUtils`
For example, using `1 min` display on the page is better than using `PT1M`

turboFei:
> I think need to refer to it here: `org.apache.amoro.config.ConfigHelpers#TimeUtils`

Thanks for the suggestion, I have updated.

docs/configuration/ams-config.md
```diff
@@ -52,10 +52,10 @@ table td:last-child, table th:last-child { width: 40%; word-break: break-all; }
 | clean-dangling-delete-files.enabled | true | Enable dangling delete files cleaning. |
 | clean-dangling-delete-files.thread-count | 10 | The number of threads used for dangling delete files cleaning. |
 | clean-orphan-files.enabled | true | Enable orphan files cleaning. |
-| clean-orphan-files.interval | 86400000 ms | Interval for cleaning orphan files. |
+| clean-orphan-files.interval | 1 d | Interval for cleaning orphan files. |
```
turboFei, line 55:
```
              // Use the largest unit if all divide evenly
              .orElse(orderedUnits.get(orderedUnits.size() - 1));
```

It is used to show `1 d` instead of `86400000 ms`.

PR #3992, opened 2025-12-04
[AMORO-3989] Amoro's support for the domestic database Damang

linjianchang:
[AMORO-3989] Amoro's support for the domestic database Damang

### DIFF
--- a/amoro-ams/pom.xml
+++ b/amoro-ams/pom.xml
@@ -229,6 +229,12 @@
         </dependency>
 
         <dependency>
+            <groupId>com.dameng</groupId>
+            <artifactId>DmJdbcDriver18</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
             <scope>provided</scope>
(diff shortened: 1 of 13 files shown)


amoro-ams/pom.xml
```diff
@@ -228,6 +228,11 @@
             <artifactId>derby</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.dameng</groupId>
+            <artifactId>DmJdbcDriver18</artifactId>
+        </dependency>
```
xxubai, line 235: It should be provided

linjianchang: > It should be provided

企业微信截图_17659411772400
I see that the JDBC driver packages for PostgreSQL and Derby are not provided

xxubai: I don’t think Dameng needs to be included in the release. Similar to MySQL, users who need it can add the JDBC dependency manually, rather than having it as a standard component.

linjianchang: > I don’t think Dameng needs to be included in the release. Similar to MySQL, users who need it can add the JDBC dependency manually, rather than having it as a standard component.

Aleady modified!

linjianchang: @xxubai Thanks for your review. Do you have any other suggestions

xxubai:
Adding support for other database connectors is technically straightforward. However, the primary concern is long-term maintenance. Is there sufficient community interest and commitment to help maintain this connector over time? I would like to invite more community members to participate in the discussion. @zhoujinsong @czy006 @Aireed

PR #4005, opened 2025-12-10
[AMORO-3804] Skip RUNTIME_CONTEXT_CACHE for TableMetaStore with local configuration

### LINKED ISSUE #3804: [Bug]: TableMetaStore's equals() ignores configuration, causing incorrect RUNTIME_CONTEXT_CACHE hits
### What happened?

When creating a `TableMetaStore` with a configuration, `getRuntimeContext` will create a `RuntimeContext `and store it in `RUNTIME_CONTEXT_CACHE`, where the key is the `TableMetaStore` itself:
Image

However, let's look at the overridden equals method of `TableMetaStore`:
Image

It does not use the configuration, and all these fields are null. That is to say, if there is another `TableMetaStore `built with another configuration, they are essentially different, but the equals method will consider them the same. 
At this point, when the second created `TableMetaStore` retrieves the `RuntimeContext` of the previous `TableMetaStore` from `RUNTIME_CONTEXT_CACHE` via getRuntimeContext, it will overwrite the configuration in the `RuntimeContext` of the previous one.
This may not only cause the issue of the configuration being overwritten, but also potentially lead to other problems.

### Affects Versions

master

_No response_

_No response_

### How to reproduce

I discovered this risk while attempting to make some custom modifications to Amoro, but it hasn't caused any actual problems.

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

turboFei:

Close #3804.

Skip RUNTIME_CONTEXT_CACHE for TableMetaStore with local configuration.

See  
https://github.com/apache/amoro/blob/da41c3551397e39c35903d3942afb6f8e097bfc6/amoro-common/src/main/java/org/apache/amoro/table/TableMetaStore.java#L768-L772

https://github.com/apache/amoro/blob/da41c3551397e39c35903d3942afb6f8e097bfc6/amoro-common/src/main/java/org/apache/amoro/table/TableMetaStore.java#L170-L182

## Brief change log

- Skip RUNTIME_CONTEXT_CACHE for TableMetaStore with local configuration

## How was this patch tested?

## Documentation


turboFei:
cc @HuangFru @zhoujinsong

amoro-common/src/main/java/org/apache/amoro/table/TableMetaStore.java
```diff
@@ -179,6 +181,7 @@ private TableMetaStore(Configuration configuration) {
     this.krbPrincipal = null;
     this.accessKey = null;
     this.secretKey = null;
+    this.configuration = configuration;
```
HuangFru: Why not just create a `RuntimeContext` here and set it. So you don't need to modify `getRuntimeContext()`.
The current code is OK with me, just my humble take.

turboFei: got it, updated, thanks

turboFei:
> I'm still a bit confused. Why is it that simply not storing the configuration in the local cache, without changing the equals and hashCode functions, can prevent multiple TableMetaStores from being equal?


Here is the comments: https://github.com/apache/amoro/issues/3804#issuecomment-3346501270

PR #4015, opened 2025-12-18
[WIP][AMORO-4014] FileIO should be shared across tables within the same catalog

### LINKED ISSUE #4014: [Improvement]: FileIO should be shared across tables within the same catalog
### Search before asking

In our long-running ams process we see too many OSS/HTTP connection pools. Heap analysis shows a very large number of org.apache.http.impl.conn.PoolingHttpClientConnectionManager, which suggests Iceberg FileIO (or its underlying HTTP client) is being created per table and not reused.
Since FileIO is expected to be thread-safe, we think FileIO (or at least the underlying HTTP/OSS client) should be shared across tables within the same catalog to avoid excessive pools, memory usage, and sockets.

### How should we improve?

Add catalog-scoped reuse/caching of FileIO (or its internal HTTP client/connection manager), so tables with the same catalog/config share the same instance and it’s closed only on catalog/service shutdown.

### Subtasks

_No response_

### Code of Conduct

turboFei:

Close #4014.
 
Currently, each table creates its own FileIO instance, even when tables belong to the same catalog. This leads to excessive  connection pools in long-running AMS processes, consuming unnecessary resources.

## Brief change log

This PR implemented a caching mechanism for FileIO instances:
- Basic FileIO instances are now shared at the catalog level (by TableMetaStore)
- RecoverableHadoopFileIO instances (with trash management) are shared per table
- Added safe thread-safe caching using ConcurrentHashMap

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/utils/InternalTableUtil.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/utils/InternalTableUtil.java
@@ -32,6 +32,9 @@ import org.apache.amoro.io.AuthenticatedFileIO;
 import org.apache.amoro.io.AuthenticatedFileIOs;
 import org.apache.amoro.properties.CatalogMetaProperties;
 import org.apache.amoro.shade.guava32.com.google.common.annotations.VisibleForTesting;
+import org.apache.amoro.shade.guava32.com.google.common.cache.CacheBuilder;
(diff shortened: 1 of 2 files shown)


turboFei:
cc @liaoyt

liaoyt:
> cc @liaoyt

LGTM

turboFei:
This PR has not been finished, due the fileIo might be closed by somewhere

PR #4038, opened 2026-01-14
[AMORO-4037] Add support for dynamic refresh interval for table metadata refreshing in TableRuntimeRefreshExecutor

### LINKED ISSUE #4037: [Feature]: Dynamic refresh interval for table
### Description

_No response_

### Use case/motivation

_No response_

### Describe the solution

_No response_

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

zhangwl9:

Close #4037.

## Brief change log

- The new feature involves two methods in "TableRuntimeRefreshExecutor".
  - #getNextExecutingTime()
    - When dynamic refresh interval is enabled，retrieving the latest refresh interval from "DefaultTableRuntime" to determine the next execution time for "TableRuntimeRefreshExecutor#execute"
  - #execute()
    - Depending on the evaluated results, determine whether the table requires optimization.
    - When the evaluation table requires no optimization, extend the interval for its next refresh (incrementally increasing the interval by a fixed step).
    - When evaluating table requiring optimization, shorten its next refresh interval (halving the interval each time).
- dynamic refresh interval in [ refresh-tables.interval(1 min)，Long.MAX_VALUE]
## How was this patch tested?

## Documentation


amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -56,19 +56,29 @@ protected boolean enabled(TableRuntime tableRuntime) {
   @Override
   protected long getNextExecutingTime(TableRuntime tableRuntime) {
     DefaultTableRuntime defaultTableRuntime = (DefaultTableRuntime) tableRuntime;
+
+    if (defaultTableRuntime.getOptimizingConfig().getRefreshTableAdaptiveEnabled()) {
+      long newInterval = defaultTableRuntime.getLatestRefreshInterval();
+      if (newInterval > 0) {
+        return newInterval;
```
zhangwl9: The newInterval in [fixed_interval(1min), max_interval(1h)], the maximum value of the newInterval exceeds the fixed_interval(48 min) below.

zhangwl9:
@zhoujinsong @turboFei @xxubai Could you help me review it when you are free? Thanks!

docs/user-guides/configurations.md
```diff
@@ -43,25 +43,28 @@ modified through [Alter Table](../using-tables/#modify-table) operations.
...
+| self-optimizing.filter                               | NULL             | Filter conditions for self-optimizing, using SQL conditional expressions, without supporting any functions. For the timestamp column condition, the ISO date-time formatter must be used. For example: op_time > '2007-12-03T10:15:30'. |
+| self-optimizing.refresh-table.adaptive.enabled       | false            | Whether to enable adaptive refresh interval for refreshing table metadata                                                                                                                                                           |
+| self-optimizing.refresh-table.adaptive.max-interval  | 3600000(1 hour)  | The maximum time interval in milliseconds to refresh table metadata                                                                                                                                                                 |
```
klion26: Maybe we can set the max-interval to an unset value, so that we can keep the behavior same as before the current pr.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
```diff
@@ -160,6 +160,18 @@ private TableProperties() {}
   public static final String SNAPSHOT_MIN_COUNT = "snapshot.keep.min-count";
   public static final int SNAPSHOT_MIN_COUNT_DEFAULT = 1;
 
+  public static final String SELF_OPTIMIZING_REFRESH_TABLE_ADAPTIVE_ENABLED =
+      "self-optimizing.refresh-table.adaptive.enabled";
```
klion26: If we need to set `self-optimizing.refresh-table.adaptive.enabled` and `self-optimizing.refresh-table.adaptive.max-interval`, how about using only `self-optimizing.refresh-table.adaptive.max-interval`? Some value (like -1 or zero) means that disable it, and some other value means that to enable it.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
```diff
@@ -160,6 +160,18 @@ private TableProperties() {}
...
+  public static final boolean SELF_OPTIMIZING_REFRESH_TABLE_ADAPTIVE_ENABLED_DEFAULT = false;
+  public static final String SELF_OPTIMIZING_REFRESH_TABLE_ADAPTIVE_MAX_INTERVAL =
+      "self-optimizing.refresh-table.adaptive.max-interval";
+  public static final long SELF_OPTIMIZING_REFRESH_TABLE_ADAPTIVE_MAX_INTERVAL_DEFAULT =
```
klion26: How about add the time unit in the name, `SELF_OPTIMIZING_REFRESH_TABLE_ADAPTIVE_MAX_INTERVAL_DEFAULT` -> `SELF_OPTIMIZING_REFRESH_TABLE_ADAPTIVE_MAX_INTERVAL_MS_DEFAULT`

klion26: Same as the other places

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -56,19 +56,29 @@ protected boolean enabled(TableRuntime tableRuntime) {
...
     OptimizingConfig optimizingConfig = tableRuntime.getOptimizingConfig();
-    if (optimizingConfig.isEnabled()
-        && tableRuntime.getOptimizingStatus().equals(OptimizingStatus.IDLE)) {
+    boolean optimizingEnabled = optimizingConfig.isEnabled();
+    if (optimizingEnabled && tableRuntime.getOptimizingStatus().equals(OptimizingStatus.IDLE)) {
 
       if (optimizingConfig.isMetadataBasedTriggerEnabled()
           && !MetadataBasedEvaluationEvent.isEvaluatingNecessary(
               optimizingConfig, table, tableRuntime.getLastPlanTime())) {
+        tableRuntime.setLatestEvaluatedNeedOptimizing(false);
```
klion26: Is it better that we change the return value of `tryEvaluatingPendingInput` to `boolean` -- indicate whether we need to optimize after evaluating? 
So that we can call `tableRuntime.setLatestEvaluteNeedOptimizing` in one place in `execute` below,

Please let me what do you think about it.

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -129,9 +148,124 @@ public void execute(TableRuntime tableRuntime) {
...
         tryEvaluatingPendingInput(defaultTableRuntime, mixedTable);
+      } else {
+        logger.debug("{} optimizing is not necessary", defaultTableRuntime.getTableIdentifier());
+        defaultTableRuntime.setLatestEvaluatedNeedOptimizing(false);
+      }
+
+      // Update adaptive interval according to evaluated result.
+      if (defaultTableRuntime.getOptimizingConfig().getRefreshTableAdaptiveEnabled()) {
+        long newInterval = getAdaptiveExecutingInterval(defaultTableRuntime);
+        defaultTableRuntime.setLatestRefreshInterval(newInterval);
```
klion26, line 167: Maybe we can call `tableRuntime.setLatestEvaluatedNeedOptimizing(true);` and  `defaultTableRuntime.setLatestRefreshInterval(newInterval);` both in this if block, so that we wouldn't miss either

amoro-ams/src/test/java/org/apache/amoro/server/scheduler/inline/TestTableRuntimeRefreshExecutor.java
```diff
@@ -0,0 +1,254 @@
...
+    // Test healthy table (not need optimizing) - interval should increase
+    DefaultTableRuntime mockTableRuntime =
+        getMockTableRuntimeWithAdaptiveInterval(tableRuntime, false, true, INTERVAL);
+
+    // Initial interval is INTERVAL, should increase by 30000
+    long expectedInterval = INTERVAL + STEP;
+    executor.execute(mockTableRuntime);
+
+    // Verify that setLatestRefreshInterval was called with the expected value
+    verify(mockTableRuntime, Mockito.times(1)).setLatestRefreshInterval(expectedInterval);
```
klion26: Instead using `verify` here to validate, is there any way we can validate the value of the `true interval` after `executor.execute` in L147.

The `true interval` is the value keeped in the table runtime.

amoro-ams/src/test/java/org/apache/amoro/server/scheduler/inline/TestTableRuntimeRefreshExecutor.java
```diff
@@ -0,0 +1,254 @@
...
+
+    // Test unhealthy table (need optimizing) - interval should decrease
+    mockTableRuntime =
+        getMockTableRuntimeWithAdaptiveInterval(tableRuntime, true, true, INTERVAL * 2);
```
klion26: Is there a way to reuse this `tableRuntime` so we can observe the change in `interval` before and after two consecutive `execute` calls? Reusing a single `tableRuntime` function would also be more suitable for production scenarios.

klion26 (commented):
Thanks for the contribution. I left some comments; please let me know what you think. thanks.

IMHO, this feature is valuable(especially when users use HMS as the catalog), as we can't use the `event-based` trigger as RestCatalog, but when there are many "static table" in the catalog, it will be a great burden for AMS to evaluate these tables, and changing the interval for every table is not a doable(as there are so many tables), and keep the default behavior as the same as before, would not add any inconvenience if we don't need it.

amoro-ams/src/test/java/org/apache/amoro/server/scheduler/inline/TestTableRuntimeRefreshExecutor.java
zhangwl9: fixup

docs/user-guides/configurations.md
zhangwl9: ok

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -88,7 +98,18 @@ private void tryEvaluatingPendingInput(DefaultTableRuntime tableRuntime, MixedTa
...
+
       tableRuntime.setTableSummary(evaluator.getPendingInput());
+      return evaluatorIsNecessary;
+    } else if (!optimizingEnabled) {
+      logger.debug(
+          "{} optimizing is not enabled, skip evaluating pending input",
+          tableRuntime.getTableIdentifier());
+      return false;
+    } else {
+      logger.debug(
```
klion26, line 112: Could we add some comments to describle the "semantics" of the return value, and why would this case would be `true`?

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -122,16 +143,133 @@ public void execute(TableRuntime tableRuntime) {
...
           || (mixedTable.isUnkeyedTable()
               && lastOptimizedSnapshotId != defaultTableRuntime.getCurrentSnapshotId())) {
-        tryEvaluatingPendingInput(defaultTableRuntime, mixedTable);
+        needOptimizing = tryEvaluatingPendingInput(defaultTableRuntime, mixedTable);
+      } else {
+        logger.debug("{} optimizing is not necessary", defaultTableRuntime.getTableIdentifier());
+      }
+
+      // Update adaptive interval according to evaluated result.
+      if (defaultTableRuntime.getOptimizingConfig().getRefreshTableAdaptiveMaxIntervalMs() > 0) {
```
klion26: Maybe we can add a function in `OptimizingConfig` which return `true/false` show whethere we enabled this feature

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
klion26: From below, the name is `needOptimizing`, looks like whethere we need to do an opitmizing, but in this branch, we don't need to optimizing in current loop? as there is an ongoing optimizing process

docs/user-guides/configurations.md
```diff
@@ -43,25 +43,27 @@ modified through [Alter Table](../using-tables/#modify-table) operations.
...
+| self-optimizing.min-plan-interval                       | 60000            | The minimum time interval between two self-optimizing planning action                                                                                                                                                                   |
+| self-optimizing.filter                                  | NULL             | Filter conditions for self-optimizing, using SQL conditional expressions, without supporting any functions. For the timestamp column condition, the ISO date-time formatter must be used. For example: op_time > '2007-12-03T10:15:30'. |
+| self-optimizing.refresh-table.adaptive.max-interval-ms  | -1               | The maximum time interval in milliseconds to refresh table metadata. The default value is -1, which disables the adaptive refresh.                                                                                                                |
```
klion26: Maybe we need to add some description here: after this has been enabled, the interval may be larger than `defaultTableRuntime.getOptimizingConfig().getMinorLeastInterval() * 4L / 5`

amoro-ams/src/test/java/org/apache/amoro/server/scheduler/inline/TestTableRuntimeRefreshExecutor.java
```diff
@@ -0,0 +1,204 @@
...
+
+    return newRuntime;
+  }
+
+  @Test
+  public void testAdaptiveRefreshIntervalScenarios() {
+    createDatabase();
+    createTable();
+
+    DefaultTableRuntime tableRuntime =
```
klion26: Could you please  reuse the _same_ table runtime, in the current test, we'll create every _new_ table runtime in `buildTableRuntimeWithAdaptiveRefresh`

klion26: If we need to update the `config` for an existing table runtime, maybe we can add a new class who extend `DefaultTableRuntime` and overrides the `getStore` function.

```
class TestTableRuntime extends DefaultTableRuntime {
   TableRuntimeStore testStore;
   TestTableRuntime(TableRuntimeStore store) {
       ...
       this.testStore = store
    }
    void updateRuntimeStore(TableRuntimeStore newStore) {
    ...
     this.testStore = newStore;
    }
    @Override
    TableConfiguration getStore() {
         this.testStore;
}
```

amoro-ams/src/test/java/org/apache/amoro/server/scheduler/inline/TestTableRuntimeRefreshExecutor.java
```diff
@@ -0,0 +1,204 @@
...
+    DefaultTableRuntime newTableRuntime =
+        buildTableRuntimeWithAdaptiveRefresh(tableRuntime, false, 0, STEP);
+    long adaptiveExecutingInterval = executor.getAdaptiveExecutingInterval(newTableRuntime);
+    long expectedInterval = INTERVAL + STEP;
+    Assert.assertEquals(expectedInterval, adaptiveExecutingInterval);
+
+    // Test minimum boundary - interval should not below INTERVAL
+    // The unhealthy table (need optimizing) - interval should decrease half
+    // The currentInterval is INTERVAL + STEP, the latest interval is INTERVAL not (INTERVAL + STEP)
+    // / 2
```
klion26: Place `(INTERVAL + STEP) / 2` in the same line is better than 2 lines

klion26 (commented):
Thanks for the update, I've left some comments, please take another look.

amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
```diff
@@ -318,6 +324,34 @@ public OptimizingConfig setEvaluationMseTolerance(long evaluationMseTolerance) {
     return this;
   }
 
+  public boolean isRefreshTableAdaptiveEnabled(long minInterval) {
```
klion26, line 333: Why do we need to pass in this parameter? will compare the `refreshTableAdaptiveMaxIntervalMs` to a default value in the current class be better?

The logic here seems is different than the doc below `The maximum time interval in milliseconds to refresh table metadata. 0 means disable adaptive refresh.`?

amoro-ams/src/main/java/org/apache/amoro/server/scheduler/inline/TableRuntimeRefreshExecutor.java
```diff
@@ -122,16 +151,148 @@ public void execute(TableRuntime tableRuntime) {
...
+  @VisibleForTesting
+  public long getAdaptiveExecutingInterval(DefaultTableRuntime tableRuntime) {
+    final long minInterval = interval;
+    final long maxInterval =
+        tableRuntime.getOptimizingConfig().getRefreshTableAdaptiveMaxIntervalMs();
+
+    if (maxInterval <= minInterval) {
+      tableRuntime
+          .getOptimizingConfig()
+          .setRefreshTableAdaptiveMaxIntervalMs(AmoroServiceConstants.INVALID_TIME);
```
klion26: Currently, we'll disable the adaptive refresh if `maxInterval <= minInterval`, but we call the function `getAdaptiveExecutingInterval` in `if(defaultTableRuntime.getOptimizingConfig().isRefreshTableAdaptiveEnabled())` block,
The logic here seems strange.

zhangwl9: The ‘setRefreshTableAdaptiveMaxIntervalMs(AmoroServiceConstants.INVALID_TIME)’ here is redundant and has been removed.

klion26 (approved):
@zhangwl9 Thanks for the contribution, LGTM, will merge this tomorrow if there are no more comments from the community.

PR #4043, opened 2026-01-15
Fixed the issue where the process table was completed after AMS restart, but the table status remained in the running state.

### LINKED ISSUE #4041: [Bug]: Restarting AMS may cause the table to remain in the running state indefinitely and never transition to another status.
### What happened?

After restarting the AMS pod, we found that some tables had actually completed the minor process, but their status remained as "minor" and would never change from this state.

Image

Image

### Affects Versions

0.8.0

Iceberg

Spark

### How to reproduce

When the table is in the committing state and the AMS pod is restarted, there is a chance of encountering this issue.

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

lintingbin:

Close #4041.

## Brief change log

- Initialize the TableOptimizingProcess to retrieve the Process Status from the database instead of using the default RUNNING status.
- In the initTableRuntime, determine whether the Process Status is successful. If successful, perform the complete process operation.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
@@ -141,9 +141,6 @@ public class OptimizingQueue extends PersistentBase {
               mapper -> mapper.getProcessState(tableRuntime.getProcessId()));
       process = new TableOptimizingProcess(tableRuntime, meta, state);
       tableRuntime.recover(process);
-      if (process.getStatus() == ProcessStatus.SUCCESS) {
-        tableRuntime.completeProcess(true);
-      }
     }
 
     if (tableRuntime.getOptimizingConfig().isEnabled()) {
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
@@ -32,6 +32,7 @@ import org.apache.amoro.optimizing.TableRuntimeOptimizingState;
 import org.apache.amoro.optimizing.plan.AbstractOptimizingEvaluator;
 import org.apache.amoro.process.AmoroProcess;
 import org.apache.amoro.process.ProcessFactory;
+import org.apache.amoro.process.ProcessStatus;
 import org.apache.amoro.process.TableProcessStore;
 import org.apache.amoro.server.AmoroServiceConstants;
 import org.apache.amoro.server.optimizing.OptimizingProcess;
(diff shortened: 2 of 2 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -141,6 +141,9 @@ private void initTableRuntime(DefaultTableRuntime tableRuntime) {
               mapper -> mapper.getProcessState(tableRuntime.getProcessId()));
       process = new TableOptimizingProcess(tableRuntime, meta, state);
       tableRuntime.recover(process);
+      if (process.getStatus() == ProcessStatus.SUCCESS) {
```
klion26: Will place this logic in `tableRuntime.recover(process)` be better?

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -500,6 +503,7 @@ public TableOptimizingProcess(
       if (processState.getToSequence() != null) {
         toSequence = processState.getToSequence();
       }
+      this.status = processMeta.getStatus();
```
klion26, line 503: Just curious, why do we need this?

lintingbin: The default status is currently ProcessStatus.RUNNING. During initialization, the correct state was not properly retrieved from the database. Here, we are only restoring the state from the database.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
lintingbin: That makes sense, I've made the changes

klion26 (approved):
Thanks for the quick update, LGTM, will merge it in next week if there are no more comments from the community.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
lintingbin, line 143: @klion26 Do you know why there is also tableRuntime.recover in the initialization function of TableOptimizingProcess? It looks a bit strange, doesn't it? Maybe we should optimize it?
```
      if (this.status != ProcessStatus.KILLED) {
        tableRuntime.recover(this);
      }
```

klion26: I don't have a strict preference for these two; these two are both ok to me, they are like different code styles to me.

klion26:
Thanks for the contribution again! @lintingbin

PR #4045, opened 2026-01-16
[AMORO-4034][Improvement]: MaintainerMetrics Unified interface and Report 

### LINKED ISSUE #4034: [Improvement]: MaintainerMetrics Unified interface and Report
### Search before asking

Currently, the maintainer has not yet perfected some metrics, and has only integrated the OrphanData-related indicators. We hope to expand the interfaces for other metrics.

### How should we improve?

- MaintainerMetrics interface is abstracted to support a multi-lake architecture
- Maintainer Metrics add other maintainer metrics interface: such as expireSnapshots/expireData/cleanDanglingDeleteFiles/autoCreateTags
- MaintainerMetrics OrphanDataFiles method Change to a more general implementation

### Subtasks

_No response_

### Code of Conduct

czy006:
[Improvement]: MaintainerMetrics Unified interface and Report  (#4034)

Close #4034 .

## Brief change log

- MaintainerMetrics interface is abstracted to support a multi-lake architecture
- Maintainer Metrics add other maintainer metrics interface: such as expireSnapshots/expireData/cleanDanglingDeleteFiles/autoCreateTags
- MaintainerMetrics OrphanDataFiles method Change to a more general implementation

## How was this patch tested?

## Documentation


amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/maintainer/IcebergTableMaintainer.java
```diff
@@ -273,26 +328,191 @@ public void expireDataFrom(DataExpirationConfig expirationConfig, Instant instan
     if (instant.equals(Instant.MIN)) {
       return;
     }
+    MaintainerMetrics metrics = context.getMetrics();
+    long startTime = System.currentTimeMillis();
+    expireDataFromWithMetrics(expirationConfig, instant, metrics, startTime);
+  }
 
+  private void expireDataFromWithMetrics(
```
majin1102: I think `expireDataFrom` is just OK. It's a little werid to me to use `FromWith`

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/maintainer/IcebergTableMaintainer.java
```diff
@@ -273,26 +328,191 @@ public void expireDataFrom(DataExpirationConfig expirationConfig, Instant instan
...
+      DataExpirationConfig expirationConfig,
+      Instant instant,
+      MaintainerMetrics metrics,
+      long startTime) {
```
majin1102: I think we might need
```
    if (instant.equals(Instant.MIN)) {
      return;
    }
```

What do you think

amoro-ams/src/main/java/org/apache/amoro/server/table/AbstractTableMaintainerMetrics.java
```diff
@@ -0,0 +1,229 @@
...
+ *
+ * <ul>
+ *   <li>Define all MetricDefine constants
+ *   <li>Provide template methods for metrics registration
+ *   <li>Handle tags (catalog, database, table, table_format, operation_type)
+ * </ul>
+ *
+ * <p>Note: All metrics include the table_format tag to distinguish Iceberg and Paimon tables
+ */
+public abstract class AbstractTableMaintainerMetrics extends AbstractTableMetrics
```
majin1102: Wondering this abstraction is really necessary if we put that format field into `AbstractTableMetrics`. Could you elaborate more on this?

amoro-ams/src/main/java/org/apache/amoro/server/table/AbstractTableMaintainerMetrics.java
```diff
@@ -0,0 +1,229 @@
...
+  protected static final String TABLE_FORMAT_HIVE = "hive";
+
+  // ========== Orphan Files Related MetricDefine ==========
+
+  /** Count of orphan data files cleaned */
+  public static final MetricDefine TABLE_ORPHAN_DATA_FILES_CLEANED_COUNT =
+      defineCounter("table_orphan_data_files_cleaned_count")
+          .withDescription("Count of orphan data files cleaned")
+          .withTags("catalog", "database", "table", "table_format")
+          .build();
```
majin1102: This seems a break change from `table_orphan_content_file_cleaning_count`. This might lead to backforward compatibility issue

amoro-common/src/main/java/org/apache/amoro/maintainer/MaintainerOperationTemplate.java
```diff
@@ -0,0 +1,151 @@
...
+      metrics.recordOperationFailure(operationType, duration, t);
+      throw t;
+    }
+  }
+
+  /** Functional interface for maintainer operation without return value. */
+  @FunctionalInterface
+  public interface MaintainerOperation {
+    /** Executes the operation. */
+    void execute() throws Throwable;
```
majin1102: I think It's a little strange to declare as `throws Throwable`.
Does any project declare like this?

And I think we could just use some interfaces JDK provides, like Runnable?

amoro-common/src/main/java/org/apache/amoro/maintainer/MaintainerOperationTemplate.java
```diff
@@ -0,0 +1,151 @@
...
+ * template.execute(
+ *     MaintainerOperationType.ORPHAN_FILES_CLEANING,
+ *     () -> {
+ *       // Operation logic here
+ *       cleanOrphanFiles();
+ *     }
+ * );
+ * }</pre>
+ */
+public class MaintainerOperationTemplate {
```
majin1102: This class looks more like an executor that wraps maintainer operations with metrics, rather than a classic template-method base class. To make the intent clearer, I suggest renaming it to `MaintainerOperationExecutor` (or MaintainerMetricsExecutor, MaintainerOperationRunner), which better reflects that it mainly wires metrics around the operation execution.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/maintainer/IcebergTableMaintainer.java
```diff
@@ -165,7 +168,17 @@ public void cleanDanglingDeleteFiles() {
         Optional.ofNullable(currentSnapshot.summary().get(SnapshotSummary.TOTAL_DELETE_FILES_PROP));
     if (totalDeleteFiles.isPresent() && Long.parseLong(totalDeleteFiles.get()) > 0) {
       // clear dangling delete files
-      doCleanDanglingDeleteFiles();
+      LOG.info("Starting cleaning dangling delete files for table {}", table.name());
```
majin1102: So why we don't use the new interfaces and wireness?

amoro-common/src/main/java/org/apache/amoro/maintainer/MaintainerOperationTemplate.java
```diff
@@ -0,0 +1,151 @@
...
+   *   <li>On failure: record operation failure via {@link MaintainerMetrics#recordOperationFailure}
+   *       and rethrow the exception
+   * </ol>
+   *
+   * @param operationType the type of operation being executed
+   * @param operation the operation to execute
+   * @return true if operation succeeded, false otherwise
+   * @throws Throwable if the operation throws an exception
+   */
+  public boolean execute(MaintainerOperationType operationType, MaintainerOperation operation)
```
majin1102: This should be a void

amoro-common/src/main/java/org/apache/amoro/maintainer/MaintainerOperationTemplate.java
```diff
@@ -0,0 +1,151 @@
...
+    void execute() throws Throwable;
+  }
+
+  /**
+   * Functional interface for maintainer operation with return value.
+   *
+   * @param <T> the result type
+   */
+  @FunctionalInterface
+  public interface MaintainerOperationWithResult<T> {
```
majin1102: I think we could make this more reasonable. like MaintainerCallable<T> and MaintainerRunner

majin1102 (commented):
Thanks for working on this. Left some comments

amoro-ams/src/main/java/org/apache/amoro/server/table/AbstractTableMaintainerMetrics.java
czy006: I'm worried that on different lake formats, there might be some unique indicators of their own, so it's abstract. Although we have added the TableFormat identifier inside, this is only for monitoring and distinction. In terms of code, we cannot isolate and distinguish it.

czy006:
cc @baiyangtx @xxubai

czy006:
@j1wonpark Can you help me review code ? This is an improvement to the monitoring abstraction of a multi-lake format.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/maintainer/IcebergTableMaintainer.java
```diff
@@ -139,107 +141,163 @@ public void cleanOrphanFiles() {
...
+  }
+
+  @VisibleForTesting
+  public void expireSnapshots(long mustOlderThan, int minCount) {
+    MaintainerMetrics metrics = context.getMetrics();
+    expireSnapshotsInternal(mustOlderThan, minCount, metrics);
+  }
+
+  private void expireSnapshots(long olderThan, int minCount, Set<String> exclude) {
+    // This method is kept for backward compatibility with potential subclasses
```
j1wonpark: This private method is no longer called after the refactor (2-param `expireSnapshots` now calls `expireSnapshotsInternal` directly). The "backward compatibility with subclasses" comment doesn't apply since it's `private`. Can be removed.

amoro-ams/src/main/java/org/apache/amoro/server/table/TableMaintainerMetrics.java
```diff
@@ -0,0 +1,531 @@
...
+    public Long getValue() {
+      return value;
+    }
+  }
+
+  /** Gauge implementation for recording operation duration (supports recording start time) */
+  private static class OperationDurationGauge implements Gauge<Long> {
+    private volatile long value = 0L;
+    private volatile long startTime = 0L;
+
```
j1wonpark: `startTime` is written in `recordStart()` but never read — `getValue()` only returns `value`. Looks like a leftover from an earlier design. Can be removed along with `recordStart()`.

j1wonpark:
`autoCreateTags()` logic is now inlined into `IcebergTableMaintainer`, but `AutoCreateIcebergTagAction` class and its test `TestAutoCreateIcebergTagAction` are not deleted in this PR. They become dead code after this change. Should be removed.

Also left a design suggestion about this in a separate comment below.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/maintainer/IcebergTableMaintainer.java
```diff
@@ -139,107 +141,163 @@ public void cleanOrphanFiles() {
...
+    executor.execute(
+        MaintainerOperationType.SNAPSHOT_EXPIRATION,
+        () -> expireSnapshotsInternal(mustOlderThan, minCount, metrics));
   }
 
-  private void expireSnapshots(long olderThan, int minCount, Set<String> exclude) {
+  private void expireSnapshotsInternal(
+      long mustOlderThan, int minCount, MaintainerMetrics metrics) {
+    long startTime = System.currentTimeMillis();
+    Set<String> exclude = expireSnapshotNeedToExcludeFiles();
```
j1wonpark, line 220: Duration is recorded twice for the same operation — once by `MaintainerOperationExecutor` (`table_maintainer_operation_duration_millis`) and once inside this method (`table_snapshots_expiration_duration_millis`). The two values will differ slightly. Consider using a single source of truth for duration.

Same pattern applies to `doExpireData` and `autoCreateTags`.

j1wonpark:
Thanks for the contribution! I left a few inline comments on dead code and duration recording.

One design suggestion: the tag creation logic from `AutoCreateIcebergTagAction` is now inlined into `IcebergTableMaintainer`, which makes the class larger, leaves the original class as dead code, and couples tag creation logic with the maintainer. It might be cleaner to keep `AutoCreateIcebergTagAction` and extend it to support metrics, keeping the responsibilities decoupled.

PR #4053, opened 2026-01-20
[AMORO-3226] Provider a better table explorer view

majin1102:

Close #3226.

## Brief change log

The new UI looks like:
image

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-web/src/views/tables/components/TableExplorer.vue
+++ b/amoro-web/src/views/tables/components/TableExplorer.vue
@@ -234,13 +234,15 @@ function handleTreeSelect(selectedKeys: (string | number)[], info: any) {
     return
   }
 
-  const dataRef = node.dataRef || node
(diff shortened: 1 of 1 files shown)


zhoujinsong:
Clipboard_Screenshot_1768965163

I found that when I don't select any table, the table details page still displays, but encounters errors. Should we provide an empty page here to avoid such errors?

majin1102:
Two optimization suggested from @zhoujinsong and @xxubai 
1. We can now click anywhere on a row to expand or collapse a catalog/database
2. If you don’t select a table, the content on the right stays empty.

Check that out when you have time!

zhoujinsong:
Thanks a lot for the improvement.

I found that when I select an empty database now, the table details page still displays, but it encounters errors. 

Clipboard_Screenshot_1769000134

majin1102:
> I found that when I select an empty database now, the table details page still displays, but it encounters errors.

Check it now

PR #4069, opened 2026-01-29
[HotFix] Fix NoClassDefFoundError for Guava ImmutableMap during Hive-based Iceberg catalog initialization

j1wonpark:

When initializing a Hive Metastore-based Iceberg catalog in amoro-ams, a `NoClassDefFoundError: com/google/common/collect/ImmutableMap` is thrown at runtime.

Hive 3.1.3's `HiveConf` directly references the original Guava class `com.google.common.collect.ImmutableMap`, but the Guava dependency in `amoro-ams/pom.xml` is declared with `<scope>test</scope>`, so it is excluded from the runtime classpath. All other Guava artifacts in the project are shaded (e.g., `amoro-shade-guava-32`, `iceberg-bundled-guava`, `flink-shaded-guava`) and do not provide classes under the original `com.google.common.*` package. Additionally, the Hive dependency explicitly excludes Guava via `<exclusion>`, preventing transitive resolution.

This issue also affects the official distribution and Docker images.

## Brief change log

- Removed `<scope>test</scope>` from the Guava dependency in `amoro-ams/pom.xml` so that it defaults to `compile` scope and is included in the runtime classpath.

## How was this patch tested?

- Verified that amoro-ams starts successfully and initializes a Hive Metastore-based Iceberg catalog without `NoClassDefFoundError`.

## Documentation

### DIFF
--- a/amoro-ams/pom.xml
+++ b/amoro-ams/pom.xml
@@ -509,7 +509,7 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>${guava.version}</version>
+            <version>${guava-hive.version}</version>
         </dependency>
 
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -159,6 +159,7 @@
         <amoro-shade-thrift.version>0.20.0</amoro-shade-thrift.version>
         <annotation-api.version>1.3.2</annotation-api.version>
         <guava.version>32.1.1-jre</guava.version>
+        <guava-hive.version>32.1.1-jre</guava-hive.version>
         <hudi.version>0.14.1</hudi.version>
         <pagehelper.version>6.1.0</pagehelper.version>
         <jsqlparser.version>4.7</jsqlparser.version>


amoro-ams/pom.xml
```diff
@@ -510,7 +510,6 @@
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
             <version>${guava.version}</version>
```
czy006: I'm encountering the same issue in the Paimon Hive Catalog. This creates a paradox with the current shared packages. Is there a better solution?

j1wonpark: Hi @czy006, thanks for reporting the Paimon Hive Catalog issue — that's very helpful!

This fix should actually resolve the Paimon case as well. Let me explain why.

I decompiled the `paimon-bundle` JAR and traced the runtime call path:

1. **Paimon's own Guava is fully shaded** — relocated to `org.apache.paimon.shade.guava30.com.google.common.*`, so Paimon's internal code has no dependency on unshaded Guava.

2. **However, Paimon's `HiveCatalog` references unshaded `org.apache.hadoop.hive.conf.HiveConf` directly:**
   ```
   private final org.apache.hadoop.hive.conf.HiveConf hiveConf;
   ```
   `HiveConf` is **not** bundled inside `paimon-bundle` — it's resolved from the amoro-ams classpath at runtime.

3. **`HiveConf` internally uses unshaded Guava** (e.g., `com.google.common.collect.ImmutableMap`), which is exactly the same root cause as the Iceberg case.

By promoting Guava from `test` to `compile` scope in `amoro-ams`, the unshaded Guava classes become available on the runtime classpath, resolving the issue for **both** Iceberg and Paimon Hive Catalogs.

Regarding the shared package concern — there is no actual conflict:

- **Amoro's own code** → `org.apache.amoro.shade.guava32.com.google.*` (relocated)
- **Paimon's internal code** → `org.apache.paimon.shade.guava30.com.google.*` (relocated)
- **Hive / Iceberg** → `com.google.common.*` (unshaded, provided by this fix)
- All three namespaces coexist without collision.
- All transitive Guava dependencies are already excluded via `<exclusion>`, so only a single unshaded Guava version (`${guava.version}`) exists on the classpath.

That said, if you are experiencing a specific error that this fix does not resolve, could you share the details (stack trace, environment, etc.)? I'd be happy to investigate further and work on a solution together.

Thanks again for the feedback! 🙏

j1wonpark: Hi @czy006, thanks for the feedback!

This fix resolves the Paimon Hive Catalog case as well. The root cause is the same — Paimon's `HiveCatalog` uses unshaded `HiveConf` from the amoro-ams classpath, and `HiveConf` internally depends on unshaded Guava (`ImmutableMap`).

There's no shared package conflict:
- **Amoro** uses `org.apache.amoro.shade.guava32.*`
- **Paimon** uses `org.apache.paimon.shade.guava30.*`
- **Hive/Iceberg** uses `com.google.common.*` (provided by this fix)

All three are separate namespaces — no collision.

If you're seeing a specific error that this fix doesn't cover, could you share the stack trace? Happy to dig into it together.

czy006: This is a terrible dependency disaster, but there seems to be no better way to resolve package sharing conflicts. I suggest that this guava be version name to **${guava-hive.version}** as it is related to the version of hive guava cc @j1wonpark

j1wonpark: Thanks @czy006! Updated the version property to `${guava-hive.version}` as suggested.

czy006:
@j1wonpark You need to rebase this PR, as your commits include content from previous PRs

j1wonpark:
@czy006 Done! Thanks

PR #4073, opened 2026-02-02
[Refactor] Introduce Iceberg TableRuntimePlugin

majin1102:

Refactoring work for format processing extensions

## Brief change log

1. Abstract TableRuntimePlugin to extend lance table runtime processing
2. Refactor DefaultTableService to abstract a IcebergTablePlugin which implements TableRuntimeHandler(only aimed for Iceberg table)

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
@@ -54,7 +54,7 @@ import org.apache.amoro.server.resource.OptimizerManager;
 import org.apache.amoro.server.scheduler.inline.InlineTableExecutors;
 import org.apache.amoro.server.table.DefaultTableManager;
 import org.apache.amoro.server.table.DefaultTableService;
-import org.apache.amoro.server.table.IcebergTablePlugin;
+import org.apache.amoro.server.table.IcebergBasedPlugin;
 import org.apache.amoro.server.table.TableManager;
 import org.apache.amoro.server.table.TableRuntimeFactoryManager;
 import org.apache.amoro.server.table.TableRuntimePlugin;
@@ -251,8 +251,8 @@ public class AmoroServiceContainer {
   private List<TableRuntimePlugin> initTablePlugins() {
     LOG.info("Setting up Iceberg table runtime plugins...");
     InlineTableExecutors.getInstance().setup(tableService, serviceConfig);
-    IcebergTablePlugin icebergTablePlugin =
-        IcebergTablePlugin.builder()
+    IcebergBasedPlugin icebergBasedPlugin =
+        IcebergBasedPlugin.builder()
             .addHandler(optimizingService.getTableRuntimeHandler())
             .addHandler(processService.getTableHandlerChain())
             .addHandler(InlineTableExecutors.getInstance().getDataExpiringExecutor())
@@ -266,7 +266,7 @@ public class AmoroServiceContainer {
             .addHandler(InlineTableExecutors.getInstance().getTableRefreshingExecutor())
             .addHandler(InlineTableExecutors.getInstance().getTagsAutoCreatingExecutor())
             .build();
-    return List.of(icebergTablePlugin);
+    return List.of(icebergBasedPlugin);
   }
 
   public void disposeOptimizingService() {
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/IcebergBasedPlugin.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/IcebergBasedPlugin.java
@@ -32,11 +32,11 @@ import javax.annotation.Nullable;
 
 import java.util.List;
 
-public class IcebergTablePlugin implements TableRuntimePlugin, TableRuntimeHandler {
+public class IcebergBasedPlugin implements TableRuntimePlugin, TableRuntimeHandler {
 
   private final RuntimeHandlerChain headHandler;
 
-  private IcebergTablePlugin(RuntimeHandlerChain headHandler) {
+  private IcebergBasedPlugin(RuntimeHandlerChain headHandler) {
     Preconditions.checkNotNull(headHandler);
     this.headHandler = headHandler;
   }
@@ -110,8 +110,8 @@ public class IcebergTablePlugin implements TableRuntimePlugin, TableRuntimeHandl
       return this;
     }
 
-    public IcebergTablePlugin build() {
-      return new IcebergTablePlugin(headHandler);
+    public IcebergBasedPlugin build() {
+      return new IcebergBasedPlugin(headHandler);
     }
   }
 }
--- a/amoro-ams/src/test/java/org/apache/amoro/server/AMSServiceTestBase.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/AMSServiceTestBase.java
@@ -26,7 +26,7 @@ import org.apache.amoro.server.process.ProcessService;
 import org.apache.amoro.server.table.DefaultTableRuntime;
 import org.apache.amoro.server.table.DefaultTableRuntimeFactory;
 import org.apache.amoro.server.table.DefaultTableService;
-import org.apache.amoro.server.table.IcebergTablePlugin;
+import org.apache.amoro.server.table.IcebergBasedPlugin;
 import org.apache.amoro.server.table.TableRuntimeFactoryManager;
 import org.apache.amoro.shade.guava32.com.google.common.collect.Lists;
 import org.junit.AfterClass;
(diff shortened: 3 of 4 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
```diff
@@ -239,33 +240,33 @@ public void startOptimizingService() throws Exception {
...
     initThriftService();
     startThriftService();
   }
 
-  private void addHandlerChain(RuntimeHandlerChain chain) {
-    if (chain != null) {
-      tableService.addHandlerChain(chain);
-    }
+  private List<TableRuntimePlugin> initTablePlugins() {
+    LOG.info("Setting up AMS table executors...");
```
czy006: Setting up AMS Table Plugins ...

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
```diff
@@ -239,33 +240,33 @@ public void startOptimizingService() throws Exception {
...
+    InlineTableExecutors.getInstance().setup(tableService, serviceConfig);
+    IcebergTablePlugin icebergTablePlugin =
```
xxubai: How to load different table runtime plugins?

amoro-ams/src/main/java/org/apache/amoro/server/table/IcebergBasedPlugin.java
```diff
@@ -0,0 +1,127 @@
...
+  private final RuntimeHandlerChain headHandler;
+
+  private IcebergTablePlugin(RuntimeHandlerChain headHandler) {
+    Preconditions.checkNotNull(headHandler);
+    this.headHandler = headHandler;
+  }
+
+  @Override
+  public boolean accept(ServerTableIdentifier tableIdentifier) {
+    return tableIdentifier.getFormat() == TableFormat.ICEBERG
```
xxubai, line 46: IcebergTablePlugin will accept other formats? It should not be named as `IcebergTablePlugin`

amoro-ams/src/main/java/org/apache/amoro/server/table/IcebergBasedPlugin.java
```diff
@@ -0,0 +1,127 @@
...
+      throw new IllegalStateException("Only support DefaultTableRuntimeStore");
+    }
+    DefaultTableRuntimeStore icebergRuntimeStore = (DefaultTableRuntimeStore) store;
+    icebergRuntimeStore.setRuntimeHandler(this);
+    return creator.create(store);
+  }
+
+  @Override
+  public void initialize(List<TableRuntime> tableRuntimes) {
+    if (headHandler != null) {
```
xxubai, line 64: Already checked null in construct function

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
```diff
@@ -182,23 +136,46 @@ public void initialize() {
...
+              () -> LOG.warn("No available table runtime factory found for table {}", identifier));
     }
+  }
 
-    if (headHandler != null) {
-      headHandler.initialize(tableRuntimes);
-    }
+  private void initTableRuntimePlugins() {
+    List<TableRuntime> tableRuntimes = new ArrayList<>(tableRuntimeMap.values());
+    tableRuntimePlugins.forEach(plugin -> plugin.initialize(tableRuntimes));
```
xxubai, line 149: Plugin is invoked in `initTableRuntimes` when create `tableRuntime`. Should we check the method workflow?

amoro-ams/src/main/java/org/apache/amoro/server/table/IcebergBasedPlugin.java
czy006: I think TableFormatPlugin should not be implemented in the current PR. The current PR is for renaming and organizing, which I can provide later

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
majin1102: nice catch

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
majin1102: I think we should still use SPI(PluginManager) to do this. I'm preparing to create a LanceTableRuntimePlugin to handle lance tables and we could see the path clearly.

But Iceberg table runtime plugin is automatically loaded.
How do you feel

amoro-ams/src/main/java/org/apache/amoro/server/table/IcebergBasedPlugin.java
majin1102: > IcebergTablePlugin will accept other formats? It should not be named as `IcebergTablePlugin`

I name this as IcebergTablePlugin because the current three are all based on Iceberg. I'm open to have another name. Do you have another idea.

amoro-ams/src/main/java/org/apache/amoro/server/table/IcebergBasedPlugin.java
majin1102: You are right

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
majin1102: Plugin is initialized in this `initTableRuntimePlugins`, not creating table runtime? what do you mean by `check the method workflow`?

amoro-ams/src/main/java/org/apache/amoro/server/table/IcebergBasedPlugin.java
xxubai: How about `IcebergLikeTablePlugin`

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableService.java
xxubai: I'm not sure if it occur NPE when invoke plugin functions

amoro-ams/src/main/java/org/apache/amoro/server/table/TableRuntimePlugin.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.List;
+
+public interface TableRuntimePlugin {
+
+  boolean accept(ServerTableIdentifier tableIdentifier);
+
+  void initialize(List<TableRuntime> tableRuntimes);
+
+  void onTableCreated(@Nullable AmoroTable<?> amoroTable, TableRuntime tableRuntime);
```
xxubai, line 37: The interface is named TableXXX. Should we simplify the method names, for example to onCreate?

amoro-ams/src/main/java/org/apache/amoro/server/table/TableRuntimePlugin.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.amoro.ServerTableIdentifier;
+import org.apache.amoro.TableRuntime;
+import org.apache.amoro.table.TableRuntimeFactory;
+import org.apache.amoro.table.TableRuntimeStore;
+
+import javax.annotation.Nullable;
+
+import java.util.List;
+
+public interface TableRuntimePlugin {
```
xxubai, line 31: **1. TableRuntimePlugin violates Single Responsibility Principle**
This interface mixes two distinct concerns:

- Factory/decorator role: accept() + createTableRuntime() intercepts and decorates TableRuntime creation.
- Lifecycle listener role: initialize() / onTableCreated() / onTableDropped() / dispose() observes table lifecycle events.

These should be separated. The factory/decorator logic belongs in the TableRuntimeFactory hierarchy (e.g. as a decorator pattern), while the lifecycle observation should be its own interface. Mixing them forces every plugin implementor to deal with both concerns even if they only care about one.

**2. Semantic overlap between TableRuntimePlugin.accept() and TableRuntimeFactory.accept()**
In DefaultTableService.createTableRuntime(), there is a two-phase accept chain: first TableRuntimeFactory.accept(identifier, properties) selects the factory, then TableRuntimePlugin.accept(identifier) selects the plugin to intercept creation. These two have overlapping semantics — both decide "does this component handle this table?" — but with inconsistent signatures (TableRuntimeFactory receives table properties, TableRuntimePlugin does not). This dual-accept flow is confusing and error-prone. Consider unifying the entry point so there is a single, clear decision path for table runtime creation.

PR #4074, opened 2026-02-05
[Hotfix] Fix uninitialized OPTIMIZER_LOG_CONF_FILE causing Log4j parsing error

j1wonpark:

When starting the optimizer, Log4j throws a parsing error:
```
ERROR StatusLogger Error parsing /usr/local/amoro/conf/optimize
org.xml.sax.SAXParseException; Content is not allowed in prolog.
```

The root cause is that `OPTIMIZER_LOG_CONF_FILE` variable is used in `optimizer.sh` but never initialized. When the variable is empty, Log4j falls back to CLASSPATH scanning and attempts to parse the `conf/optimize` directory as an XML file, causing the SAXParseException.

## Brief change log

- Initialize `OPTIMIZER_LOG_CONF_FILE` variable in `dist/src/main/amoro-bin/bin/optimizer.sh` to point to `${AMORO_CONF_DIR}/optimize/log4j2.xml`

## How was this patch tested?

## Documentation

### DIFF
--- a/dist/src/main/amoro-bin/bin/optimizer.sh
+++ b/dist/src/main/amoro-bin/bin/optimizer.sh
@@ -21,6 +21,10 @@ CURRENT_DIR="$( cd "$(dirname "$0")" ; pwd -P )"
 
 . $CURRENT_DIR/load-config.sh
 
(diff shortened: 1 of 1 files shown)


czy006 (approved):
LGTM

PR #4076, opened 2026-02-05
[Hotfix] Fix optimizer group not updated on table refresh

j1wonpark:

When the `self-optimizing.group` property of a table is changed (either directly or through catalog default settings), the `group_name` column in the `table_runtime` database table is not updated during table refresh. This causes:

1. Tables remain bound to the old optimizer group even after configuration changes
2. `handleConfigChanged()` uses stale `group_name` from DB for queue binding
3. Tables get stuck in `pending` status permanently when the original optimizer group no longer exists

**Root Cause:** `DefaultTableRuntime.refresh()` method was missing `updateGroup()` call in the store transaction.

## Brief change log

- **DefaultTableRuntime.java**: Added `updateGroup(g -> newGroupName)` call in `refresh()` method to update the `group_name` column in DB when optimizer group changes
- **TestDefaultTableRuntimeHandler.java**: Added `testRefreshUpdatesOptimizerGroup()` test to verify that `getGroupName()` returns the new group name after refresh

## How was this patch tested?

## Documentation

- Does this pull request introduce a new feature? **no**
- If yes, how is the feature documented? **not applicable**

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
@@ -309,8 +309,6 @@ public class DefaultTableRuntime extends AbstractTableRuntime
   public DefaultTableRuntime refresh(AmoroTable<?> table) {
     Map<String, String> tableConfig = table.properties();
     TableConfiguration newConfiguration = TableConfigurations.parseTableConfig(tableConfig);
-    TableConfiguration oldConfiguration = getTableConfiguration();
-    boolean configChanged = !newConfiguration.equals(oldConfiguration);
     String newGroupName = newConfiguration.getOptimizingConfig().getOptimizerGroup();
 
     if (!Objects.equals(getGroupName(), newGroupName)) {


amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
```diff
@@ -311,9 +311,9 @@ public DefaultTableRuntime refresh(AmoroTable<?> table) {
     TableConfiguration newConfiguration = TableConfigurations.parseTableConfig(tableConfig);
     TableConfiguration oldConfiguration = getTableConfiguration();
     boolean configChanged = !newConfiguration.equals(oldConfiguration);
```
klion26: Seems the `configChanged` has never been used, maybe we can return immediately if configChanged is false?

j1wonpark: Removed the unused `configChanged` variable and `oldConfiguration` in the latest commit. Thanks for the review!

PR #4083, opened 2026-02-11
[Hotfix] Fix incorrect Last Commit Time showing year 58000+ in partition view

j1wonpark:

The "Last Commit Time" column in the table partition view displays incorrect dates (year ~58000) instead of the actual commit time.

**Root cause:** The `last_updated_at` column from Iceberg's `PARTITIONS` metadata table stores timestamps in **microseconds** since epoch, but the value was passed directly to the frontend which expects **milliseconds**. This 1000x difference causes `new Date()` to interpret the value as a far-future date.

For example:
- Actual microsecond value: `1,770,000,000,000,000`
- Frontend interprets as milliseconds → year ~58,060

Other code paths (e.g., `snapshot.timestampMillis()` used in file list and fallback partition scan) correctly return milliseconds, so only the `PARTITIONS` metadata table path was affected.

## Brief change log

- `MixedAndIcebergTableDescriptor#collectPartitionsFromTable`: Divide `last_updated_at` by 1000 to convert microseconds to milliseconds before setting `lastCommitTime`
- `TestIcebergServerTableDescriptor#testGetTablePartitions`: Add assertion to verify `lastCommitTime` is in milliseconds range (below year 2100)

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

PR #4084, opened 2026-02-12
[Feature] Add metric category filtering support to Prometheus exporter

j1wonpark:

Currently, the Prometheus exporter exposes all registered metrics (~61 metrics) unconditionally. As the number of tables grows, the volume of metrics increases significantly. Users need the ability to selectively filter metrics to reduce noise and focus on what matters to their use case.

## Brief change log

- Added `MetricFilter` class in `amoro-metrics-prometheus` module with regex-based includes/excludes pattern matching
- Modified `MetricsCollector` to accept a `MetricFilter` and apply it in `isValidMetric()`
- Modified `PrometheusMetricsReporter` to parse filter settings from `properties` in `open()` and pass them to `MetricsCollector`
- Updated `metric-reporters.yaml` with regex filtering configuration examples
- Updated Helm chart `values.yaml` and `amoro-configmap_test.yaml` to reflect the new configuration options

### Configuration example:
```yaml
metric-reporters:
  - name: prometheus-exporter
    enabled: true
    properties:
       port: 7001
       # Include only specific metric patterns (optional, all included by default)
       metric-filter.includes: "table_optimizing_.*|optimizer_group_.*"
       # Exclude specific metric patterns (optional, excludes take precedence over includes)
       metric-filter.excludes: "table_summary_.*"
```

All metrics are included by default for backward compatibility. Users can use `metric-filter.includes` to whitelist, `metric-filter.excludes` to blacklist, or both (excludes take precedence).

## How was this patch tested?

  - `MetricFilterTest` (7 tests): accept-all, includes-only, excludes-only, includes+excludes, excludes precedence, exact match, empty properties
  - `MetricsCollectorRegexFilterTest` (5 tests): no filter, includes filter, excludes filter, includes+excludes, accept-all filter
  - `./mvnw test -pl amoro-metrics/amoro-metrics-prometheus -am` — 12 tests passed
  - `helm unittest charts/amoro` — 75 tests passed

## Documentation

- If yes, how is the feature documented? docs (metric-reporters.yaml comments and Helm values.yaml comments)


zhoujinsong:
@j1wonpark Thanks a lot for the proposal.

I think it is useful to support filtering some metrics. 
However, I'm wondering if we could open up a regular expression to allow users to freely configure the metrics they want to filter.

j1wonpark:
@zhoujinsong Thanks for the feedback! Regex support sounds like a great idea.
I’ll update it so users can configure their own filtering patterns. Will get it done soon!

amoro-metrics/amoro-metrics-prometheus/src/main/java/org/apache/amoro/metrics/promethues/MetricFilter.java
```diff
@@ -0,0 +1,64 @@
...
+  public static MetricFilter fromProperties(Map<String, String> properties) {
+    String includes = properties.get(INCLUDES_KEY);
+    String excludes = properties.get(EXCLUDES_KEY);
+
+    if (includes == null && excludes == null) {
+      return ACCEPT_ALL;
+    }
+
+    Pattern includePattern = includes != null ? Pattern.compile(includes) : null;
+    Pattern excludePattern = excludes != null ? Pattern.compile(excludes) : null;
```
xxubai: Pattern.compile(includes/excludes) throws a `PatternSyntaxException` when the regex is invalid, which fails reporter initialization. We should catch the exception, log a clear error message, and fall back to ACCEPT_ALL.

charts/amoro/tests/amoro-configmap_test.yaml
```diff
@@ -151,4 +151,18 @@ tests:
...
+  - it: Amoro configMap should reflect metric regex filtering
+    set:
+      plugin:
+        metricReporters:
+          prometheusExporter:
+            name: prometheus-exporter
+            enabled: true
+            properties:
+              port: 7001
+              metric-filter.excludes: "table_summary_.*"
```
xxubai, line 164: add test for `metric-filter.includes`

xxubai (commented):
Thanks for the contribution. I left some comments

j1wonpark:
@xxubai Thanks for the review! I've addressed both comments in the latest commit (6cffced):

1. **Invalid regex fallback**: `MetricFilter.fromProperties()` now catches `PatternSyntaxException`, logs a warning with the invalid pattern details, and falls back to `ACCEPT_ALL` instead of failing reporter initialization. Added two test cases for this.

2. **Helm test for `metric-filter.includes`**: Added a new test case that verifies the includes configuration is reflected in the configMap. Also fixed the missing newline at end of file.

xxubai (approved):
LGTM

PR #4097, opened 2026-02-18
[Improvement]: Refactor process API and table runtime extension interfaces

baiyangtx:

This MR introduces the new process extension APIs in common module, without touching most AMS internals yet.

## Brief change log

- Promote `ProcessFactory` to a richer abstraction in `amoro-common`, preparing for plugin-based table process extension.
- Adjust `Action`/`IcebergActions` and `TableRuntime` to align with the new process model.
- Move `ActionCoordinator` into `amoro-common` so it can be shared as a public abstraction.
- Introduce `ProcessTriggerStrategy` to describe trigger policies for processes.
- Clean up legacy process state classes which will be replaced by the new abstractions in follow-up MRs.

**Note** 

- This commit only updates the common module and the shared `ActionCoordinator` API; AMS-side wiring and runtime refactors will be done in separate branches/MRs as discussed.

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-common/src/main/java/org/apache/amoro/TableRuntime.java
+++ b/amoro-common/src/main/java/org/apache/amoro/TableRuntime.java
@@ -49,10 +49,6 @@ public interface TableRuntime {
    */
   List<? extends TableProcessStore> getProcessStates(Action action);
 
-  void registerProcess(TableProcessStore processStore);
-
-  void removeProcess(TableProcessStore processStore);
-
   /** Get the group name of the table runtime. */
   String getGroupName();
 
--- a/amoro-common/src/main/java/org/apache/amoro/process/ProcessFactory.java
+++ b/amoro-common/src/main/java/org/apache/amoro/process/ProcessFactory.java
@@ -45,7 +45,9 @@ public interface ProcessFactory extends ActivePlugin {
   Map<TableFormat, Set<Action>> supportedActions();
 
   /** How to trigger a process for the action. */
-  ProcessTriggerStrategy triggerStrategy(TableFormat format, Action action);
+  default ProcessTriggerStrategy triggerStrategy(TableFormat format, Action action) {
+    return ProcessTriggerStrategy.METADATA_TRIGGER;
+  }
 
   /**
    * Try trigger a process for the action.
@@ -60,8 +62,10 @@ public interface ProcessFactory extends ActivePlugin {
    * Recover a process for the action from a state.
    *
    * @param tableRuntime table runtime
-   * @param state state of the process
+   * @param store storage of the process
    * @return target process which has not been submitted yet.
+   * @throws RecoverProcessFailedException if recover failed
    */
-  TableProcess recover(TableRuntime tableRuntime, TableProcessStore store);
+  TableProcess recover(TableRuntime tableRuntime, TableProcessStore store)
+      throws RecoverProcessFailedException;
 }
--- a/amoro-common/src/main/java/org/apache/amoro/process/ProcessTriggerStrategy.java
+++ b/amoro-common/src/main/java/org/apache/amoro/process/ProcessTriggerStrategy.java
@@ -23,6 +23,9 @@ import java.time.Duration;
 /** Process trigger strategy. */
 public final class ProcessTriggerStrategy {
 
+  public static final ProcessTriggerStrategy METADATA_TRIGGER =
+      new ProcessTriggerStrategy(Duration.ofDays(1), true, 1);
+
   private final Duration triggerInterval;
 
   private final boolean triggerOnNewSnapshot;
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/AbstractTableRuntime.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/AbstractTableRuntime.java
@@ -20,7 +20,6 @@ package org.apache.amoro.server.table;
 
 import org.apache.amoro.Action;
 import org.apache.amoro.ServerTableIdentifier;
-import org.apache.amoro.TableRuntime;
 import org.apache.amoro.config.TableConfiguration;
 import org.apache.amoro.process.TableProcessStore;
 import org.apache.amoro.server.persistence.PersistentBase;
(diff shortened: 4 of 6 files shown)


amoro-common/src/main/java/org/apache/amoro/process/ProcessFactory.java
```diff
@@ -44,5 +63,5 @@ public interface ProcessFactory {
    * @param state state of the process
    * @return target process which has not been submitted yet.
    */
-  AmoroProcess recover(TableRuntime tableRuntime, TableProcessState state);
+  TableProcess recover(TableRuntime tableRuntime, TableProcessStore store);
```
majin1102: I think we should also use Optional here for the case something went wrong with the recovering or incapatibility issues

amoro-common/src/main/java/org/apache/amoro/process/ProcessTriggerStrategy.java
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.amoro.process;
+
+import java.time.Duration;
+
+/** Process trigger strategy. */
+public final class ProcessTriggerStrategy {
```
majin1102, line 24: Why do we need this?

I originally thought we should always triggered process when we have new snapshots, config changed and time based. And this should be unextensiable to me

amoro-common/src/main/java/org/apache/amoro/table/TableRuntimeFactory.java
```diff
@@ -29,6 +31,10 @@
 /** Table runtime factory. */
 public interface TableRuntimeFactory extends ActivePlugin {
 
+  List<ActionCoordinator> supportedCoordinators();
```
majin1102, line 34: So this extension should be aware of the implementation of Coordinator?

I feel a little weried for Lance scenario. Can we provide an abstract TableRuntime if the coordinator is really necessary here? (Seriouslly I think this should be a legacy issue?)

amoro-common/src/main/java/org/apache/amoro/TableRuntime.java
```diff
@@ -48,6 +48,10 @@ public interface TableRuntime {
    */
   List<? extends TableProcessStore> getProcessStates(Action action);
 
+  void registerProcess(TableProcessStore processStore);
```
majin1102: I think we should always use table runtime to trigger a process instead of register a process into table runtime.

If not, I don't get the meaning of putting factories into runtimes

amoro-common/src/main/java/org/apache/amoro/Action.java
```diff
@@ -20,38 +20,33 @@
...
-  /** supported table formats of this action */
-  private final TableFormat[] formats;
+  private static final Map<String, Action> registeredActions = new ConcurrentHashMap<>();
 
   private final String name;
-  /**
-   * the weight number of this action, the bigger the weight number, the higher positions of
-   * schedulers or front pages
-   */
-  private final int weight;
```
majin1102, line 38: The weight is not used so far.
But it was designed to be used what action should be prioritied if we have multiple actions to schedule in one resouce group.

I was considering to use a priority here and add a default value here. WDYT

majin1102 (commented):
Thanks for working on this.
Left some comments

amoro-common/src/main/java/org/apache/amoro/process/ProcessTriggerStrategy.java
baiyangtx: At least, the factory should tell the scheduler the trigger interval. 

I can provide a default implement.

amoro-common/src/main/java/org/apache/amoro/table/TableRuntimeFactory.java
baiyangtx: You are not need to implement this plugin. 

The `default` TableRuntimeFactory will create all coordinator automaticly, the only plugin you should provide is  `ProcessFactory`.

amoro-common/src/main/java/org/apache/amoro/TableRuntime.java
baiyangtx: This method is used for process scheduler.  

You are not need to care about this method.  And also, you are no longer need to provide the implement of `TableRuntime`

amoro-common/src/main/java/org/apache/amoro/Action.java
baiyangtx: Good idea.

amoro-common/src/main/java/org/apache/amoro/process/ProcessFactory.java
baiyangtx: Maybe in this case you should throw an exception? 

How about define an `UnableRecoverProcess` Exception ?

majin1102: These approaches aren't mutually exclusive. We can throw exceptions or return null to skip this process when needed.

amoro-common/src/main/java/org/apache/amoro/table/TableRuntimeFactory.java
majin1102: You mean I don't need to provide a LanceTableRuntime?

How can I return a LanceTableConfig? Could you provide a simple demo

baiyangtx: You can call  `TableRuntime.getTableConfig(): Map<String, String>`

amoro-common/src/main/java/org/apache/amoro/process/ProcessFactory.java
baiyangtx: When the recover method is called, there must be a record of Process information in a managed state within the system database. From a system‑level perspective, if recovery is not possible at this point, it should be considered an abnormal situation, meaning the system needs to clean up resources and mark the Process as failed or abnormal. Therefore, I suggest explicitly indicating through an exception that this Process can no longer be recovered.

amoro-common/src/main/java/org/apache/amoro/TableRuntime.java
baiyangtx: -- -- getProcessStates -> -- getProcessStore or getProcess

PR #4100, opened 2026-02-26
[Improvement]: Load process factories via DefaultTableRuntimeFactory

baiyangtx:
## Summary

This PR is a subtask of #4100

Wire the new process plugin model(Introduced by #4097) into AMS so that table processes are discovered from `ProcessFactory` implementations and scheduled via the existing `ProcessService`. 

## Details
 

- Promote `ProcessFactory` to the single extension point for table processes.
  - `DefaultTableRuntimeFactory` aggregates all `ProcessFactory` implementations, exposes derived `ActionCoordinator`s, and merges their `requiredStates()` with `DefaultTableRuntime.REQUIRED_STATES`.

- Simplify `TableRuntimeFactory` so it only describes table runtime & process wiring.
  - It no longer extends `ActivePlugin`; lifecycle methods (`open`/`close`/`name`) are removed from the interface and from `DefaultTableRuntimeFactory`.

- Wire AMS directly to `DefaultTableRuntimeFactory` instead of a generic plugin manager.
  - `AmoroServiceContainer` creates a single `DefaultTableRuntimeFactory`, passes it to `DefaultTableService`, then initializes it with all `ProcessFactory` implementations and uses `supportedCoordinators()` to build the coordinator list for `ProcessService`.
  - `DefaultTableService` now depends on a `TableRuntimeFactory` directly and uses it to create `TableRuntime` instances.

- Clean up `ProcessService` constructors to match the new model.
  - Remove the deprecated constructor using `ActionCoordinatorManager`.
  - AMS uses the constructor with an explicit `List<ActionCoordinator>`; tests use the 2‑argument convenience overload.

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntimeFactory.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntimeFactory.java
@@ -122,7 +122,9 @@ public class DefaultTableRuntimeFactory implements TableRuntimeFactory {
       // 2) Extra states from all process factories for this format (if any)
       Map<Action, ProcessFactory> byAction = factoriesByFormat.get(format);
       if (byAction != null) {
-        byAction
+        Map<String, ProcessFactory> deFactories = new HashMap<>();
+        byAction.forEach((a, f) -> deFactories.putIfAbsent(f.name(), f));
+        deFactories
             .values()
             .forEach(
                 factory ->


amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
```diff
@@ -64,8 +64,8 @@ public class ProcessService extends PersistentBase {
       new ConcurrentHashMap<>();
   private final Map<EngineType, ExecuteEngine> executeEngines = new ConcurrentHashMap<>();
 
-  private final ActionCoordinatorManager actionCoordinatorManager;
```
zhoujinsong, line 67: It seems the `ActionCoordinatorManager` class can be dropped now.

amoro-common/src/main/java/org/apache/amoro/table/TableRuntimeFactory.java
```diff
@@ -29,7 +28,7 @@
 import java.util.Optional;
 
 /** Table runtime factory. */
-public interface TableRuntimeFactory extends ActivePlugin {
+public interface TableRuntimeFactory {
```
zhoujinsong, line 31: Do we still need other table runtime factory implementation now, or is the default the only one?

amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
```diff
@@ -74,16 +74,16 @@ public class ProcessService extends PersistentBase {
       new ConcurrentHashMap<>();
 
   public ProcessService(Configurations serviceConfig, TableService tableService) {
-    this(serviceConfig, tableService, new ActionCoordinatorManager(), new ExecuteEngineManager());
+    this(serviceConfig, tableService, Collections.emptyList(), new ExecuteEngineManager());
   }
 
   public ProcessService(
       Configurations serviceConfig,
```
zhoujinsong: The `serviceConfig` seems not be used.

amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
```diff
@@ -229,16 +233,26 @@ public void transitionToFollower() {
...
 
-    tableService =
-        new DefaultTableService(serviceConfig, catalogManager, tableRuntimeFactoryManager);
+    tableService = new DefaultTableService(serviceConfig, catalogManager, defaultRuntimeFactory);
 
     optimizingService =
         new DefaultOptimizingService(serviceConfig, catalogManager, optimizerManager, tableService);
 
-    processService = new ProcessService(serviceConfig, tableService);
+    // Load process factories and build action coordinators from default table runtime factory.
```
zhoujinsong, line 236:
```
    // Load process factories and build action coordinators from default table runtime factory.
    TableProcessFactoryManager tableProcessFactoryManager = new TableProcessFactoryManager();
    tableProcessFactoryManager.initialize();
    List<ProcessFactory> processFactories = tableProcessFactoryManager.installedPlugins();

    DefaultTableRuntimeFactory defaultRuntimeFactory = new DefaultTableRuntimeFactory();
    defaultRuntimeFactory.initialize(processFactories);

    List<ActionCoordinator> actionCoordinators = defaultRuntimeFactory.supportedCoordinators();
    ExecuteEngineManager executeEngineManager = new ExecuteEngineManager();

    tableService = new DefaultTableService(serviceConfig, catalogManager, defaultRuntimeFactory);
    processService =
        new ProcessService(serviceConfig, tableService, actionCoordinators, executeEngineManager);
    optimizingService =
        new DefaultOptimizingService(serviceConfig, catalogManager, optimizerManager, tableService);
```
How about changing the code structure like this?

amoro-common/src/main/java/org/apache/amoro/table/TableRuntimeFactory.java
baiyangtx: no, only this one

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntimeFactory.java
```diff
@@ -28,45 +29,108 @@
...
     public List<StateKey<?>> requiredStateKeys() {
-      return DefaultTableRuntime.REQUIRED_STATES;
+      Map<String, StateKey<?>> merged = new LinkedHashMap<>();
+      // 1) DefaultTableRuntime required states
+      for (StateKey<?> stateKey : DefaultTableRuntime.REQUIRED_STATES) {
+        merged.put(stateKey.getKey(), stateKey);
+      }
+
+      // 2) Extra states from all process factories for this format (if any)
+      Map<Action, ProcessFactory> byAction = factoriesByFormat.get(format);
```
xxubai, line 123: If `DefaultTableRuntime.REQUIRED_STATES` declares a state key with the same name as a `ProcessFactory`'s required state but with a different type or default value, the factory's key silently overwrites the built-in one.

Suggestion: Add a type-compatibility check or throw on conflicting keys:

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntimeFactory.java
```diff
@@ -28,45 +29,108 @@
...
-  public void close() {}
+  /** Mapping from table format to its supported actions and corresponding process factory. */
+  private final Map<TableFormat, Map<Action, ProcessFactory>> factoriesByFormat = new HashMap<>();
 
-  @Override
-  public String name() {
-    return "default";
-  }
+  /** Coordinators derived from all installed process factories. */
+  private final List<ActionCoordinator> supportedCoordinators = Lists.newArrayList();
```
xxubai, line 53: Callers can accidentally mutate the internal supportedCoordinators list. Return Collections.unmodifiableList(supportedCoordinators) instead.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultActionCoordinator.java
```diff
@@ -0,0 +1,124 @@
...
+
+  @Override
+  public Action action() {
+    return action;
+  }
+
+  @Override
+  public long getNextExecutingTime(TableRuntime tableRuntime) {
+    // Fixed-rate scheduling based on configured trigger interval.
+    return strategy.getTriggerInterval().toMillis();
```
xxubai, line 93: The method returns strategy.getTriggerInterval().toMillis() — a duration, not an absolute timestamp.
Do you mean `System.currentTimeMillis() + interval`?

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntimeFactory.java
baiyangtx: I will add a check to throw an exception when conflicts occur.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultActionCoordinator.java
baiyangtx: The getNextExecutingTime method actually returns the duration. This is a legacy issue.

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntimeFactory.java
xxubai: When a single `ProcessFactory` supports multiple actions for the same format, byAction.values() will contain that factory instance multiple times. Each invocation of `factory.requiredStates()` produces the same keys, which will hit the merged.containsKey check and throw an `IllegalStateException`

Fix: Deduplicate factories before iterating?

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultActionCoordinator.java
xxubai: Maybe we should update the interface in a separate PR?

amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntimeFactory.java
baiyangtx: I have deduplicate factories.

PR #4104, opened 2026-03-03
[AMORO-4103] Release table runtime when the group name is changed to a non-existent resource group

### LINKED ISSUE #4103: [Bug]: Tables stuck in PENDING status forever after clear self-optimizing config
### What happened?

When tables are in PLANNING or PENDING optimizing status and their associated resource group is not exist , these tables will be stuck in PLANNING/PENDING status forever. They can never be scheduled for optimizing or return to IDLE status.

This happens in the following scenario:

> Config change + AMS restart race condition: A table's self-optimizing config is changed (e.g., optimizer group changed from "old_group" to "default"), but AMS restarts before the optimizing process is properly closed. The table's persisted status is still PENDING with the group name, which no longer exists. After restart, the table is orphaned and stuck.

Expected behavior: Tables in PLANNING/PENDING status whose resource group no longer exists should be automatically released to IDLE status during AMS startup recovery.

### Affects Versions

0.8.x

Iceberg

AMS

### How to reproduce

1. Create a custom resource group (e.g., "custom_group") and assign some tables to it.
2. Write data to the tables so that they enter PENDING optimizing status.
3. Clear the tables' optimizer group config(`self-optimizing.enabled`=`false` and `self-optimizing.group`=`default`)
4. Restart AMS.
5. Observe that the affected tables remain stuck in PENDING status and are never optimized.

Image

### Relevant log output

```shell

```

### Anything else

The problem occurs every time AMS is restarted when tables with non-existent groups are in PLANNING/PENDING status. The tables will remain stuck until manually reset.

### Code of Conduct

xxubai:
When a table's optimizer group configuration is changed (e.g., from "old_group" to "default"), if the AMS restarts before the optimizing process is properly closed, the table runtime may remain stuck in `PENDING` status forever. This happens in two scenarios:

- AMS restart with non-existent resource group: During l`oadOptimizingQueues`, tables whose persisted optimizer group no longer exists are left in a groupToTableRuntimes map but never released — the previous code only logged a warning without taking any corrective action.
- Runtime config change to non-existent group: When `handleConfigChanged` is triggered and the table's new optimizer group does not exist, there is no fallback to release the table's optimizing process, causing it to hang indefinitely.

Close #4103.

## Brief change log

- **`DefaultOptimizingService.loadOptimizingQueues()`**: Replaced the original logic that merely logged a warning for unloaded table runtimes in non-existent groups. Now, tables in `PLANNING` or `PENDING` status whose resource group does not exist are actively released back to `IDLE` via `completeEmptyProcess()`.

- **`DefaultOptimizingService.ConfigChangeHandler.handleConfigChanged()`**: After refreshing the table in the new group's queue, added logic to release the table from the queue. If the new group's queue does not exist, `completeEmptyProcess()` is called directly on the table runtime to prevent it from being stuck.

- **`TestOptimizingQueue`**: Added two new test cases:
  - `testReleaseOrphanedPlanningTableOnRestart`: Verifies that a table persisted with `PLANNING` status in a non-existent group is correctly released to `IDLE` during AMS restart.
  - `testReleaseOrphanedPendingTableOnRestart`: Verifies the same behavior for tables persisted with `PENDING` status.
  - Added helper method `simulateLoadOptimizingQueuesForNonExistentGroup()` to simulate the `loadOptimizingQueues` logic in tests.

- **`TestDefaultOptimizingService`**: Added two new test cases:
  - `testHandleConfigChangedGroupChanged`: Verifies that when the optimizer group changes to a different **existing** group, the table is properly released from both old and new queues without exceptions.
  - `testHandleConfigChangedGroupNotExist`: Verifies that when the optimizer group changes to a **non-existent** group, `completeEmptyProcess()` is called on the table runtime.

## How was this patch tested?

## Documentation

- If yes, how is the feature documented? Not applicable

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
@@ -398,22 +398,22 @@ public class DefaultOptimizingService extends StatedPersistentBase
     public void handleConfigChanged(TableRuntime runtime, TableConfiguration originalConfig) {
       DefaultTableRuntime tableRuntime = (DefaultTableRuntime) runtime;
       String originalGroup = originalConfig.getOptimizingConfig().getOptimizerGroup();
+      Optional<OptimizingQueue> newQueue = getOptionalQueueByGroup(tableRuntime.getGroupName());
       if (!tableRuntime.getGroupName().equals(originalGroup)) {
         getOptionalQueueByGroup(originalGroup).ifPresent(q -> q.releaseTable(tableRuntime));
+        // If the new group doesn't exist, close the process to avoid the table in limbo(PENDING)
+        // status.
+        if (newQueue.isEmpty()) {
+          LOG.warn(
+              "Cannot find the resource group: {}, try to release optimizing process of table {} directly",
+              tableRuntime.getGroupName(),
+              tableRuntime.getTableIdentifier());
+          tableRuntime.completeEmptyProcess();
+        }
       }
-      getOptionalQueueByGroup(tableRuntime.getGroupName())
-          .ifPresent(q -> q.refreshTable(tableRuntime));
 
-      Optional<OptimizingQueue> queue = getOptionalQueueByGroup(tableRuntime.getGroupName());
-      if (queue.isPresent()) {
-        queue.get().releaseTable(tableRuntime);
-      } else {
-        LOG.warn(
-            "Cannot find resource group: {}, try to release optimizing process of table {} directly",
-            tableRuntime.getGroupName(),
-            tableRuntime.getTableIdentifier());
-        tableRuntime.completeEmptyProcess();
-      }
+      // Binding new queue if the new group exists
+      newQueue.ifPresent(q -> q.refreshTable(tableRuntime));
     }
 
     @Override


amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -384,6 +403,17 @@ public void handleConfigChanged(TableRuntime runtime, TableConfiguration origina
...
+      Optional<OptimizingQueue> queue = getOptionalQueueByGroup(tableRuntime.getGroupName());
+      if (queue.isPresent()) {
+        queue.get().releaseTable(tableRuntime);
+      } else {
+        LOG.warn(
+            "Cannot find resource group: {}, try to release optimizing process of table {} directly",
+            tableRuntime.getGroupName(),
+            tableRuntime.getTableIdentifier());
+        tableRuntime.completeEmptyProcess();
+      }
```
j1wonpark: This block is placed outside the group-change condition (`if (!tableRuntime.getGroupName().equals(originalGroup))`), so it executes on every config change — not just when the group changes.

Two unintended side effects:

- When the new group exists: releaseTable() is called right after refreshTable() on the same queue, immediately undoing the refresh. This removes the table from the scheduler and force-closes any in-flight optimizing process.
- When the group hasn't changed at all (e.g., only the optimizing interval was updated): the table is still unconditionally released from its current queue.
Only the case where the group changed to a non-existent group works as intended.

Suggestion: move this fallback inside the group-change block and remove the duplicate getOptionalQueueByGroup call.

xxubai: Thanks for the review.

PR #4107, opened 2026-03-05
[Improvement]: Refactor snapshot-expiring via ProcessFactory plugin

baiyangtx:

Refactor snapshot expiring action via ProcessFactory plugin

## Brief change log

- Refactor Iceberg snapshot expiration from the inline scheduler into a pluggable process model (ProcessFactory + ExecuteEngine).
- Introduce `IcebergProcessFactory` and `SnapshotsExpiringProcess`, executed by the default `LocalExecutionEngine` with async execution, status tracking and cancellation.
- Externalize enablement/interval knobs via plugin YAML and new config options (e.g. `expire-snapshots.enabled`, `expire-snapshots.interval`).
- Add unit tests for `IcebergProcessFactory` and `LocalExecutionEngine`, and update the “last snapshots expiring” timestamp only after the actual expiration run finishes to preserve correct interval semantics.

## How was this patch tested?

## Documentation


amoro-ams/src/main/java/org/apache/amoro/server/process/iceberg/IcebergProcessFactory.java
```diff
@@ -0,0 +1,132 @@
...
+
+/** Default process factory for Iceberg-related maintenance actions in AMS. */
+public class IcebergProcessFactory implements ProcessFactory {
+
+  public static final String PLUGIN_NAME = "iceberg";
+  public static final ConfigOption<Boolean> SNAPSHOT_EXPIRE_ENABLED =
+      ConfigOptions.key("expire-snapshots.enabled").booleanType().defaultValue(true);
+
+  public static final ConfigOption<Duration> SNAPSHOT_EXPIRE_INTERVAL =
+      ConfigOptions.key("expire-snapshot.interval")
```
czy006: YAML is **expire-snapshots.interval**

amoro-common/src/main/java/org/apache/amoro/process/LocalExecutionEngine.java
```diff
@@ -0,0 +1,205 @@
...
+  public void open(Map<String, String> properties) {
+    Configurations configs = Configurations.fromMap(properties);
+    int defaultSize = configs.getInteger(DEFAULT_POOL_SIZE);
+    pools.put(DEFAULT_POOL, newFixedPool(DEFAULT_POOL, defaultSize));
+
+    Set<String> customPools =
+        properties.keySet().stream()
+            .filter(key -> key.startsWith(POOL_CONFIG_PREFIX))
+            .map(key -> key.substring(POOL_CONFIG_PREFIX.length()))
+            .map(key -> key.substring(0, key.indexOf(".") + 1))
```
czy006: last result is **pool.default..thread-count / pool.snapshots-expiring..thread-count**, that's not get the pool

amoro-ams/src/main/resources/META-INF/services/org.apache.amoro.server.process.executor.ExecuteEngine
czy006, line 1: org.apache.amoro.process.ExecuteEngine

amoro-ams/src/main/java/org/apache/amoro/server/process/iceberg/SnapshotsExpiringProcess.java
```diff
@@ -0,0 +1,82 @@
...
+
+  @Override
+  public String tag() {
+    return getAction().getName().toLowerCase();
+  }
+
+  @Override
+  public void run() {
+    try {
+      AmoroTable<?> amoroTable = tableRuntime.loadTable();
```
czy006, line 54: The problem is that the new scheduling path no longer preserves the old “run, then record cleanup time” behavior for snapshot expiration.

In the old implementation, SnapshotsExpiringExecutor.java executed **tableMaintainer.expireSnapshots()** synchronously. Only after that finished did PeriodicTableScheduler.java (line 125) update lastCleanTime and schedule the next run. So the interval was effectively measured from the end of the previous cleanup.

In the new path, **ActionCoordinatorScheduler.java** (line 103) only submits/registers a process and returns immediately. After that return, PeriodicTableScheduler still updates lastCleanTime right away, even though the real cleanup work has not finished yet. The actual cleanup now happens later in SnapshotsExpiringProcess.java (line 53).

czy006:
It looks like `IcebergProcessFactory` receives available execute engines too early.

In `AmoroServiceContainer`, `availableExecuteEngines(executeEngineManager.installedPlugins())` is called before `executeEngineManager.initialize()`, so `installedPlugins()` is still empty at that point. As a result, `IcebergProcessFactory.localEngine` is never set.

Later, when snapshot expiration is triggered, `triggerExpireSnapshot()` returns `Optional.empty()` because `localEngine == null`, so no `SnapshotsExpiringProcess` is ever created or submitted.

In other words, the new expire-snapshots path is effectively disabled due to initialization order. We probably need to initialize execute engines before injecting them into process factories, or re-inject them after engine initialization.

amoro-ams/src/main/java/org/apache/amoro/server/process/iceberg/SnapshotsExpiringProcess.java
j1wonpark: Building on your observation — the async submission also introduces a state-loss issue in `LocalExecutionEngine.getStatus()`.

`getStatus()` removes the `Future` from the map on terminal states (`isDone`/`isCancelled`), making it **non-idempotent**:

```
Call 1: future.isDone() == true → remove → SUCCESS
Call 2: future == null          → KILLED  (wrong!)
```

`TableProcessExecutor` polls `getStatus()` in a loop (line 107), so if any retry or concurrent access queries the same identifier twice after completion, it gets `KILLED` instead of the real result.

There's also a TOCTOU race between `containsKey` and `get` across `cancelingInstances`/`activeInstances` (lines 67-70), since the compound check-then-act isn't atomic even with `ConcurrentHashMap`.

amoro-common/src/main/java/org/apache/amoro/process/LocalExecutionEngine.java
baiyangtx: I have fix it and add some tests

amoro-ams/src/main/java/org/apache/amoro/server/process/iceberg/SnapshotsExpiringProcess.java
baiyangtx: The information of a completed process is not cleaned up immediately, but waits for a TTL period, which is long enough for the process executor to obtain its final status and persist the state into the database.

PR #4112, opened 2026-03-07
[Feature] Introduce a new plugin RestExtension to extend rest service

baiyangtx:

Provide a mechanism to extend the HTTP API of AMS

## Brief change log

- Add RestExtension / RestExtensionFactory / RestExtensionManager to load REST extensions via ServiceLoader and the existing plugin framework.
- Refactor the existing Iceberg REST Catalog service into a built-in RestExtension (iceberg-rest-catalog) and make it configurable.
- Enable http-server.iceberg-rest-catalog.enabled=true by default to keep the original behavior.

## How was this patch tested?

## Documentation

### DIFF
--- a/dist/src/main/amoro-bin/conf/plugins/rest-extensions.yaml
+++ b/dist/src/main/amoro-bin/conf/plugins/rest-extensions.yaml
@@ -15,8 +15,8 @@
 # limitations under the License.
 #
 
-# This file is the config file for plugin event listener
-# configurations of event listeners
+# This file is the config file for plugin rest extension
+# configurations of rest extensions
 rest-extensions:
   - name: iceberg-rest-catalog
     enabled: true


zhoujinsong:
Thanks for the contribution! I noticed an inconsistency in the plugin configuration approach compared to other existing plugins.

**Current approach in this PR:**

The `iceberg-rest-catalog` extension is configured inside `config.yaml` under the `http-server` section:
```yaml
http-server:
  iceberg-rest-catalog:
    enabled: true
```
This is achieved by overriding `loadPluginConfigurations()` in `RestExtensionManager` to read the `enabled` flag directly from the `Configurations` object, bypassing the standard plugin config mechanism.

**Existing plugin configuration pattern:**

All other plugins (e.g., `metric-reporters`, `event-listeners`, `table-runtime-factories`) use a dedicated YAML file under `conf/plugins/<category>.yaml`:
```yaml
# conf/plugins/metric-reporters.yaml
metric-reporters:
  - name: prometheus-exporter
    enabled: false
    properties:
      port: 7001
```
This is handled automatically by the default `loadPluginConfigurations()` in `AbstractPluginManager`, which supports `enabled`, `priority`, and custom `properties` out of the box.

**Problems with the current approach:**

1. **Inconsistency** — Users familiar with the existing plugin system would expect to configure REST extensions in `conf/plugins/rest-extension.yaml`, not buried inside `config.yaml`.
2. **Limited extensibility** — The design only supports an `enabled` toggle. Third-party `RestExtension` implementations that require additional configuration (e.g., bind port, auth tokens) have no standard place to declare their `properties`.
3. **Reinvents the wheel** — `AbstractPluginManager` already provides full config loading, priority ordering, and lifecycle management. The current implementation duplicates this logic unnecessarily.

**Suggestion:**

Introduce `conf/plugins/rest-extension.yaml` and rely on the default `loadPluginConfigurations()` inherited from `AbstractPluginManager`, keeping it consistent with all other plugin categories. This would also give third-party extensions a proper way to pass custom properties.

zhoujinsong:
Thanks for the quick update! The configuration inconsistency has been addressed nicely.

One minor issue in the newly added `conf/plugins/rest-extensions.yaml` — the comment block appears to be copied from `event-listeners.yaml` and still references the wrong plugin type:

```yaml
# This file is the config file for plugin event listener
# configurations of event listeners
```

Should be updated to something like:

```yaml
# This file is the config file for rest extension plugins
# configurations of rest extensions
```

Small nit, but worth fixing to avoid confusion for users reading the config file.

PR #4114, opened 2026-03-08
[AMORO-4113] Do not update snapshot markers on completeProcess(false)

### LINKED ISSUE #4113: [Bug]: completeProcess incorrectly updates lastOptimizedSnapshotId
### What happened?

When a table's `self-optimizing.group` is changed while an optimizing process is running, the process is forcefully closed via `completeProcess(false)`. However, `completeProcess()` unconditionally updates the `lastOptimizedSnapshotId` and `lastOptimizedChangeSnapshotId` snapshot markers regardless of the `success` parameter.

This causes `SchedulingPolicy.isTablePending()` to return `false` (because `lastOptimizedSnapshotId == currentSnapshotId`), so the scheduler permanently skips the table. The table remains stuck in PENDING status until new data arrives and changes the current snapshot.

**Expected behavior:** When a process is closed without success (`completeProcess(false)`), the snapshot markers should not be updated since no actual optimization was performed. The table should be eligible for re-planning in the new optimizer group.

**Actual behavior:** The snapshot markers are updated even on failure, causing the scheduler to believe the table has already been optimized at the current snapshot, permanently skipping it.

### Affects Versions

master

Iceberg, Mixed-Iceberg, Mixed-Hive

AMS

### How to reproduce

1. Create a table and insert data so that optimization is triggered
2. Wait for the table to enter an optimizing state (MINOR_OPTIMIZING, MAJOR_OPTIMIZING, etc.) with a running `optimizingProcess`
3. Change the optimizer group: `ALTER TABLE ... SET TBLPROPERTIES ('self-optimizing.group' = '<another_group>')`
4. Ensure the new group has active optimizers
5. Observe that the table transitions to PENDING but never progresses to PLANNING
6. The table stays stuck in PENDING indefinitely until new data arrives

This also affects other `completeProcess(false)` paths such as task retry exhaustion and process failures.

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

j1wonpark:

When an optimizing process is closed without success (e.g., optimizer group change, task retry exhaustion), `completeProcess(false)` unconditionally updates `lastOptimizedSnapshotId` and `lastOptimizedChangeSnapshotId`. This causes `SchedulingPolicy.isTablePending()` to return `false`, permanently preventing the table from being re-planned. The table remains stuck in PENDING status until new data arrives.

Close #4113.

## Brief change log

- `DefaultTableRuntime.completeProcess()`: Only update `lastOptimizedSnapshotId` and `lastOptimizedChangeSnapshotId` when `success` is `true`
- `TestOptimizingQueue`: Add `testProcessCloseKeepsLastOptimizedSnapshotId` to verify snapshot markers are preserved on failed process close
- `TestOptimizingQueue`: Add `testTableRescheduledAfterProcessClose` to verify the table can be rescheduled after a forced process close

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
@@ -385,8 +385,11 @@ public class DefaultTableRuntime extends AbstractTableRuntime {
         .updateState(
             OPTIMIZING_STATE_KEY,
             state -> {
-              state.setLastOptimizedSnapshotId(optimizingProcess.getTargetSnapshotId());
-              state.setLastOptimizedChangeSnapshotId(optimizingProcess.getTargetChangeSnapshotId());
+              if (success) {
+                state.setLastOptimizedSnapshotId(optimizingProcess.getTargetSnapshotId());
+                state.setLastOptimizedChangeSnapshotId(
+                    optimizingProcess.getTargetChangeSnapshotId());
+              }
               if (processType == OptimizingType.MINOR) {
                 state.setLastMinorOptimizingTime(optimizingProcess.getPlanTime());
               } else if (processType == OptimizingType.MAJOR) {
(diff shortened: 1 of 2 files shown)


zhoujinsong (approved):
LGTM! The fix is clean and correct — failed/force-closed optimizing processes should not update the snapshot markers, otherwise the table would be incorrectly treated as already optimized and skip necessary future optimization runs. Test coverage is solid too.

PR #4116, opened 2026-03-10
[Improvement]: Extract TableProcessStore from AmoroProcess interface.

baiyangtx:

Extract tableProcessStore from AmoroProcess interface, so that the ProcessFactory will not need to care about how process is presist.

## Brief change log

- AmoroProcess will no longer return a TableProcessStore, so ProcessService will controll all processMeta lifecycle.
- Move ExecuteEngine interface to amoro-common module. 

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
@@ -186,6 +186,14 @@ public class ProcessService extends PersistentBase {
     }
 
     ExecuteEngine executeEngine = executeEngines.get(store.getExecutionEngine());
+    if (executeEngine == null) {
+      LOG.error(
+          "Can't found execution engine:{} for process:{}, table:{}",
+          store.getExecutionEngine(),
+          store.getAction(),
+          process.getTableIdentifier());
+      return;
+    }
 
     TableProcessExecutor executor = new TableProcessExecutor(process, store, executeEngine);
     executor.onProcessFinished(
@@ -238,13 +246,14 @@ public class ProcessService extends PersistentBase {
 
     ExecuteEngine executeEngine = executeEngines.get(store.getExecutionEngine());
 
-    executeEngine.tryCancelTableProcess(process, store.getExternalProcessIdentifier());
-
-    LOG.info(
-        "Cancel table process {} in engine {}, process id:{}",
-        process,
-        executeEngine.engineType(),
-        store.getProcessId());
+    if (executeEngine != null) {
+      executeEngine.tryCancelTableProcess(process, store.getExternalProcessIdentifier());
+      LOG.info(
+          "Cancel table process {} in engine {}, process id:{}",
+          process,
+          executeEngine.name(),
+          store.getProcessId());
+    }
   }
 
   /**


amoro-common/src/main/java/org/apache/amoro/process/ProcessFactory.java
```diff
@@ -49,6 +50,8 @@ default ProcessTriggerStrategy triggerStrategy(TableFormat format, Action action
     return ProcessTriggerStrategy.METADATA_TRIGGER;
   }
 
+  void availableExecuteEngines(Collection<ExecuteEngine> allAvailableEngines);
```
zhoujinsong: This new method is missing a JavaDoc comment. Please clarify:
1. **Purpose** — What should a `ProcessFactory` implementation do when it receives the available engine list? When is this called?
2. **Parameter semantics** — Does `allAvailableEngines` contain all registered engines, or only those compatible with this factory?
3. **Default implementation** — Existing `ProcessFactory` implementations do not override this method, which will cause compilation failures. If this method is optional, please add a `default` empty implementation:
   ```java
   default void availableExecuteEngines(Collection<ExecuteEngine> allAvailableEngines) {}
   ```

amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
```diff
@@ -546,6 +506,17 @@ protected void doDispose() {
     }
   }
 
+  @VisibleForTesting
```
zhoujinsong, line 521: Using `public final` fields directly in `TableProcessHolder` breaks encapsulation. Callers are already accessing `holder.store` and `holder.process` throughout the codebase, which will make it hard to change later.

Please consider using getter methods instead:
```java
public static class TableProcessHolder {
    private final TableProcess process;
    private final TableProcessStore store;

    public TableProcess getProcess() { return process; }
    public TableProcessStore getStore() { return store; }
}
```

amoro-common/src/main/java/org/apache/amoro/process/ProcessFactory.java
baiyangtx: 1. When create a process, it must provide a engine name, so that process service could find target execution engine to submit the process. 
2. All registered engines. The procesFactory determines which engine to use to create processes, such as if no RemoteEngine is available, the factory could create a process which run on a local engine.
3. OK

amoro-common/src/main/java/org/apache/amoro/process/ProcessFactory.java
```diff
@@ -49,6 +50,13 @@ default ProcessTriggerStrategy triggerStrategy(TableFormat format, Action action
     return ProcessTriggerStrategy.METADATA_TRIGGER;
   }
 
+  /**
+   * ALl execute engine registered.
```
zhoujinsong:
```suggestion
   * All execute engine registered.
```

zhoujinsong (approved):
LGTM.
Thanks a lot for the work!

amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
```diff
@@ -249,37 +217,35 @@ private void executeOrTraceProcess(TableProcess process) {
...
+        store.getExternalProcessIdentifier(),
+        "Gracefully Canceled.",
+        process.getProcessParameters(),
+        process.getSummary());
+    untrackTableProcessInstance(
+        process.getTableRuntime().getTableIdentifier(), store.getProcessId());
 
-    ExecuteEngine executeEngine =
-        executeEngines.get(EngineType.of(process.store().getExecutionEngine()));
+    ExecuteEngine executeEngine = executeEngines.get(EngineType.of(process.getExecutionEngine()));
```
xxubai: In `executeOrTraceProcess`: `executeEngines.get(EngineType.of(store.getExecutionEngine()))`  
In `cancelProcess`: `executeEngines.get(EngineType.of(process.getExecutionEngine()))`  

Both should use the same source. In the cancel path, the store's engine value (persisted) is more reliable since it reflects the actual state at registration time.

amoro-ams/src/main/java/org/apache/amoro/server/process/TableProcessMeta.java
```diff
@@ -188,6 +191,24 @@ public static TableProcessMeta fromTableProcessStore(TableProcessStore tableProc
...
+  public static TableProcessMeta createProcessMeta(TableProcess process) {
+    TableProcessMeta tableProcessMeta = new TableProcessMeta();
+    tableProcessMeta.setProcessId(idGenerator.generateId());
+    tableProcessMeta.setTableId(process.getTableIdentifier().getId());
+    tableProcessMeta.setExternalProcessIdentifier("");
+    tableProcessMeta.setStatus(ProcessStatus.PENDING);
+    tableProcessMeta.setProcessType(process.getAction().getName());
+    tableProcessMeta.setProcessStage(process.getProcessStage());
+    tableProcessMeta.setExecutionEngine(process.getExecutionEngine());
+    tableProcessMeta.setRetryNumber(1);
```
xxubai: This means every new process starts with retry count 1 instead of 0. Is this intentional?

amoro-common/src/main/java/org/apache/amoro/TableRuntime.java
```diff
@@ -59,6 +59,13 @@ public interface TableRuntime {
    */
   ServerTableIdentifier getTableIdentifier();
 
+  /**
+   * Load the current table instance for this runtime.
+   *
+   * <p>This method is mainly intended for in-AMS processes.
+   */
+  AmoroTable<?> loadTable();
```
xxubai, line 67: Adding loadTable() to the TableRuntime interface in amoro-common ties the runtime abstraction to table loading, which is an AMS-level concern. Consider a separate TableLoader interface or keeping it on DefaultTableRuntime only.

amoro-ams/src/main/java/org/apache/amoro/server/process/TableProcessMeta.java
baiyangtx: It should be 0.

amoro-common/src/main/java/org/apache/amoro/TableRuntime.java
baiyangtx: When processFactory try to create a Process for table, it need to visit the original datalake table. 

Adding a `loadTable` to tableRuntime make ProcessFactory implement simpler. 

The goal of this PR is to shift the extension of format and process from TableRuntime to the ProcessFactory interface, so that in the future, users do not need to extend a TableRuntime implementation, And also no longer need to worry about how to provide the TableLoader.

amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
```diff
@@ -430,15 +393,15 @@ private void trackTableProcess(
...
     if (inner == null) {
       return null;
     }
-    TableProcess removed = inner.remove(processId);
+    TableProcessHolder removed = inner.remove(processId);
     if (inner.isEmpty()) {
       activeTableProcess.remove(serverTableIdentifier, inner);
     }
-    return removed;
+    return removed.getProcess();
```
xxubai:
```suggestion
    return removed != null ? removed.getProcess() : null;
```

amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
```diff
@@ -249,37 +216,35 @@ private void executeOrTraceProcess(TableProcess process) {
...
+        store.getExternalProcessIdentifier(),
+        "Gracefully Canceled.",
+        process.getProcessParameters(),
+        process.getSummary());
+    untrackTableProcessInstance(
+        process.getTableRuntime().getTableIdentifier(), store.getProcessId());
 
-    ExecuteEngine executeEngine =
-        executeEngines.get(EngineType.of(process.store().getExecutionEngine()));
+    ExecuteEngine executeEngine = executeEngines.get(process.getExecutionEngine());
```
xxubai:
```suggestion
    ExecuteEngine executeEngine = executeEngines.get(store.getExecutionEngine());
```
While currently the values should match, using different sources is fragile. Suggest always using store.getExecutionEngine() for consistency.

amoro-ams/src/main/java/org/apache/amoro/server/process/ProcessService.java
```diff
@@ -249,37 +216,35 @@ private void executeOrTraceProcess(TableProcess process) {
...
 
-    executeEngine.tryCancelTableProcess(process, process.getExternalProcessIdentifier());
+    executeEngine.tryCancelTableProcess(process, store.getExternalProcessIdentifier());
```
xxubai: executeEngines.get(...) may return null if engine is not registered, leading to NPE on the next call. Same risk in executeOrTraceProcess.

PR #4119, opened 2026-03-11
[Security] Disable external entity processing in XML upload to prevent XXE

zhoujinsong:

When users upload XML configuration files (e.g. `core-site.xml`, `hdfs-site.xml`) via the AMS dashboard, the uploaded bytes are parsed by `Hadoop Configuration.addResource()`. Although the current classpath includes Woodstox (which does not expand external entities by default), this implicit protection is fragile — it can silently break if the dependency is excluded due to a version conflict in the future.

This patch adds explicit XXE protection at the Amoro layer before delegating to Hadoop `Configuration`, ensuring the security guarantee holds regardless of the underlying XML parser implementation on the classpath.

Without explicit protection, a malicious user could upload a crafted XML file containing an external entity reference like:

```xml
<?xml version="1.0"?>
<!DOCTYPE foo [
  <!ENTITY xxe SYSTEM "file:///etc/passwd">
]>
<configuration>&xxe;</configuration>
```

If the XXE implicit protection were ever lost (e.g. Woodstox excluded), this could allow:
- Arbitrary local file read from the AMS server
- SSRF (Server-Side Request Forgery) via external URLs in entity references

## How was this patch tested?

Manual testing: uploaded a well-formed XML file (accepted) and an XML file with an external entity reference (rejected with error response).

## Does this PR introduce _any_ user-facing change?

No. Legitimate Hadoop XML configuration files (`core-site.xml`, `hdfs-site.xml`, `hive-site.xml`) do not use external entities. Valid files continue to upload successfully.

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/PlatformFileInfoController.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/PlatformFileInfoController.java
@@ -27,6 +27,9 @@ import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.hadoop.conf.Configuration;
 
+import javax.xml.XMLConstants;
+import javax.xml.stream.XMLInputFactory;
+
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/PlatformFileInfoController.java
```diff
@@ -52,7 +55,15 @@ public void uploadFile(Context ctx) throws IOException {
     // validate xml config
     if (name.toLowerCase().endsWith(".xml")) {
       try {
-        Configuration configuration = new Configuration();
+        // Explicitly disable external entity processing to prevent XXE attacks,
+        // regardless of the underlying XML parser implementation on the classpath.
+        XMLInputFactory xif = XMLInputFactory.newInstance();
+        xif.setProperty(XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES, false);
+        xif.setProperty(XMLInputFactory.SUPPORT_DTD, false);
+        xif.setProperty(XMLConstants.FEATURE_SECURE_PROCESSING, true);
```
xxubai, line 63: `FEATURE_SECURE_PROCESSING` is not a standard property of `XMLInputFactory` and may throw an exception.

amoro-ams/src/main/java/org/apache/amoro/server/dashboard/controller/PlatformFileInfoController.java
```diff
@@ -52,7 +55,15 @@ public void uploadFile(Context ctx) throws IOException {
...
+        xif.createXMLStreamReader(new ByteArrayInputStream(bytes)).close();
+
+        Configuration configuration = new Configuration(false);
         configuration.addResource(new ByteArrayInputStream(bytes));
```
xxubai, line 67: The XXE fix is still incomplete. The new XMLInputFactory parse only acts as a pre-check, but the actual validation still happens in Configuration.addResource(new ByteArrayInputStream(bytes)). In Hadoop 3.4.0, this overload does not use the restricted parser path by default, so the XML is reparsed in a way that can still allow DTD / external entity processing. In other words, this adds an extra parse, but it does not fully close the XXE risk.

xxubai (commented):
Can you also add some junit5 tests for PlatformFileInfoController?

PR #4136, opened 2026-03-19
[AMORO-4135][amoro-web] Fix blank duration column for sub-second optimizing durations

### LINKED ISSUE #4135: [Bug]: Frontend duration column is blank when optimizing duration is less than 1 second
## What happened?

In the Amoro web UI, the optimizing table duration column becomes blank when the backend returns a duration between `1ms` and `999ms`.

This happens because `formatMS2Time()` in `amoro-web/src/utils/index.ts` only appends non-zero day/hour/minute/second units. For values smaller than `1000ms`, all displayed units are `0`, so the formatted result becomes an empty string.

Expected behavior:

- Durations between `1ms` and `999ms` should be rendered as `<1s`
- `1000ms` should still be rendered as `1s`
- Existing formatting for longer durations should remain unchanged

This issue affects the optimizing table page and any other frontend view that reuses the same formatter.

Image

## Affects Versions

master/0.8.1

- Iceberg

- AMS
- Optimizer

## How to reproduce

1. Start the frontend in mock mode:

   ```bash
   cd amoro-web
   npm run dev:mock
   ```

2. Open the optimizing table page in the web UI.
3. Prepare or mock one or more optimizing table records whose `duration` value is between `1` and `999`.
4. Observe the duration column.

Current result:

- The corresponding duration cell is blank.

Expected result:

- The same cell should display `<1s`.

## Relevant log output

No backend log is required. This is a frontend formatting issue.

## Anything else

Suggested implementation scope:

- Update `formatMS2Time()` in `amoro-web/src/utils/index.ts`
- Add handling for `0 < ms < 1000`
- Optionally add or update mock data in `amoro-web/mock/modules/optimize.js` so the behavior is easy to verify manually

Suggested acceptance criteria:

- `500ms -> <1s`
- `999ms -> <1s`
- `1000ms -> 1s`
- Existing values such as `3000ms -> 3s` still work

Bingtagui404:
## Problem

When the backend returns an optimizing duration between `1ms` and `999ms`, the duration column in the Optimizing table page is **blank**.

This happens because `formatMS2Time()` in `amoro-web/src/utils/index.ts` only appends time units whose value is greater than 0. For values smaller than `1000ms`, all units (days/hours/minutes/seconds) are `0` after `Math.floor`, so the formatted result is an empty string.

fix #4135

## Fix Scope

Single file: `amoro-web/src/utils/index.ts`

Added a check after the formatting loop: if the result string is empty but `ms > 0`, return `<1s` instead of an empty string.

```typescript
if (!result && ms > 0) {
  return '<1s'
}
```

## Verification

Tested with `npm run dev:mock` — added a mock record with `duration: 500` and confirmed the duration column displays `<1s`.

**Before (duration: 500ms):** blank cell
**After (duration: 500ms):** `<1s`

### Unit test results

| Input (ms) | Expected | Actual | Result |
|---|---|---|---|
| 0 | `` | `` | ✅ |
| 1 | `<1s` | `<1s` | ✅ |
| 500 | `<1s` | `<1s` | ✅ |
| 999 | `<1s` | `<1s` | ✅ |
| 1000 | `1s` | `1s` | ✅ |
| 1500 | `1s` | `1s` | ✅ |
| 3000 | `3s` | `3s` | ✅ |
| 60000 | `1min` | `1min` | ✅ |
| 61000 | `1min 1s` | `1min 1s` | ✅ |
| 1195501081 | `13d 20h` | `13d 20h` | ✅ |

### DIFF
--- a/amoro-web/src/utils/index.ts
+++ b/amoro-web/src/utils/index.ts
@@ -93,6 +93,10 @@ export function formatMS2Time(ms: number): string {
   }
 
   // Trim any trailing spaces and return the formatted result
+  // Handle sub-second durations (0 < ms < 1000)
+  if (!result && ms > 0) {
+    return '<1s'
+  }
   return result.trim()
 }
 /**


Bingtagui404:
amoro-fix-screenshot

Issue #4131, opened 2026-03-18
[Improvement]: Add non-partitioned table test coverage for mixed-hive rewrite and overwrite operations


xxubai:
### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/amoro/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

The mixed-hive test suite already covers several rewrite and overwrite scenarios, but multiple test methods still contain the same TODO: add cases for tables without partition spec.

Examples can be found in:
- `amoro-format-mixed/amoro-mixed-hive/src/test/java/org/apache/amoro/hive/op/TestRewriteFiles.java`
- `amoro-format-mixed/amoro-mixed-hive/src/test/java/org/apache/amoro/hive/op/TestOverwriteFiles.java`

Currently these cases are guarded by `Assume.assumeTrue(isPartitionedTable())`, so behavior for non-partitioned tables is not covered well enough.

### How should we improve?

We should add non-partitioned table test coverage for mixed-hive rewrite and overwrite operations.

A contributor can start by:
- identifying the existing partitioned-only test cases in `TestRewriteFiles` and `TestOverwriteFiles`
- adding equivalent or adapted cases for tables without partition spec
- verifying whether non-partitioned tables should succeed or throw the expected exception in each scenario
- keeping the new tests independent and easy to understand

### Are you willing to submit PR?

- [ ] Yes I am willing to submit a PR!

### Subtasks

- [ ] Review the existing partitioned-table cases in `TestRewriteFiles`
- [ ] Review the existing partitioned-table cases in `TestOverwriteFiles`
- [ ] Add non-partitioned table coverage for the missing scenarios
- [ ] Run the relevant mixed-hive tests to validate the expected behavior

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

Takch02:
Hi, I would like to work on this issue as my first code contribution! Since I am currently studying the codebase and this is my first time writing tests for this project, it might take a week or two. Could you assign it to me?

xxubai:
@Takch02 Sure, go ahead

JeonDaehong:
@xxubai Hi, there has been no PR for this issue for over 2 weeks. Would it be okay if I work on this and submit a PR?

Takch02:
[...]

PR #4137, opened 2026-03-20
[AMORO-4131] Add non-partitioned table tests for RewriteFiles and OverwriteFiles

Takch02:
#4131 
Motivation

Currently, TestRewriteFiles and TestOverwriteFiles only verify the behavior of partitioned tables, expecting a CannotAlterHiveLocationException when partial rewrites or overwrites attempt to reference multiple directories. Non-partitioned tables, however, are not bound by Hive's strict 1-partition-1-folder constraint and should successfully commit these operations. This PR adds the missing coverage for non-partitioned tables.

Key Changes

Added NonPartitioned methods in TestRewriteFiles and TestOverwriteFiles using Assume.assumeFalse(isPartitionedTable()).

Verified successful commits instead of expecting exceptions, as non-partitioned tables gracefully handle files across multiple directories.

Introduced Dynamic Assertions: Replaced hardcoded integer values with dynamic size assertions to prevent fragile tests. This ensures the tests remain robust regardless of table properties (e.g., Primary Key bucketing or target file size limits).

Comprehensive State Verification: Implemented a two-step verification process for each test:

Action Verification: Checked the exact number of files added in the latest transaction using lastedAddedFiles().

State Verification: Checked the total number of live files in the current snapshot using baseStore.newScan().planFiles().

Example of Verification Logic:

Rewrite: Verified afterFiles.size() and totalLiveFiles against addFiles.size().

Overwrite: Verified afterFiles.size() against secondDataFiles.size(), and totalLiveFiles against firstDataFiles.size() + secondDataFiles.size().


Takch02:
I re-uploaded it after checking the CI and adhering to code conventions.

Takch02:
@xxubai 
Please review the PR! Thank you.

xxubai (commented):
Thanks for adding the non-partitioned coverage. I think the new assertions are still a bit weak for the expected-success cases.

For unpartitioned tables, the current implementation still updates the Hive table location to only the directory of the first added file. So if the rewritten/overwritten files are actually under multiple directories, the commit may succeed, but Hive may still see only part of the data.

Because of that, checking only `lastedAddedFiles()` and `planFiles()` does not seem sufficient here. These assertions show that the Iceberg commit succeeded, but they do not verify that the Hive-side result is actually correct.

Takch02:
> Thanks for adding the non-partitioned coverage. I think the new assertions are still a bit weak for the expected-success cases.
> 
> For unpartitioned tables, the current implementation still updates the Hive table location to only the directory of the first added file. So if the rewritten/overwritten files are actually under multiple directories, the commit may succeed, but Hive may still see only part of the data.
> 
> Because of that, checking only `lastedAddedFiles()` and `planFiles()` does not seem sufficient here. These assertions show that the Iceberg commit succeeded, but they do not verify that the Hive-side result is actually correct.

Thank you for the review!

You are right. Currently, I was only checking the Iceberg assertion and not looking at the Hive side.

I fully understand that in non-partitioned tables, if files exist in multiple directories, the Hive location might only point to the first folder, meaning only a portion of the data can be visible.

I will add the actual data validation code for the Hive side and push it soon.

Takch02:
Hi @xxubai, I added Hive-side validation, and as you mentioned, Hive can only see the files located in the directory of the first added file. The files in other directories become invisible to Hive, which causes the validation with all afterFiles to fail.

```
    List<DataFile> afterFiles = HiveDataTestHelpers.lastedAddedFiles(baseStore);

    int totalLiveFiles = Lists.newArrayList(baseStore.newScan().planFiles()).size();
    Assert.assertEquals(addFiles.size(), totalLiveFiles);

      Table hiveTable = TEST_HMS.getHiveClient()
              .getTable(getMixedTable().id().getDatabase(), getMixedTable().id().getTableName());
      String hiveLocation = hiveTable.getSd().getLocation();

      List<DataFile> filesVisibleToHive = afterFiles.stream()
              .filter(f -> f.path().toString().startsWith(hiveLocation))
              .collect(Collectors.toList());

      UpdateHiveFilesTestHelpers.validateHiveTableValues(
              TEST_HMS.getHiveClient(), getMixedTable(), filesVisibleToHive); // OK

      UpdateHiveFilesTestHelpers.validateHiveTableValues(
              TEST_HMS.getHiveClient(), getMixedTable(), afterFiles);  // Failed
```

Given this behavior, how should we handle this in the test? Should we expect a CannotAlterHiveLocationException here as well (similar to how partitioned tables behave when spanning multiple directories), or should we just assert that only a subset of files is visible to Hive to document the current limitation?

Takch02:
@xxubai
Please review! Thank you.

xxubai:
cc @Aireed Can you help review this Mixed Hive unit tests

Takch02:
@xxubai @Aireed I know you're busy, but please review this just this once!

xxubai:
Thanks for verifying this. I think we should not treat partial Hive visibility as expected behavior. If Hive cannot see all data across multiple directories, the operation should be considered invalid, so I'd prefer expecting `CannotAlterHiveLocationException` for these non-partitioned cases as well.

Takch02:
@xxubai Thank you for the feedback! I agree that partial Hive visibility should not be treated as valid behavior.

However, after investigating, I found that the current production code does not throw `CannotAlterHiveLocationException` for non-partitioned tables — it simply commits successfully even when files span multiple directories. So if I update the non-partitioned tests to expect the exception, they will fail against the current implementation.

To fully address this, the production logic (likely in `UpdateHiveFiles`) would also need to be updated to enforce the same directory constraint for non-partitioned tables.

Would it be okay to include that production fix in this PR, or would you prefer to handle it in a separate PR?

```

@Test
  public void testRewritePartFilesNonPartitioned() {
    Assume.assumeFalse(isPartitionedTable());
   ...
   List<DataFile> dataFiles =
        HiveDataTestHelpers.writerOf(getMixedTable()).transactionId(1L).writeHive(insertRecords);
    RewriteFiles rewriteFiles = baseStore.newRewrite();
    rewriteFiles.rewriteFiles(Sets.newHashSet(deleteFile), Sets.newHashSet(dataFiles));
    Assert.assertThrows(CannotAlterHiveLocationException.class, rewriteFiles::commit);  // False
  }

czy006:
> @xxubai Thank you for the feedback! I agree that partial Hive visibility should not be treated as valid behavior.
> 
> However, after investigating, I found that the current production code does not throw `CannotAlterHiveLocationException` for non-partitioned tables — it simply commits successfully even when files span multiple directories. So if I update the non-partitioned tests to expect the exception, they will fail against the current implementation.
> 
> To fully address this, the production logic (likely in `UpdateHiveFiles`) would also need to be updated to enforce the same directory constraint for non-partitioned tables.
> 
> Would it be okay to include that production fix in this PR, or would you prefer to handle it in a separate PR?
> 
> ```
> 
> @Test
>   public void testRewritePartFilesNonPartitioned() {
>     Assume.assumeFalse(isPartitionedTable());
>    ...
>    List<DataFile> dataFiles =
>         HiveDataTestHelpers.writerOf(getMixedTable()).transactionId(1L).writeHive(insertRecords);
>     RewriteFiles rewriteFiles = baseStore.newRewrite();
>     rewriteFiles.rewriteFiles(Sets.newHashSet(deleteFile), Sets.newHashSet(dataFiles));
>     Assert.assertThrows(CannotAlterHiveLocationException.class, rewriteFiles::commit);  // False
>   }
> ```

I think it's possible to add this to the PR to complete the process.

Takch02:
@czy006 I'll upload it including production code modifications and test code!

Takch02:
@czy006 @xxubai 

Non-partitioned Hive tables support only a single directory as their location. Previously, partial rewrites or overwrites that moved files to a new directory were allowed even when existing files remained in the old location, causing those files to become invisible to Hive.

Added checkNonPartitionedHiveLocationChange() in UpdateHiveFiles to validate that all files in the current Hive location are included in the delete set before allowing a location change. The check uses the Iceberg snapshot rather than the filesystem to avoid false positives from files written before commit.

Also fixed non-partitioned test cases to TestRewriteFiles and TestOverwriteFiles to verify that CannotAlterHiveLocationException is thrown in these scenarios, consistent with the existing behavior for partitioned tables.

czy006 (approved):

PR #4142, opened 2026-03-24
[AMORO-3853] Stabilize continuous optimizing tests on Java 17

xxubai:

Some continuous optimizing tests are unstable on Java 17 due to non-deterministic test data generation and missing test build configuration for generated sources.

Follow #3853.

## Brief change log

- make `RandomRecordGenerator` deterministic for the affected test flows
- update keyed/unkeyed continuous optimizing tests to use the deterministic seed
- exclude `src/main/gen-java/**` from checkstyle in `amoro-common`
- keep surefire configured in `amoro-common` for test execution

## How was this patch tested?

Local verification:
- `./mvnw -pl amoro-ams -am -Dtest=TestKeyedContinuousOptimizing,TestUnKeyedContinuousOptimizing test`

## Documentation


zhoujinsong (approved):
The fix is correct — the root cause is that `randomRecord()` was creating a new `Random()` on every call, completely bypassing the seed passed to the constructor. The other changes (UUID generation, fixed base timestamps) are also necessary to achieve full determinism.

Minor nit: `truncatedTo(ChronoUnit.MICROS)` after `.plusNanos(random.nextInt(1_000_000) * 1_000L)` is redundant since the result is already microsecond-aligned. Not a blocker.

LGTM.

xxubai:
Thanks @zhoujinsong

PR #4161, opened 2026-04-01
[hotfix] Implement an independent process history cleanup mechanism and migrate the configuration item to the Duration type.

czy006:

Implement an independent process history cleanup mechanism and migrate the configuration item to the Duration type.

Close #xxx.

## Brief change log

- process.history-data-keep-time and self-optimizing.runtime-data-expire-interval use duration
- OptimizingExpiringExecutor rename to ProcessDataExpiringExecutor
- ProcessDataExpiringExecutor expire process history

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/AmoroManagementConf.java
@@ -479,19 +479,37 @@ public class AmoroManagementConf {
           .defaultValue(10)
           .withDescription("The number of threads that self-optimizing uses to submit results.");
 
+  /** @deprecated Use {@link #OPTIMIZING_RUNTIME_DATA_KEEP_TIME} instead. */
+  @Deprecated
   public static final ConfigOption<Integer> OPTIMIZING_RUNTIME_DATA_KEEP_DAYS =
       ConfigOptions.key("self-optimizing.runtime-data-keep-days")
           .intType()
           .defaultValue(30)
           .withDescription(
-              "The number of days that self-optimizing runtime data keeps the runtime.");
+              "Deprecated: use 'self-optimizing.runtime-data-keep-time' instead. "
+                  + "The number of days that self-optimizing runtime data keeps the runtime.");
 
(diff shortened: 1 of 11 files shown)


zhoujinsong:
Thanks for the PR! One question on the config design: is there a specific reason to have two separate keep-time parameters (`self-optimizing.runtime-data-keep-time` and `process.history-data-keep-time`)? Since `table_process` already gets cleaned by `keepTime` anyway (step 1 deletes everything older than 30d), the `processKeepTime` only adds an earlier cleanup for terminal records.

If the data volume of `table_process` is not a concern, we could simplify this to a single `runtime-data-keep-time` parameter and avoid the cognitive overhead of having users understand the distinction between the two. Could you share the motivation for keeping them separate?

zhoujinsong:
There's a potential data leak issue: the cleanup in `ProcessDataExpiringExecutor` is driven per `TableRuntime`. If a table is dropped, its `TableRuntime` is removed and the executor will never process it again — leaving orphaned records in `table_process` (and related tables) that will never be cleaned up.

`PeriodicTableScheduler.handleTableRemoved()` currently does nothing, so there's no trigger to clean up process history on table drop.

This doesn't need to block this PR, but worth tracking as a follow-up: either hook into `handleTableRemoved` to do a one-shot cleanup, or add a periodic global scan that deletes records for table IDs no longer in `table_metadata`.

czy006:
- The table corresponding to OPTIMIZING_RUNTIME_DATA_EXPIRE_INTERVAL_HOURS is optimizing_process_state. 
- The table corresponding to PROCESS_HISTORY_DATA_KEEP_DAYS is table_process.

Currently, there are two records: one for PROCESS execution and one for OPTIMIZING_PROCESS. Should we merge them later? cc @zhoujinsong

zhoujinsong:
> Currently, there are two records: one for PROCESS execution and one for OPTIMIZING_PROCESS. Should we merge them later? cc @zhoujinsong

Merging them later look good to me.

zhoujinsong (approved):
LGTM.

Thanks for the work!

PR #4178, opened 2026-04-13
[AMORO-4122][ams] Fix partition filter not working due to @JsonIgnore on PendingInput.partitions

### LINKED ISSUE #4122: [Bug]: Partition filter does not work
### What happened?

The field `partitions` in `org.apache.amoro.optimizing.plan.AbstractOptimizingEvaluator.PendingInput` is marked as "@JsonIgnore"

Image

When `partitions` is set into the table runtime and subsequently retrieved while in the `planning` state via the TableRuntimeStore, `partitions` loss occurs:

Image

Image

Image

### Affects Versions

master

_No response_

_No response_

### How to reproduce

_No response_

### Relevant log output

```shell

```

### Anything else

_No response_

### Code of Conduct

lintingbin:

Close #4122

The `partitions` field in `AbstractOptimizingEvaluator.PendingInput` is annotated with `@JsonIgnore`, which causes partition information to be silently dropped during JSON serialization/deserialization through `TableRuntimeStore` (stored as CLOB in the database).

When `PendingInput` is read back from the database, the `partitions` map is always empty. This causes the partition filter in `IcebergTableUtil.createOptimizingPlanner()` to fall back to `Expressions.alwaysTrue()`, meaning all partitions are scanned instead of only the ones that need optimizing.

### Fix

Add a serializable `partitionPaths` field (`Set<String>`) that stores partition info as `specId:partitionPath` strings. On serialization, build paths from the in-memory partitions map via `PartitionSpec.partitionToPath()`; on deserialization, rebuild the partitions map from paths using `DataFiles.data()`.

## How was this patch tested?

- `TestIcebergTableUtil`: 4 tests passed
- `TestOptimizingEvaluator`: 20 tests passed
- `TestOptimizingPlanner`: 20 tests passed


amoro-format-iceberg/src/main/java/org/apache/amoro/optimizing/plan/AbstractOptimizingEvaluator.java
```diff
@@ -275,6 +286,72 @@ public Map<Integer, Set<StructLike>> getPartitions() {
...
+          String partitionPath = path.substring(colonIdx + 1);
+          PartitionSpec spec = MixedTableUtil.getMixedTablePartitionSpecById(table, specId);
+          if (spec != null) {
+            StructLike struct;
+            if (spec.isUnpartitioned() || partitionPath.isEmpty()) {
+              // For unpartitioned tables, use an empty record since there's no path to parse
+              struct = TablePropertyUtil.EMPTY_STRUCT;
+            } else {
+              struct = DataFiles.data(spec, partitionPath);
+            }
```
czy006, line 329: Using `org.apache.iceberg.DataFiles.data()` here will break partition rebuild for Mixed-format tables. Iceberg's native `DataFiles.data()` delegates to `Conversions.fromPartitionString()`, which only handles standard Iceberg type conversions. It does **not** support Amoro's custom partition transforms:

- **`month` transform**: Mixed-format tables serialize month partitions as `yyyy-MM` (e.g., `month=2024-01`). Iceberg's parser expects an integer month offset, not a date string, and will throw `NumberFormatException`.
- **`hour` transform**: Similarly serialized as `yyyy-MM-dd-HH`, which Iceberg cannot parse.
- **Hive null sentinel**: `__HIVE_DEFAULT_PARTITION__` is not handled by Iceberg's parser.

Amoro already provides `MixedDataFiles.data(PartitionSpec, String)` in `org.apache.amoro.utils.MixedDataFiles` (line 110) that correctly handles all these cases via its own `fromPartitionString()`.

The `catch(Exception)` block below will silently swallow these parse failures, logging only a warning while leaving `partitions` empty — producing the same `Expressions.alwaysTrue()` fallback as the original bug. This effectively makes the fix a no-op for Mixed-format tables with custom transforms.

**Suggested fix:**
```java
import org.apache.amoro.utils.MixedDataFiles;
// ...
struct = MixedDataFiles.data(spec, partitionPath);
```

Also, please consider adding test cases for:
1. Mixed-format tables with `month`/`hour` partition transforms (round-trip serialization)
2. Unpartitioned tables (ensure `EMPTY_STRUCT` path works correctly)

lintingbin: Modified

czy006:
cc @zhoujinsong

PR #4180, opened 2026-04-14
[AMORO-3988][ams] Automatic restart the optimizers if it is down unexpected

lintingbin:
### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/amoro/issues?q=is%3Aissue) and found no similar issues.

### What type of PR is this?

- [x] Improvement
- [ ] Bug Fix
- [ ] Feature
- [ ] Refactoring

### What does this PR do?

Add an automatic restart mechanism for optimizers that die unexpectedly.

#### Problem
When an optimizer process crashes, the `OptimizerKeeper` detects the heartbeat timeout and removes the optimizer record from the `optimizer` table. However, the corresponding resource record in the `resource` table is **not cleaned up**, leaving an "orphaned" resource with no active optimizer. Currently there is no mechanism to detect this situation and restart the optimizer.

#### Solution
Extend the existing `OptimizerGroupKeeper` to periodically detect orphaned resources and automatically restart the optimizer:

1. **Orphaned resource detection**: In `OptimizerGroupKeeper.processTask()`, cross-check the `resource` table against the `optimizer` table to find resources without any active optimizer instances.
2. **Automatic restart**: For each orphaned resource, call `container.requestResource(resource)` to restart the optimizer process.
3. **Retry tracking**: Track the number of restart attempts per orphaned resource. If a restart exceeds the configurable maximum retries, clean up the orphaned resource from the database instead of retrying indefinitely.

#### Configuration
Two new configuration options are added (disabled by default):

| Config Key | Default | Description |
|---|---|---|
| `optimizer.auto-restart-enabled` | `false` | Whether to enable automatic restart of unexpectedly down optimizers |
| `optimizer.auto-restart-max-retries` | `5` | Maximum restart attempts per orphaned resource before cleanup |

#### Bug fix
Also fixes `ResourceMapper.selectResourcesByGroup` result mapping where property names did not match `Resource` class fields (`group`→`groupName`, `container`→`containerName`, `totalMemory`→`memoryMb`).

### Checklist
- [x] I have added corresponding tests for my changes.
- [x] This PR only changes one thing, and it is clear from the title.
- [ ] I have documented my changes in the relevant documentation.

Closes #3988

PR #4185, opened 2026-04-16
[AMORO-4163][ams] Fix Version-N-already-exists race in newTableOperations for legacy mixed-iceberg

lintingbin:

Follow-up fix for [AMORO-4163](https://github.com/apache/amoro/pull/4182). The original fix changed `ops.current()` to `ops.refresh()` inside `InternalMixedIcebergHandler.newTableOperations()` so that the returned `TableMetadata` reference satisfies Iceberg 1.7.x's reference-equality check in `commit()`. However, this introduced a new race condition:

`DefaultTableManager.createTable()` consists of two sequential steps:
1. `catalog.createTable()` — saves the table to the database (the table is now visible to all threads)
2. `onTableCreated()` → `triggerTableAdded()` → `newTableOperations()` → `ops.commit()` — tries to write `v2.metadata.json`

The background `tableExplorerScheduler` runs `exploreInternalCatalog()` concurrently. If it scans the database between steps 1 and 2, it finds the new table (not yet in `tableRuntimeMap`) and also calls `triggerTableAdded()` → `newTableOperations()` → `ops.commit()`. Whichever thread wins writes `v2.metadata.json` first; the other then fails with:

```
CommitFailedException: Version 2 already exists: .../base/metadata/v2.metadata.json
```

**Fix**: catch `CommitFailedException` in the commit block, refresh to obtain the latest on-disk metadata, and proceed normally if a concurrent writer already committed the correct mixed-format properties. Re-throw only when the properties still differ after refresh, indicating a genuine conflict.

## How was this patch tested?

The existing `CompatibilityCatalogTests.testNewCatalogLoadHistorical` in `TestInternalMixedCatalogService` covers this scenario — it was consistently failing in CI before this fix.

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/internal/InternalMixedIcebergHandler.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/internal/InternalMixedIcebergHandler.java
@@ -32,6 +32,7 @@ import org.apache.amoro.utils.TablePropertyUtil;
 import org.apache.hadoop.fs.Path;
 import org.apache.iceberg.TableOperations;
 import org.apache.iceberg.catalog.TableIdentifier;
+import org.apache.iceberg.exceptions.CommitFailedException;
 
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


czy006 (approved):

lintingbin:
Gentle nudge — this fix also unblocks CI on #4180.

The [recent CI run on #4180](https://github.com/apache/amoro/actions/runs/24772466254) fails with the exact pattern described in this PR's body:

```
TestInternalMixedCatalogService$CompatibilityCatalogTests.testNewCatalogLoadHistorical(boolean)[2]
  → IllegalStateException: update table meta failed
    at BasicMixedCatalog$MixedTableBuilder.createTableMeta(BasicMixedCatalog.java:404)
```

And the cascading symptoms match too — once the commit race fails once, `@After` can't drop `test_ns`, so every subsequent nested class (`TestTableCommit`, `TestTableOperation`, `TestDatabaseOperation`) fails at `.before:259` / `:186` / `:168` with `Database test_ns already exists`.

Same pattern showed up on a recent master hadoop3 run ([#24488461765](https://github.com/apache/amoro/actions/runs/24488461765), 2026-04-16), confirming it's pre-existing flakiness rather than regressions introduced by the feature PRs that hit it.

Would be great to get this merged so downstream CI stabilizes.

PR #4174, opened 2026-04-10
[Subtask]: Optimizations and troubleshooting for the Master-Slave mode. 

wardlican:
## Why are the changes needed?

Close #4171 .

## Brief change log
1. For existing historical tables, if Master-Slave mode is enabled, a `bucket_id` must be assigned to them.
2. Fixed an issue where concurrent addition of new tables could result in multiple tables being assigned the same `bucket_id`, leading to bucket imbalance.
3. Fixed an issue in Master-Slave mode where the `-msm` flag was not passed during the optimizer's startup.
4. Fixed an issue where tables were erroneously deleted following a migration.

-

## How was this patch tested?

- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (yes / no)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #4197, opened 2026-04-30
[AMORO-4198][optimizer] Support graceful shutdown for in-progress tasks

### LINKED ISSUE #4198: [Improvement]: Support graceful shutdown for in-progress optimizer tasks
## Search before asking

## What would you like to be improved?

When an optimizer process receives SIGTERM (e.g., K8s pod termination, rolling restart), in-progress tasks are silently dropped and AMS later re-schedules the same task — doubling work and potentially causing duplicate commits when the original task was already past its commit point.

This happens due to three compounding root causes:

1. **Dropped results**: `stopOptimizing()` only flips a stopped flag, causing any subsequent `completeTask()` call to be silently skipped by the gated retry loop.
2. **Signal not reaching the JVM**: The optimizer container runs as `sh -c <args>`, leaving `sh` as PID 1. SIGTERM from K8s is delivered to `sh` and never forwarded to the Java process, so the pod is killed by SIGKILL before graceful shutdown can run.
3. **Shutdown ordering**: Hadoop's `FileSystem` cache cleanup and the graceful shutdown hook run concurrently via JVM `Runtime` shutdown hooks, causing in-flight HDFS writers to hit `ClosedChannelException` during row-group flush.

## How should we improve?

- **Drain in-progress tasks**: `stopOptimizing()` should wait for executor threads to finish up to a configurable timeout (`--shutdown-timeout-ms`, default 10 min), keeping the toucher alive during the drain so AMS heartbeats continue.
- **Best-effort result reporting**: After shutdown is requested, `completeTask()` should fall back to a single direct call instead of silently dropping the result.
- **Signal delivery**: Wrap the container command with `exec` so the JVM replaces the shell and receives SIGTERM directly. Apply to both `KubernetesOptimizerContainer` and `optimizer.sh start-foreground`.
- **Shutdown ordering**: Register the graceful shutdown hook on Hadoop's `ShutdownHookManager` with higher priority than `FS_CACHE` (and `SPARK_CONTEXT_SHUTDOWN_PRIORITY` for Spark), with an explicit timeout that matches `shutdown-timeout-ms`.
- **K8s grace period**: Derive `terminationGracePeriodSeconds` automatically from `shutdown-timeout-ms` plus a buffer so the pod is not SIGKILL'd before the drain completes.

## Code of Conduct

j1wonpark:

Close #4198.

When an optimizer receives SIGTERM, in-progress tasks are silently dropped and
AMS re-schedules them — doubling work and potentially causing duplicate commits.

## Brief change log

- `Optimizer.stopOptimizing()`: join executor threads up to `--shutdown-timeout-ms`
  (default 10 min); keep toucher alive during drain so AMS heartbeats continue
- `OptimizerExecutor.completeTask()`: best-effort direct call after shutdown so
  results are not silently dropped
- `OptimizerToucher.stop()`: interrupt runner thread to wake it from sleep immediately
- `AbstractOptimizerOperator.waitAShortTime()`: preserve interrupt flag
- `OptimizerConfig`: new `-st` / `--shutdown-timeout-ms` option
- `StandaloneOptimizer` / `SparkOptimizer`: register graceful shutdown hook on
  Hadoop's `ShutdownHookManager` above `FS_CACHE` priority with explicit timeout
- `KubernetesOptimizerContainer`: `exec` prefix in container command; derive
  `terminationGracePeriodSeconds` from shutdown-timeout-ms + 30s buffer
- `optimizer.sh start-foreground`: `exec $CMDS` so Java receives SIGTERM directly

## How was this patch tested?

## Documentation

- If yes, how is the feature documented? JavaDocs / option `usage` string

## Review updates

- Wired the `shutdown-timeout-ms` optimizer group property to the `-st` startup arg and the K8s grace period (single typed source; the CLI re-parser is removed), and documented it. (@czy006 review point 1)
- Added a graceful drain to the Flink optimizer: `FlinkExecutor.close()` waits up to the shutdown timeout, capped below Flink's `task.cancellation.timeout`. (@czy006 review point 2)
- Hardened the new shutdown paths: safe publication of executor threads, force-interrupt reaching all executors, no busy-spin on stray interrupts, `completeTask` retrying through a bounded drain window, and no toucher re-registration while draining.

## Known limitations (proposed follow-ups)

- **Spark container release path**: `spark-submit --kill` does not extend the driver pod's termination grace period (K8s default 30s applies); operators can work around it via `spark-conf.spark.kubernetes.appKillPodDeletionGracePeriod`.
- **Local container release path**: `LocalOptimizerContainer.releaseResource()` uses `kill -9`, which skips JVM shutdown hooks; graceful drain only applies to externally sent SIGTERM.


j1wonpark:
Gentle ping for review 🙏 @zhoujinsong @czy006 — this builds on the master-slave optimizer work you reviewed in #4174 / #3937, and you both merged `master` in earlier. Just rebased onto latest `master` as a single clean commit. Key bits: heartbeat/shutdown ordering in `Optimizer.stopOptimizing`, the best-effort `completeTask` path after stop, and the `ShutdownHookManager` priorities. Thanks!

czy006 (commented):
Thanks for your contributor cc @j1wonpark 

- **--shutdown-timeout-ms** is added as a configurable parameter, but AMS does not append **shutdown-timeout-ms** from resource/container properties when generating optimizer startup arguments. Although K8s terminationGracePeriodSeconds parses this value from the startup arguments, the current AMS-managed K8s optimizer effectively can only use the default 600s + 30s buffer.
- Although the **Flink optimizer** inherits the shared stopOptimizing() / completeTask() changes, it does not register an equivalent drain hook for external termination. I suggest adding this part.

j1wonpark:
Thanks for the review @czy006! Both points are addressed:

**1.** Fixed in 99367db3 — the optimizer group property `shutdown-timeout-ms` is now appended as `-st` in the startup args (same pattern as `-hb`), and the K8s `terminationGracePeriodSeconds` is derived from the same property instead of re-parsing the CLI string, so the two can never diverge. Documented in `managing-optimizers.md`.

**2.** Added in 21d75def — `FlinkExecutor.close()` now drains the in-flight task up to the shutdown timeout (capped below `task.cancellation.timeout` to avoid failing the TaskManager) before force-interrupting. Since Flink cancels the toucher source before the operators, the drain also keeps the registration alive by touching AMS with the existing token at the heartbeat interval (e34d2bc2) — otherwise AMS would expire the optimizer after `optimizer.heart-beat-timeout` and reset the in-flight task mid-drain. It only touches, never re-registers, so once `close()` returns AMS cleans up through the normal expiration path.

I also fixed several races I found in the new shutdown paths (b4e5ae37).

czy006 (approved):
LGTM

PR #4200, opened 2026-05-01
[AMORO-1503][mixed-hive] Add INFO logs for partition operations

lintingbin:
## Summary

Closes #1503.

Adds INFO/WARN logs on the Mixed Hive Table partition operation paths
(create / drop / alter location). These operations are infrequent, so
logging them at INFO level makes after-the-fact diagnosis of missing
or mis-located partitions much easier with negligible runtime cost.

Logs are added at two layers:

- `HivePartitionUtil` — per-partition `LOG.info` on `createPartitionIfAbsent`,
  `dropPartition`, and `updatePartitionLocation`, plus matching `LOG.warn` on
  the failure paths. The existing `alterPartition` log now reports both old
  and new locations.
- `UpdateHiveFiles` / `ReplaceHivePartitions` — per-batch `LOG.info` for the
  drop / create / alter sets emitted by `commitPartitionedTable`, with a
  count and a sampled list of partitions (capped at 5, with `...` for the
  rest) to avoid log explosion on large commits, plus matching `LOG.warn`
  on commit failures.

Each log line carries the table identifier, partition values, location
(old → new on alter), and transaction id where available, so a single
log record is enough to identify the partition and trace it back to a
commit.

## Sample logs

Per-partition (single-partition path, `HivePartitionUtil`):

```
INFO  o.a.a.h.utils.HivePartitionUtil - Creating Hive partition for table catalog.db.tbl, partition values [2024-01-01], location s3://bkt/db/tbl/dt=2024-01-01/.amoro_xact_42
INFO  o.a.a.h.utils.HivePartitionUtil - Dropping Hive partition for table catalog.db.tbl, partition values [2024-01-01], location s3://bkt/db/tbl/dt=2024-01-01/.amoro_xact_41
INFO  o.a.a.h.utils.HivePartitionUtil - Altering Hive partition location for table catalog.db.tbl, partition dt=2024-01-01, location s3://bkt/db/tbl/dt=2024-01-01/.amoro_xact_41 -> s3://bkt/db/tbl/dt=2024-01-01/.amoro_xact_42
```

Per-batch (transaction commit path, `UpdateHiveFiles`):

```
[...]

PR #4201, opened 2026-05-01
[ams][test] Make RestCatalogService nested tests order-independent

lintingbin:

`amoro-ams/src/test/java/org/apache/amoro/server/RestCatalogServiceTestBase` and its subclasses (`TestInternalMixedCatalogService`, `TestInternalIcebergCatalogService`) share a single static `AmsEnvironment` and a hard-coded `database = \"test_ns\"` / `table = \"test_iceberg_tbl\"` across every `@Nested` class. Whenever cleanup leaks (the existing `@AfterEach` swallows `dropDatabase` exceptions) or background catalog scans race the tests, sibling nested classes inherit dirty catalog state and assertions that hard-code the catalog being empty start failing.

This PR makes the tests order-independent without restructuring the suite:

1. **Unique database name per test instance** — `RestCatalogServiceTestBase.database` is now `\"test_ns_\" + COUNTER.incrementAndGet()`. Each test method gets a fresh test instance (JUnit 5 default `Lifecycle.PER_METHOD`), so each test gets a fresh database name. Pollution between tests can no longer cause `AlreadyExists` on `createDatabase` or pre-populated catalog snapshots.
2. **Delta-based catalog/namespace assertions** in `TestInternalMixedCatalogService.TestDatabaseOperation.test` and `TestInternalIcebergCatalogService.NamespaceTests.testNamespaceOperations`. The previous code asserted `catalog.listDatabases().isEmpty()` which only holds in a single-test-class-per-JVM execution.

These tests have been flaky on `master` for a while — for example, the `Core/hadoop3 CI` run on 2026-04-23 failed inside `TestInternalIcebergCatalogService\$CatalogPropertiesTest`'s suite, and PR #4200's CI just hit the same family of failures, all surfacing the same root cause:

```
TestInternalMixedCatalogService\$TestDatabaseOperation.test:168
    expected: <true> but was: <false>          // listDatabases().isEmpty() failed because a previous test leaked test_ns

TestInternalMixedCatalogService\$TestTableOperation.before:186
    AlreadyExists Database: test_ns already exists  // previous @AfterEach swallowed dropDatabase failure

TestInternalMixedCatalogService\$TestTableCommit.testTableCommit:295
    NoSuchTableException: ...test_iceberg_tbl not exists  // sibling test's @AfterEach.dropTable raced into our just-created table
```

The shared `test_ns` / `test_iceberg_tbl` names are the load-bearing reason these races collide. With unique per-instance database names, every test gets its own namespace, so concurrent or leaked cleanup from a sibling test no longer touches it.

This is complementary to #4185, which fixed a `Version-N-already-exists` race in `InternalMixedIcebergHandler.newTableOperations` triggered by the same shared-state pattern.

## How was this patch tested?

Ran the affected suites locally on Java 11:

```
./mvnw test -pl amoro-ams -am -Pspark-3.5,support-all-formats,openapi-sdk \\
  -Dtest='TestInternalMixedCatalogService,TestInternalIcebergCatalogService'
# Tests run: 16, Failures: 0, Errors: 0, Skipped: 0
```

`./mvnw spotless:check checkstyle:check -pl amoro-ams -am` is clean.

## Does this PR introduce any user-facing change?

No — test-only change.

### DIFF
--- a/amoro-ams/src/test/java/org/apache/amoro/server/RestCatalogServiceTestBase.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/RestCatalogServiceTestBase.java
@@ -43,13 +43,16 @@ import org.junit.jupiter.api.BeforeEach;
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
 
 public abstract class RestCatalogServiceTestBase {
 
   static AmsEnvironment ams = AmsEnvironment.getIntegrationInstances();
   static String restCatalogUri = RestCatalogService.ICEBERG_REST_API_PREFIX;
 
-  protected final String database = "test_ns";
+  private static final AtomicLong TEST_INSTANCE_COUNTER = new AtomicLong();
+
+  protected final String database = "test_ns_" + TEST_INSTANCE_COUNTER.incrementAndGet();
   protected final String table = "test_iceberg_tbl";
   protected final TableIdentifier tableIdentifier =
       TableIdentifier.of(catalogName(), database, table);
--- a/amoro-ams/src/test/java/org/apache/amoro/server/TestInternalIcebergCatalogService.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/TestInternalIcebergCatalogService.java
@@ -104,12 +104,18 @@ public class TestInternalIcebergCatalogService extends RestCatalogServiceTestBas
   public class NamespaceTests {
     @Test
     public void testNamespaceOperations() throws IOException {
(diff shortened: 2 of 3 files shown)


lintingbin:
Hi @zhoujinsong, would you mind taking a look when you have time? Thanks!

This is test-only and is meant to fix the flaky `TestInternalMixedCatalogService` / `TestInternalIcebergCatalogService` failures that have been intermittently breaking CI on `master` and on #4200. Complementary to your #4185 — same test suite, but a different (test-isolation) flavor of the same shared-state pattern.

Issue #4203, opened 2026-05-07
[Improvement]: Remove JUnit 4 from the repository


lintingbin:
### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/amoro/issues?q=is%3Aissue) and found no similar issues.

### What would you like to be improved?

Remove JUnit 4 (`org.junit.*`) from the entire repository so the parent POM can drop `junit:junit` and `junit-vintage-engine`. Currently ~246 test files still depend on JUnit 4, including ~89 `@RunWith(Parameterized.class)` classes.

This issue tracks the umbrella effort across all modules. The original #3974 was scoped only to `amoro-common`, but the shared test bases there (`TestAms`, `TestHMS`, `AmoroCatalogTestBase`, `TestAmoroCatalogBase`, `TestServerTableDescriptor`, `AmoroRunListener`) are extended or used as `@Rule`/`@ClassRule`/`RunListener` by tests in many other modules. Migrating those bases in isolation breaks downstream modules — that is what blocked #4004 and #4059. So the migration has to happen across all modules in coordinated steps and needs a single umbrella ticket to track ordering.

### How should we improve?

Strategy: **leaves first, root last** — migrate concrete test classes module by module, keeping the shared `amoro-common` helpers on JUnit 4 until every consumer is gone. CI stays green at every step. Each step is a separate PR with body `Closes part of #THIS_ISSUE`.

- [x] **Step 1** — `amoro-common` self-contained tests (#4199)
- [x] **Step 2** — `amoro-optimizer-common` (5 files, smallest leaf, validates the pattern with reviewers) (#4204)
- [ ] **Step 3** — `amoro-format-iceberg` (~55 files; first time exercising the `@RunWith(Parameterized.class)` → `@ParameterizedTest @MethodSource` rewrite, plus iceberg-side consumers of `TestAmoroCatalogBase`/`TestServerTableDescriptor`) (#4205)
- [ ] **Step 4** — `amoro-format-mixed-hive` (~23 files)
- [ ] **Step 5** — `amoro-format-mixed-spark` 3.3 / 3.4 / 3.5 (~13 files)
- [ ] **Step 6** — `amoro-format-mixed-flink-common` (~52 files; converts the `AmoroRunListener` surefire wiring to a JUnit 5 `TestExecutionListener` at the same time) (#4206)
[...]

PR #4204, opened 2026-05-07
[AMORO-4203][optimizer-common] Migrate tests from JUnit 4 to JUnit 5

lintingbin:
## What is this PR for?

Closes part of #4203 (Step 2 — `amoro-optimizer-common`, the smallest leaf module).

Migrates the 5 test files in `amoro-optimizer-common` to JUnit Jupiter:

- `OptimizerTestBase`
- `TestOptimizer`
- `TestOptimizerToucher`
- `TestOptimizerExecutor`
- `TestOptimizerConfig`

## How TestAms is consumed without migrating it yet

`OptimizerTestBase` previously held `@ClassRule public static TestAms TEST_AMS = new TestAms();`. `TestAms` extends `org.junit.rules.ExternalResource` and is shared across many modules, so per the staged plan in #4203 it stays on JUnit 4 until the very last (Step 9) PR.

To unblock this step, the base class now invokes `TEST_AMS.before()` from `@BeforeAll` and `TEST_AMS.after()` from `@AfterAll`. Both methods are already `public` on `TestAms`, so no reflection or wrapper is needed and the lifecycle is identical to what `@ClassRule` produced. Once #4203 reaches Step 9, `TestAms` will move to a JUnit 5 `Extension` and these direct calls can be replaced with `@RegisterExtension` for free.

## Other mechanical changes

- `org.junit.*` → `org.junit.jupiter.api.*`
- `@Before` / `@After` → `@BeforeEach` / `@AfterEach`; `@BeforeClass` → `@BeforeAll`
- `Assert.assertX(...)` → `Assertions.assertX(...)`
- `TestOptimizerConfig`: three `@Test(expected = CmdLineException.class)` rewritten as `Assertions.assertThrows(CmdLineException.class, () -> ...)`
- `TestOptimizer`: dropped its duplicate `@BeforeClass reduceCallAmsInterval()` because the parent already calls `OptimizerTestHelpers.setCallAmsIntervalForTest()` from `@BeforeAll`

## Type of change

- [x] Improvement (test infrastructure)

## How was this patch tested?

`mvn -pl amoro-optimizer/amoro-optimizer-common -am test` → 10/10 tests pass under the JUnit Platform provider, `mvn spotless:check` clean.

### DIFF
--- a/amoro-optimizer/amoro-optimizer-common/src/test/java/org/apache/amoro/optimizer/common/OptimizerTestBase.java
+++ b/amoro-optimizer/amoro-optimizer-common/src/test/java/org/apache/amoro/optimizer/common/OptimizerTestBase.java
@@ -19,19 +19,25 @@
 package org.apache.amoro.optimizer.common;
 
 import org.apache.amoro.TestAms;
-import org.junit.Before;
-import org.junit.BeforeClass;
-import org.junit.ClassRule;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
 
 public abstract class OptimizerTestBase {
-  @ClassRule public static TestAms TEST_AMS = new TestAms();
+  protected static final TestAms TEST_AMS = new TestAms();
 
-  @BeforeClass
-  public static void reduceCallAmsInterval() {
+  @BeforeAll
+  public static void startTestAms() throws Exception {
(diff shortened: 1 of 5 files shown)


lintingbin:
@czy006 @xxubai CI is green. This is the smallest leaf in #4203 (5 files in `amoro-optimizer-common`) and serves as the pattern validator before fanning out — would appreciate a review when you have a chance.

xxubai (approved):

PR #4205, opened 2026-05-07
[AMORO-4203][format-iceberg] Migrate tests from JUnit 4 to JUnit 5

lintingbin:
## What is this PR for?

Closes part of #4203 (Step 3 — `amoro-format-iceberg`). 49 test files migrated to JUnit Jupiter, including 15 `@RunWith(Parameterized.class)` classes rewritten as `@ParameterizedTest @MethodSource` per @czy006's preference expressed on #4004.

## Dual-mode test bases (the central design decision)

`CatalogTestBase`, `TableTestBase`, and `TableDataTestBase` are still extended by ~19 JUnit 4 `@RunWith(Parameterized.class)` subclasses in `amoro-format-mixed-hive` and `amoro-ams` that Steps 4 and 8 of #4203 will migrate later. Migrating them now would either force a 200+ file mega-PR or drop into the same trap as #4004. Instead, this PR keeps the JUnit 4 surface intact and adds a parallel JUnit 5 surface:

- The original `(CatalogTestHelper, TableTestHelper)` constructor and `@Before setupCatalog()` / `@Before setupTable()` are preserved — JUnit 4 Parameterized callers continue to work unchanged.
- A no-arg constructor is added.
- New `protected void setupCatalog(CatalogTestHelper)` / `protected void setupTable(CatalogTestHelper, TableTestHelper)` entry points let JUnit 5 callers initialise the catalog/table state explicitly from their `@ParameterizedTest` method body.
- `@ClassRule TestAms` and `@Rule TemporaryFolder` are kept for the JUnit 4 path. For the JUnit 5 path we add `@BeforeAll startTestAms()` / `@AfterAll stopTestAms()` calling the public `before()` / `after()` of `TestAms` directly, and the `TemporaryFolder` is initialised manually inside `setupCatalog(helper)` because JUnit 4 `@Rule` does not fire under the JUnit Platform engine.

Cross-module compile is verified: `mvn -pl amoro-format-mixed-hive,amoro-ams -am test-compile` builds cleanly. All this dual plumbing is removed in the closing PR of #4203.

## Files intentionally kept on JUnit 4 in this PR

- `TestMixedCatalog`, `TestTaskReader`, `TestTaskWriter`, `TestTableTrashManagers` — same dual-mode reasoning, they are subclassed by `TestMixedHiveCatalog`/`TestHiveTaskReader`/`TestHiveTaskWriter`/`TestHiveTableTrashManagers` in mixed-hive.
[...]

PR #4004, opened 2025-12-09
[AMORO-3974] Migrate JUnit 4 to JUnit 5 in amoro-common module

Jungkihong07:
## Why are the changes needed?

JUnit 4 is no longer actively maintained, and JUnit 5 provides a more modern and powerful testing framework. 
This PR migrates all JUnit 4 tests in the `amoro-common` module to JUnit 5, enabling the use of the latest testing framework and reducing technical debt.

Close #3974.

### JUnit 4 → JUnit 5 Migration for amoro-common Module

**Migrated Test Files (9 files):**

1. `AmoroRunListener.java` - Converted `RunListener` → `TestExecutionListener`
2. `TestServerTableDescriptor.java` - Converted `@Rule`, `@ClassRule` → `@BeforeAll`/`@AfterAll` and `@TempDir`
3. `MemorySizeTest.java` - Converted `@Test(expected=...)` → `assertThrows()`
4. `JacksonUtilTest.java`
5. `TestSimpleFuture.java`
6. `TestAmoroCatalogBase.java`
7. `AmoroCatalogTestBase.java`
8. `TestPoolConfig.java`
9. `MockZookeeperServer.java`

**Key Changes:**

- Import changes: `org.junit.*` → `org.junit.jupiter.api.*`
- Annotation changes: `@Before/After` → `@BeforeEach/AfterEach`, `@BeforeClass/AfterClass` → `@BeforeAll/AfterAll` (static)
- Assertions changes: `Assert.*` → `Assertions.*`
- Exception testing: `@Test(expected=...)` → `Assertions.assertThrows()`
- Rules → Extensions: `@Rule` → `@RegisterExtension` or `@TempDir`
- RunListener → TestExecutionListener: `AmoroRunListener` conversion

**Detailed Implementation:**

- **TestAmoroCatalogBase.java**: Replaced all `Assert.*` calls with `Assertions.*`, updated imports from `org.junit.*` to `org.junit.jupiter.api.*`
- **AmoroCatalogTestBase.java**: Removed JUnit 4 annotations (`@Before`, `@After`) and related methods (`setupCatalogJUnit4()`, `cleanCatalogJUnit4()`), simplified to use only JUnit 5 lifecycle annotations (`@BeforeEach`, `@AfterEach`) with `@TempDir` for temporary directory management
[...]

PR #4206, opened 2026-05-07
[AMORO-4203][mixed-flink-common] Migrate tests from JUnit 4 to JUnit 5

lintingbin:
## What is this PR for?

Closes part of #4203 (Step 6 — `amoro-format-mixed/amoro-mixed-flink/amoro-mixed-flink-common`). 53 test files migrated to JUnit Jupiter, including 14 `@RunWith(Parameterized.class)` classes rewritten as `@ParameterizedTest @MethodSource`, plus the `AmoroRunListener` test execution listener migration.

## AmoroRunListener: JUnit 4 RunListener → JUnit 5 TestExecutionListener

`AmoroRunListener` was a JUnit 4 `RunListener` registered through this module's surefire `<listener>` property. JUnit Platform's surefire provider does not honour that property, so the class is rewritten to implement `org.junit.platform.launcher.TestExecutionListener` and registered via SPI.

Concretely:
- The class moves from `amoro-common/src/test/java/org/apache/amoro/listener/AmoroRunListener.java` to `amoro-format-mixed/amoro-mixed-flink/amoro-mixed-flink-common/src/test/java/...`. It had no other consumers in the repository — verified via grep — so the move keeps it co-located with its only user and lets the closing PR of #4203 leave `amoro-common` test sources untouched.
- The lifecycle hooks map cleanly: `testRunStarted/testRunFinished` → `testPlanExecutionStarted/Finished`, `testStarted/testFinished` → `executionStarted/Finished` filtered to `TestIdentifier#isTest()`. Method names come from `TestIdentifier#getDisplayName()` instead of `Description#getMethodName()`.
- A new SPI service file `src/test/resources/META-INF/services/org.junit.platform.launcher.TestExecutionListener` registers the listener for this module only (it stays opt-in elsewhere).
- `junit-platform-launcher` is added as an explicit `<scope>test</scope>` dependency in this module's pom — surefire pulls it in transitively at runtime, but the listener references the launcher API at compile time. The old `<properties><property><name>listener</name>…</property></properties>` block is removed; `<argLine>-verbose:class</argLine>` stays.

## Test base classes: clean migration, no dual-mode

Unlike #4205 (Step 3 in iceberg), the test bases in this module (`FlinkTestBase`, `AmoroCatalogITCaseBase`, `CatalogITCaseBase`, `TestMixedCatalog`) have **no external consumers** — verified via grep across the repository. So they migrate cleanly with no JUnit 4 surface.

**One non-mechanical structural decision worth your attention** (please push back if you'd prefer the alternative):

`FlinkTestBase`, `AmoroCatalogITCaseBase`, `CatalogITCaseBase`, and `TestMixedCatalog` no longer extend `org.apache.amoro.catalog.TableTestBase` / `CatalogTestBase` / `org.apache.amoro.formats.AmoroCatalogTestBase`. The catalog/table lifecycle is reimplemented inline against the same `CatalogTestHelper` / `TableTestHelper` / `AmoroCatalogTestHelper` contracts.

The reasons:
1. The parent classes pass helpers via constructor injection (`super(catalogTestHelper, tableTestHelper)`), which is incompatible with `@ParameterizedTest` — Jupiter doesn't inject test instances from method parameters into the constructor. Even with the dual-mode no-arg constructor that #4205 adds, every flink ITCase would still need to call `setupTable(c, t)` from each method — at which point inheriting the parent gives little leverage and adds a JUnit 4 dependency that #4203's closing PR would still have to clean up.
2. Once `LookupITCase` / `TestLookupSecondary` re-extended `FlinkTestBase` via the new `CatalogITCaseBase`, the inherited static helper `createKeyedTaskWriter(KeyedTable, RowType, boolean, long)` clashed with a same-signature default method on `FlinkTableTestBase`. The 4-arg static was renamed to `createKeyedTaskWriterWithMask`; its only external caller (`TestRowDataReaderFunction`) was updated. The 2-arg / 3-arg overloads keep their original names for source compatibility.

The cost of inlining is some duplicated setup boilerplate. The benefit is that flink-common is now JUnit 5-pure: the closing PR of #4203 will not need to touch this module at all.

## Parameterized rewrite (14 classes)

Standard template per czy006's preference on #4004:

```java
public class TestX extends FlinkTestBase {
  public static Stream<Arguments> parameters() {
    return Stream.of(Arguments.of(c1, t1), Arguments.of(c2, t2), ...);
  }

  @ParameterizedTest(name = "{0}, {1}")
  @MethodSource("parameters")
  public void testFoo(CatalogTestHelper c, TableTestHelper t) throws Exception {
    setup(c, t);
    // existing body
  }
}
```

Classes converted: `FlinkAmoroCatalogITCase`, `FlinkUnifiedCatalogITCase`, `TestMixedCatalog`, `TestKVTable`, `MixedIncrementalLoaderTest`, `TestRoundRobinShuffleRulePolicy`, `TestKeyed`, `TestTableRefresh`, `TestUnkeyed`, `TestUnkeyedOverwrite`, `TestAdaptHiveWriter`, `TestAutomaticLogWriter`, `TestFlinkSink`, `TestMixedFormatFileWriter`.

## Other mechanical changes

- `org.junit.*` → `org.junit.jupiter.api.*`
- `@Before` / `@After` → `@BeforeEach` / `@AfterEach`; `@BeforeClass` / `@AfterClass` → `@BeforeAll` / `@AfterAll`
- `Assert.assertX(...)` → `Assertions.assertX(...)` with corrected message-arg order (JUnit 4's leading-message position flips to trailing in JUnit 5)
- `@Rule TemporaryFolder` → `@TempDir Path`; `temp.newFolder()` → `Files.createTempDirectory(temp, "...").toFile()`
- `@Test(expected = X.class)` → `Assertions.assertThrows(X.class, () -> ...)`
- `org.junit.rules.TestName` → `TestInfo` parameter injection; `TestUtil.getUtMethodName(TestInfo)` updated (only flink-common consumers)
- Mockito: `@RunWith(MockitoJUnitRunner.class)` → `@ExtendWith(MockitoExtension.class)`
- `@Ignore` → `@Disabled`; `org.junit.Assume` → `Assumptions`

## Type of change

- [x] Improvement (test infrastructure)

## How was this patch tested?

- `mvn -pl amoro-format-mixed/amoro-mixed-flink/amoro-mixed-flink-common -am test` → **423 run, 0 failures, 0 errors, 102 skipped** in ~9:24. The 102 skipped are Hive-only parameter sets gated by `Assumptions.assumeTrue/False(...)` on macOS — same skip count as on master.
- `mvn -pl amoro-format-mixed/amoro-mixed-flink/amoro-mixed-flink-common spotless:check` clean.
- Sanity grep for leftover JUnit 4 imports / annotations in this module's test sources is empty.

### DIFF
--- a/amoro-format-mixed/amoro-mixed-flink/amoro-mixed-flink-common/pom.xml
+++ b/amoro-format-mixed/amoro-mixed-flink/amoro-mixed-flink-common/pom.xml
@@ -408,6 +408,17 @@
             <version>${assertj.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <!--
+          AmoroRunListener (registered via META-INF/services/org.junit.platform.launcher.TestExecutionListener)
+          requires the JUnit Platform launcher API at compile time. Surefire pulls it in transitively at
+          runtime, but it is not part of junit-jupiter, so we declare it explicitly.
+        -->
+        <dependency>
+            <groupId>org.junit.platform</groupId>
+            <artifactId>junit-platform-launcher</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
@@ -416,12 +427,6 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
-                    <properties>
-                        <property>
-                            <name>listener</name>
-                            <value>org.apache.amoro.listener.AmoroRunListener</value>
-                        </property>
-                    </properties>
                     <argLine>-verbose:class</argLine>
                 </configuration>
             </plugin>
--- a/amoro-format-mixed/amoro-mixed-flink/amoro-mixed-flink-common/src/test/java/org/apache/amoro/flink/FlinkTestBase.java
+++ b/amoro-format-mixed/amoro-mixed-flink/amoro-mixed-flink-common/src/test/java/org/apache/amoro/flink/FlinkTestBase.java
@@ -23,12 +23,19 @@ import static org.apache.amoro.flink.kafka.testutils.KafkaContainerTest.KAFKA_CO
 import static org.apache.flink.table.api.config.TableConfigOptions.TABLE_DYNAMIC_TABLE_OPTIONS_ENABLED;
 
 import org.apache.amoro.BasicTableTestHelper;
+import org.apache.amoro.MockAmoroManagementServer;
+import org.apache.amoro.TableFormat;
 import org.apache.amoro.TableTestHelper;
+import org.apache.amoro.TestAms;
+import org.apache.amoro.UnifiedCatalog;
+import org.apache.amoro.api.CatalogMeta;
 import org.apache.amoro.catalog.CatalogTestHelper;
-import org.apache.amoro.catalog.TableTestBase;
 import org.apache.amoro.flink.catalog.factories.CatalogFactoryOptions;
 import org.apache.amoro.flink.write.MixedFormatRowDataTaskWriterFactory;
 import org.apache.amoro.io.reader.GenericKeyedDataReader;
+import org.apache.amoro.mixed.CatalogLoader;
+import org.apache.amoro.mixed.MixedFormatCatalog;
+import org.apache.amoro.properties.CatalogMetaProperties;
 import org.apache.amoro.scan.CombinedScanTask;
 import org.apache.amoro.scan.KeyedTableScanTask;
 import org.apache.amoro.shade.guava32.com.google.common.collect.ImmutableList;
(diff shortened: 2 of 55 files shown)


lintingbin:
@czy006 @xxubai CI is green. This is the flink-common piece of #4203 — 53 files + the `AmoroRunListener` migration to a JUnit 5 `TestExecutionListener` (registered via SPI), 14 `@RunWith(Parameterized.class)` classes rewritten. One non-mechanical decision worth your eye is flagged in the PR body: `FlinkTestBase` / `AmoroCatalogITCaseBase` / `CatalogITCaseBase` no longer extend the still-JUnit-4 iceberg/amoro-common bases; the catalog/table lifecycle is reimplemented inline. Happy to refactor to keep the inheritance if you'd prefer — please push back either way.

lintingbin:
@xxubai Thanks for keeping the branch fresh with master 🙏. Is there anything blocking this one — particularly the non-mechanical design call to reimplement `FlinkTestBase` / `AmoroCatalogITCaseBase` / `CatalogITCaseBase` inline instead of extending the still-JUnit-4 iceberg/amoro-common bases, or the `AmoroRunListener` move + SPI registration? Happy to refactor in either direction if you'd prefer the alternative.

PR #4231, opened 2026-05-21
[ams] Force ReachableFileCleanup when snapshots exist outside main ancestry

j1wonpark:

When AMS expires snapshots on a single-ref Iceberg table, Iceberg auto-selects `IncrementalFileCleanup`. That strategy walks the current snapshot's ancestor chain via `SnapshotUtil.ancestorIds`, and if a parent snapshot is missing from metadata the walk stops silently. Snapshots below the break are then treated as non-ancestors, so the `ADDED` entries of the manifests they wrote get reverted during cleanup. If such a manifest was superseded by a merge but its data file is still carried over (as `EXISTING`) in a manifest the current snapshot references, cleanup deletes a file the live table still points at, and reads fail with `NotFoundException`.

This is the same root cause as [apache/iceberg#13568](https://github.com/apache/iceberg/issues/13568), fixed by [apache/iceberg#13614](https://github.com/apache/iceberg/pull/13614) in Iceberg **1.10.0**, where the strategy selector now falls back to `ReachableFileCleanup` when snapshots exist outside the main ancestry. Iceberg `1.6.x ~ 1.9.x` lack that guard and unconditionally pick the incremental strategy for a single ref. Since bumping Iceberg core isn't always possible for existing deployments, this PR backports an equivalent guard at the Amoro layer.

## Brief change log

- `IcebergTableMaintainer`: before committing `expireSnapshots`, check `hasSnapshotsOutsideMainAncestry()` — the same `SnapshotUtil.ancestorIds` walk the cleanup uses — and force `ReachableFileCleanup` only in that at-risk state. Healthy tables keep the automatic incremental selection, so there is no cost impact.
- `ReachableFileCleanupBridge` (new, in package `org.apache.iceberg`): `RemoveSnapshots#withIncrementalCleanup(boolean)` is package-private, so the bridge calls it without reflection.

## How was this patch tested?

- [x] Added a regression test (`TestExpireSnapshotsKeepReferencedFiles`) that reproduces the carry-over + broken-chain state and asserts referenced data files survive expiration.

## Documentation


j1wonpark:
Hi @xxubai @zhoujinsong — when you have a moment, could you take a look at this PR?

This is a small, targeted backport of [apache/iceberg#13614](https://github.com/apache/iceberg/pull/13614) (shipped in Iceberg 1.10.0) to the Amoro layer, so users on Iceberg 1.6.x–1.9.x are protected from the silent data-file deletion when snapshots exist outside the main ancestry. The fix only adds a pre-commit check; healthy tables keep the existing `IncrementalFileCleanup` path, so there's no cost impact for the common case. A regression test (`TestExpireSnapshotsKeepReferencedFiles`) is included.

Happy to address any feedback. Thanks!

xxubai (approved):
Thanks for the contribution. Great work on this patch!

PR #4247, opened 2026-06-10
[AMORO-4246] Refactor process data expiring via ProcessFactory plugin

zhangwl9:
## Why are the changes needed?

Close #4246 .

## Brief change log

- Add `ProcessDataExpiringProcess` as a `TableProcess` implementation for expiring process runtime data and history data
- Register `expire-process-data` action in `IcebergProcessFactory` with configurable trigger interval and keep-time parameters
- Move configuration from `AmoroManagementConf` to `IcebergProcessFactory` under `expire-process-data.*` namespace (Duration type)
- Remove `ProcessDataExpiringExecutor` and `InlineTableExecutors`
- Update `process-factories.yaml` and `deployment.md` documentation
- Migrate tests from `TestProcessDataExpiringExecutor` to `TestProcessDataExpiringProcess` with identical test coverage

## How was this patch tested?

- [x] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (yes / no)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #4252, opened 2026-06-12
Revert "[AMORO-4246] Refactor process data expiring via ProcessFactory plugin"

zhoujinsong:
Reverts apache/amoro#4247

**Why revert?**

After further consideration,  ⁠ProcessDataExpiring⁠  is an AMS-internal housekeeping task (cleaning up stale process runtime/history data). It does not fit the  ⁠TableProcess⁠  model, which is designed for user-facing table maintenance operations that benefit from:
 ⦁State tracking and history viewing
 ⦁Visibility in the dashboard
 ⦁Potential external scheduling
Internal cleanup tasks like this should remain in the inline executor pattern. The  ⁠TableProcess⁠  /  ⁠ProcessFactory⁠  plugin model should be reserved for operations where tracking, history, and external execution actually add value (e.g., expire-snapshots, clean-orphan-files).


xxubai (approved):
+1

Issue #4191, opened 2026-04-19
[Feature]: Dynamic Resource Allocation for Optimizer


j1wonpark:
## Description

Add workload-aware dynamic resource allocation to Amoro's optimizer, enabling automatic scale-up when optimization demand exceeds capacity and scale-down when optimizers are idle. Inspired by Spark's Dynamic Resource Allocation.

## Use case/motivation

Currently, Amoro's optimizer only supports a static `min-parallelism` floor — it scales up to meet the minimum but never scales down, and it does not respond to actual workload.

**1. Resource waste during low-load periods**: Optimizer pods remain running even when there are no optimization tasks for hours or days, consuming cluster resources that could serve other workloads.

**2. Manual intervention for load spikes**: When a burst of table optimizations is triggered (e.g., after a large batch ingestion), users must manually adjust `min-parallelism` — which is slow, error-prone, and operationally painful.

## Describe the solution

All scaling logic lives in AMS (`DefaultOptimizingService.OptimizerGroupKeeper`), making it container-agnostic. Container implementations (`KubernetesOptimizerContainer`, etc.) remain unchanged. Kubernetes is the primary target for testing and tuning.

```
                  ┌──────────────────────────────────┐
                  │   DefaultOptimizingService (AMS)  │
                  │                                   │
                  │   OptimizerGroupKeeper            │
                  │   ├─ min-parallelism (existing)   │
                  │   ├─ demand-driven scale-up       │
                  │   ├─ idle-driven scale-down       │
                  │   └─ graceful drain               │
                  └──────────────┬────────────────────┘
                                 │
                  requestResource() / releaseResource()
                                 │
           ┌─────────────────────┼─────────────────────┐
           │                     │                      │
     K8s Container ★       Flink Container        Spark Container
     (primary target)      (unchanged)            (unchanged)
```

### New configuration properties (ResourceGroup level, opt-in)

| Property | Default | Description |
|---|---|---|
| `dynamic-allocation.enabled` | `false` | Opt-in flag. Existing behavior fully preserved when disabled. |
| `max-parallelism` | `Integer.MAX_VALUE` | Upper bound on total threads |
| `dynamic-allocation.executor-idle-timeout` | `5m` | Remove an idle optimizer after this duration |
| `dynamic-allocation.scheduler-backlog-timeout` | `1m` | Trigger scale-up after sustained demand |
| `dynamic-allocation.sustained-backlog-timeout` | `30s` | Interval between subsequent scale-up rounds |
[...]

PR #4254, opened 2026-06-14
[AMORO-4253] AIP-5 Phase 1: dynamic allocation configuration and validation

### LINKED ISSUE #4253: [Subtask]: AIP-5 Phase 1 — Dynamic allocation configuration, validation, and min-parallelism deprecation
## Search before asking

## Description

This is the first implementation phase of **AIP-5: Dynamic Resource Allocation for Optimizer**. It lays the configuration foundation that the later scaling phases build on, with no runtime scaling behavior yet. Dynamic allocation is opt-in and disabled by default (`dynamic-allocation.enabled = false`), so existing groups are unaffected and this phase is safe to merge independently. The full `dynamic-allocation.*` property set is declared here so the configuration surface and its validation are settled in one place, rather than drip-fed per phase.

Scope:

1. **Configuration properties.** Declare the `dynamic-allocation.*` properties as constants in `OptimizerProperties`, each annotated with `@since`:
   - `dynamic-allocation.enabled` (default `false`)
   - `dynamic-allocation.min-parallelism` (default `0`)
   - `dynamic-allocation.max-parallelism` (required when enabled; hard-capped at `1024`)
   - `dynamic-allocation.scheduler-backlog-timeout` (default `1min`)
   - `dynamic-allocation.sustained-backlog-timeout` (default `30s`)
   - `dynamic-allocation.executor-idle-timeout` (default `5min`, minimum `30s`)
   - `dynamic-allocation.scale-down-cooldown` (default `1min`)
   - `dynamic-allocation.drain-timeout` (default `15min`)

2. **`DynamicAllocationConfig` parsing + validation.** A new `DynamicAllocationConfig.parse(ResourceGroup)` / `validate()` that enforces:
   - `max-parallelism` is mandatory when enabled;
   - `min-parallelism ≤ max-parallelism ≤ 1024`;
   - `executor-idle-timeout ≥ 30s`;
   - all durations parse to positive values;
   - DRA is not enabled on an externally-registered optimizer (conservative check in this phase: reject the `external` container).

3. **Failure handling.**
   - **REST create/update — fail fast.** `OptimizerGroupController.createResourceGroup` / `updateResourceGroup` reject an invalid DRA config with **HTTP 400**. (The update path previously had no property-value validation at all.)
   - **Startup load — fail safe, never silent.** A persisted group with an invalid DRA config no longer crashes AMS; it logs a warning and falls back to DRA-disabled behavior. (The `optimizer_group_config_invalid` gauge is wired in the observability phase.)

4. **`min-parallelism` deprecation.** The flat `min-parallelism` property is deprecated in favor of `dynamic-allocation.min-parallelism` but still honored. Resolution order: namespaced → legacy → `0`. A one-off deprecation warning is logged at config-entry points (startup load, REST), not on the keeper hot path.

5. **Auto-reset compatibility.** The keeper already auto-resets a group's `min-parallelism` after `optimizer-group.max-keeping-attempts` consecutive failed optimizer creations. Two adjustments keep this consistent with the new property model:
   - **Disabled when DRA is effectively enabled.** Once a group opts into dynamic allocation (and its config is valid), DRA owns the group's scale decisions, so the keeper no longer erodes its `min-parallelism` floor — it keeps retrying instead. "Effectively enabled" means opted-in *and* valid; an invalid config is treated as disabled, matching the startup fail-safe, so such groups still auto-reset as before.
   - **Writes the key the group actually uses.** When auto-reset does apply, it now writes whichever `min-parallelism` key the group reads from (namespaced if present, else the legacy flat key). Previously it always wrote the flat key; for a group using the namespaced property that write was shadowed by resolution order, turning auto-reset into an endless no-op loop (repeated warning + DB update with no effect).

Out of scope for this phase (later phases): demand-driven scale-up, idle tracking, scale-down + graceful drain, and the new metrics. The `config_invalid` gauge and `scale_*_total` counters land in the observability phase.

## Parent issue

AIP-5: https://github.com/apache/amoro/issues/4191

## Are you willing to submit PR?

## Code of Conduct

j1wonpark:

Close #4253.

This is the first implementation phase of **AIP-5: Dynamic Resource Allocation for Optimizer** (#4191). It lays the configuration foundation that the later scaling phases build on. There is **no runtime scaling behavior in this PR** — it is purely opt-in configuration plumbing, so it is safe to merge independently and has no effect on existing groups (`dynamic-allocation.enabled` defaults to `false`).

Splitting AIP-5 into self-contained PRs follows the design's five linear phases; demand-driven scale-up, idle tracking, scale-down + graceful drain, and the new metrics land in subsequent PRs. The full set of `dynamic-allocation.*` properties is declared here (rather than drip-fed per phase) so the public configuration surface and its validation are settled in one place: a group created between phases cannot carry an unvalidated DRA property. A few getters (the timeout durations, `max-parallelism`) therefore have no caller until the phase that consumes them — this is intentional, not dead code.

## Brief change log

- **`OptimizerProperties`**: declared the `dynamic-allocation.*` constants (each with a `@since` tag) — `enabled`, `min-parallelism`, `max-parallelism` (+ the `1024` hard-cap constant), `scheduler-backlog-timeout`, `sustained-backlog-timeout`, `executor-idle-timeout` (+ its `30s` minimum), `scale-down-cooldown`, `drain-timeout`. Deprecated the flat `min-parallelism` in favor of the namespaced `dynamic-allocation.min-parallelism`; it is still honored as a fallback.

- **`DynamicAllocationConfig`** (new): `parse(ResourceGroup)` builds a typed, immutable config from the group's properties; `validate()` enforces the AIP-5 constraints **only when DRA is enabled** — `max-parallelism` is mandatory, `min-parallelism ≤ max-parallelism ≤ 1024` (an over-limit value is rejected, not silently clamped), `executor-idle-timeout ≥ 30s`, all durations positive, and DRA is rejected on an externally-registered optimizer (conservative check this phase: the `external` container, which AMS cannot scale). `resolveMinParallelism()` centralizes the namespaced → legacy → `0` resolution; a one-off deprecation warning is emitted at config-entry points, not on the keeper hot path.

- **`OptimizerGroupController`**: `createResourceGroup` / `updateResourceGroup` now reject an invalid DRA config with **HTTP 400** before persisting. (The update path previously had no property-value validation at all.)

- **`DefaultOptimizingService`**:
  - Startup load is **fail-safe, never silent**: a persisted group with an invalid DRA config no longer crashes AMS — it logs a warning and falls back to DRA-disabled behavior. (The `optimizer_group_config_invalid` gauge is wired in the observability phase.)
  - `getMinParallelism()` now delegates to `resolveMinParallelism()`, preserving the existing lenient behavior while honoring the namespaced key.
  - The keeper's existing `min-parallelism` auto-reset (after `optimizer-group.max-keeping-attempts` failed creations) is adjusted for the new property model, per the AIP's Compatibility section:
    - It is **skipped when DRA is effectively enabled** (opted-in *and* valid) — DRA owns the group's scale decisions, so the keeper no longer erodes its `min-parallelism` floor and keeps retrying instead. An invalid config counts as disabled, matching the startup fail-safe.
    - When it does apply, it now writes **whichever key the group actually reads** (namespaced if present, else the legacy flat key). Previously it always wrote the flat key; for a group using the namespaced property that write was shadowed by resolution order, turning auto-reset into an endless no-op loop (repeated warning + DB update with no effect). This is fixed here.

## How was this patch tested?

  - `TestDynamicAllocationConfig`: rejection of enabled-without-`max-parallelism`, `max < min`, over-`1024`, sub-`30s` idle timeout, unparsable duration, zero/non-positive duration, and the `external` container; correct parsing of a valid config; the namespaced → legacy → `0` resolution order; `isEffectivelyEnabled` (valid+enabled → true, disabled → false, enabled+invalid → false) and `effectiveMinParallelismKey`.
  - `TestOptimizerGroupController`: invalid create/update return 400, valid create succeeds.
  - `TestOptimizerGroupKeeper`: auto-reset is skipped when DRA is enabled (the floor is preserved); a group using the namespaced key has the reset written to that key and no deprecated flat key is introduced; the existing legacy-group auto-reset scenarios still pass (no regression).

- [ ] Add screenshots for manual tests if appropriate (not applicable — no UI change)

- [x] Run test locally before making a pull request (affected `amoro-common` + `amoro-ams` tests pass — 35 tests green; spotless and checkstyle clean)

## Documentation

- Does this pull request introduce a new feature? (yes — AIP-5 foundation, opt-in and disabled by default)
- If yes, how is the feature documented? (JavaDocs — `@since` tags and `DynamicAllocationConfig` doc comments; the full design is in AIP-5 / #4191, and user-facing configuration docs land with the phase that first exposes runtime behavior)


j1wonpark:
Hi @xxubai @czy006 @zhoujinsong — first PR for AIP-5 (#4191), config-only and disabled by default; would appreciate a review on whether the dynamic-allocation.* config surface and validation look right before later phases build on them. Thanks!

xxubai:
Thanks for your contribution @j1wonpark ! Will take a look soon

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/dra/DynamicAllocationConfig.java
```diff
@@ -0,0 +1,362 @@
...
+   */
+  public static DynamicAllocationConfig parse(ResourceGroup group) {
+    Map<String, String> properties = group.getProperties();
+    boolean enabled =
+        PropertyUtil.propertyAsBoolean(
+            properties,
+            OptimizerProperties.DYNAMIC_ALLOCATION_ENABLED,
+            OptimizerProperties.DYNAMIC_ALLOCATION_ENABLED_DEFAULT);
+    int minParallelism = resolveMinParallelism(group);
+    String minParallelismRaw = rawMinParallelism(group);
```
xxubai: Why need additional raw `min-parallelism`

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/dra/DynamicAllocationConfig.java
```diff
@@ -0,0 +1,362 @@
...
+    }
+    return properties.get(OptimizerProperties.OPTIMIZER_GROUP_MIN_PARALLELISM);
+  }
+
+  /**
+   * Log a one-off deprecation warning when a group still relies on the flat {@code
+   * min-parallelism}. Intended for config-entry points (startup load, REST create/update), not the
+   * keeper hot path.
+   */
+  public static void warnDeprecatedMinParallelism(ResourceGroup group) {
```
xxubai, line 176: `warnDeprecatedMinParallelism` can be included in `parse`

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -181,6 +182,18 @@ private void loadOptimizingQueues(List<DefaultTableRuntime> tableRuntimeList) {
     optimizerGroups.forEach(
         group -> {
           String groupName = group.getName();
+          // Fail-safe: a persisted group carrying an invalid DRA config (e.g. manual DB edits)
+          // must not crash AMS. Surface it and fall back to DRA-disabled behavior.
```
xxubai: Do you have a more implicit way to disable it? It doesn’t seem appropriate to validate the group's configuration when loading optimizing queues.

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/dra/DynamicAllocationConfig.java
j1wonpark: `min-parallelism` wasn't parsed strictly, so I added a guard for that — keeping both a lenient `int` and the raw `String` so `validate()` could tell "unset" from "explicitly set but malformed" (the lenient `int` collapses both to `0`). That turned out to be overkill.

Simplified by dropping the dual representation: `min-parallelism` is now a single nullable `Integer`, parsed strictly in `parse()` (mirroring `max-parallelism`), with `null` meaning "unset".

amoro-ams/src/main/java/org/apache/amoro/server/optimizing/dra/DynamicAllocationConfig.java
j1wonpark: The reason it's separate: `parse()` is also reached off the keeper path via `isEffectivelyEnabled()` (called in the auto-reset branch), so folding the warning into `parse()` would re-log the deprecation on every keeper cycle instead of once. `warnDeprecatedMinParallelism` is meant to fire only at config-entry points (startup load, REST create/update).

I could fold it in by splitting `parse()` into a warning wrapper + a warn-free core that `isEffectivelyEnabled()` uses — happy to do that if you prefer the single call site, though it trades one extra line at two callers for an extra private method. Let me know which you'd rather have.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
j1wonpark: The block was meant as a startup fail-safe — the AIP's "never silent" requirement, so a persisted invalid config would be surfaced rather than silently ignored.

But as you pointed out, validating inside the queue loader is the wrong place, and looking closer the block was behaviorally inert: the disable is already implicit (`isEffectivelyEnabled()` treats an invalid config as disabled wherever DRA is consumed), and in Phase 1 nothing strictly parses DRA config at load, so there was no crash to guard against either. Removed it — surfacing invalid configs as an alertable signal belongs with the metric that lands in the observability phase.

xxubai (approved):
LGTM overall. cc @zhoujinsong

czy006 (approved):
LGTM

PR #4275, opened 2026-07-17
[AMORO-4274] Remove Roaring64Bitmap.isEmpty() check in CombinedDelete…

### LINKED ISSUE #4274: [Improvement]: Remove Roaring64Bitmap.isEmpty() check in CombinedDeleteFilter
### Search before asking

`CombinedDeleteFilter` builds a per-file position-delete predicate. For every record it looks up the file path in `positionMap` and, when a `Roaring64Bitmap` is present, checks `posSet.isEmpty()` before `posSet.contains(...)`:

```java
Roaring64Bitmap posSet = positionMap.get(structLikeForDelete.filePath());
if (posSet == null || posSet.isEmpty()) {
  return false;
}
return posSet.contains(structLikeForDelete.getPosition());
```

The `isEmpty()` call is redundant and wastes CPU on the hot path (invoked once per scanned record):

1. Entries in `positionMap` are only created via `computeIfAbsent(...).add(...)` while reading position-delete files, so any value stored in the map is guaranteed to be a **non-empty** `Roaring64Bitmap`. `positionMap.get(...)` therefore returns either `null` (key absent) or a non-empty bitmap — never an empty one. The `isEmpty()` branch can never be true.
2. `Roaring64Bitmap.isEmpty()` is implemented on top of `getLongCardinality()`, which iterates over the whole set of high-to-low containers. It is not a cheap O(1) check, so paying it once per record is a measurable overhead on tables with large position-delete sets.

### How should we improve?

Drop the `isEmpty()` check and keep only the null guard:

```java
if (posSet == null) {
  return false;
}
```

Behavior is unchanged (the removed branch was unreachable given how `positionMap` is populated) while eliminating a per-record `getLongCardinality()` traversal.

### Subtasks

_No response_

### Code of Conduct

byungnam:

`CombinedDeleteFilter` builds a per-file position-delete predicate that runs once per scanned record. For each record it looks up the file path in `positionMap` and checks `posSet.isEmpty()` before `posSet.contains(...)`:

```java
Roaring64Bitmap posSet = positionMap.get(structLikeForDelete.filePath());
if (posSet == null || posSet.isEmpty()) {
  return false;
}
return posSet.contains(structLikeForDelete.getPosition());
```

The `isEmpty()` check is unnecessary and wastes CPU on the hot path:

1. `positionMap` is populated only via `computeIfAbsent(...).add(...)` while reading position-delete files, so every value stored in the map is a **non-empty** `Roaring64Bitmap`. `positionMap.get(...)` therefore returns either `null` (key absent) or a non-empty bitmap — never an empty one, so the `isEmpty()` branch is unreachable.
2. `Roaring64Bitmap.isEmpty()` is backed by `getLongCardinality()`, which traverses all high-to-low containers. It is not an O(1) check, so paying it once per record is measurable overhead on tables with large position-delete sets.

Close #4274

## Brief change log

- `amoro-format-iceberg` — `CombinedDeleteFilter`: in the position-delete predicate, drop the unreachable `posSet.isEmpty()` branch, keeping only the `posSet == null` guard.

## How was this patch tested?

- [x] Existing tests: covered by the existing delete-filter / reader tests in `amoro-format-iceberg`; behavior is unchanged because the removed branch was unreachable.

## Documentation

### DIFF
--- a/amoro-format-iceberg/src/main/java/org/apache/amoro/io/reader/CombinedDeleteFilter.java
+++ b/amoro-format-iceberg/src/main/java/org/apache/amoro/io/reader/CombinedDeleteFilter.java
@@ -415,7 +415,7 @@ public abstract class CombinedDeleteFilter<T extends StructLike> {
     return structLikeForDelete -> {
       Roaring64Bitmap posSet = positionMap.get(structLikeForDelete.filePath());
 
-      if (posSet == null || posSet.isEmpty()) {
+      if (posSet == null) {
         return false;
       }
       return posSet.contains(structLikeForDelete.getPosition());


zhoujinsong (approved):
LGTM.

Thanks a lot for the work!

PR #4278, opened 2026-07-20
[AMORO-4277] imporove expire_snapshots performance

### LINKED ISSUE #4277: [Improvement]: expire_snapshot may delete some parent directory many times
### Search before asking

RollingFileCleaner::doCleanFiles deletes files in batches, consider this scenario: 

1. the first-level partition is daily, and the second-level partition uses bucket(70000, id). 
2. After I expire data for a certain day, when the snapshot expires, it will clean up the files under the 70,000 second-level partitions for that day. 
3. Since the deletion is batched, after each bucket partition's files are deleted, a deleteEmptyDirectory operation will be executed on the daily partition.

Image

### How should we improve?

1.remove deleteParentDirectories logical from doCleanFiles
2.  add cleanParentDirectories() method.
3. call it in method clear()

### Subtasks

_No response_

### Code of Conduct

Aireed:

Close #4277 .

## Brief change log

- call  docleanEmptyDirectory once

## How was this patch tested?

## Documentation


amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java
```diff
@@ -27,8 +27,11 @@
 import org.slf4j.LoggerFactory;
 
 import java.net.URI;
+import java.util.Comparator;
```
zhoujinsong: `amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java` has no references anywhere in the codebase. It is a stale duplicate of `amoro-format-iceberg/.../RollingFileCleaner.java` which is the one actually used. Instead of modifying it here, please delete it to avoid maintaining two copies.

Maybe move the unit test code to model `amoro-format-iceberg` too.

zhoujinsong (commented):
Thanks a lot for the work!

I left a minor suggestion.

amoro-ams/src/main/java/org/apache/amoro/server/utils/RollingFileCleaner.java
Aireed: 3ks for your review, i have fixed it . PTAL again

zhoujinsong (approved):
LGTM.

Thanks a lot for the work!

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/utils/RollingFileCleaner.java
```diff
@@ -110,27 +113,34 @@ private void doCleanFiles() {
...
       LOG.debug("Cleaned expired a file group, total files: {}", collectedFiles.size());
     } finally {
-      // Always drop both buffers so a failure in directory cleanup does not cause the
-      // already-deleted files to be re-submitted on the next batch.
-      parentDirectories.clear();
       collectedFiles.clear();
     }
   }
 
+  public void doCleanParentDirectory() {
```
xxubai: nit: can you limit the method scope? such as `private`

amoro-format-iceberg/src/main/java/org/apache/amoro/utils/TableFileUtil.java
```diff
@@ -75,15 +76,24 @@ public static String getFileDir(String filePath) {
...
+    deleteEmptyDirectory(io, directoryPath, exclude, new HashSet<>());
+  }
+
   /**
    * Try to recursiveDelete the empty directory
    *
    * @param io mixed-format file io
    * @param directoryPath directory location
-   * @param exclude the directory will not be deleted
+   * @param exclude the directory will not be deleted @Param directoriesToBeDeleted: all the
```
xxubai:
```suggestion
   * @param exclude the directory will not be deleted 
   * @Param directoriesToBeDeleted: all the directories that need to be deleted
```

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/utils/RollingFileCleaner.java
Aireed: thanks for your review. i have fixed it. PTAL again

Issue #4240, opened 2026-06-01
[Improvement]: Iceberg REST Catalog compatibility issue with Trino: namespace location is not returned and CREATE TABLE reports non-empty location after table is created


xaoyun:
### Search before asking

- [x] I have searched in the [issues](https://github.com/apache/amoro/issues?q=is%3Aissue) and found no similar issues.

## 正文草稿

```markdown
Hi Amoro community,

We found two compatibility issues between Amoro Iceberg REST Catalog and Trino Iceberg connector.

These issues make Trino DDL unreliable: `CREATE TABLE` may return an error to the client, while the table is actually created successfully in Amoro and can be queried afterward.

## Environment

- Amoro: 0.6.1
- Object storage: MinIO / S3-compatible storage
- Query engine: Trino
- Catalog type: Iceberg REST Catalog
- Amoro REST Catalog URI:

```properties
iceberg.rest-catalog.uri=http://<amoro-host>:1630/api/iceberg/rest
iceberg.rest-catalog.warehouse=lakehouse_catalog
```

Trino catalog config example:

```properties
connector.name=iceberg
iceberg.catalog.type=rest
iceberg.rest-catalog.uri=http://<amoro-host>:1630/api/iceberg/rest
iceberg.rest-catalog.warehouse=lakehouse_catalog
iceberg.unique-table-location=true
iceberg.file-format=PARQUET

fs.native-s3.enabled=true
s3.endpoint=http://<minio-host>:9000
s3.region=us-east-1
s3.path-style-access=true
s3.aws-access-key=***
s3.aws-secret-key=***
```

## Issue 1: namespace properties are not preserved or returned

When creating a namespace with `location`:

```sql
CREATE SCHEMA iceberg_minio.dwd_stage_v2
WITH (
    location = 's3://enerpoch-pub-bucket/warehouse/dwd_stage_v2'
);
```

or calling the Iceberg REST API directly:

```bash
curl -s -X POST \
  "http://<amoro-host>:1630/api/iceberg/rest/v1/catalogs/lakehouse_catalog/namespaces" \
  -H "Content-Type: application/json" \
  -d '{
    "namespace": ["diag_stage_loc"],
    "properties": {
      "location": "s3://enerpoch-pub-bucket/warehouse/diag_stage_loc"
    }
  }'
```

Amoro returns:

```json
{
  "namespace": ["diag_stage_loc"],
  "properties": {}
}
```

Loading the namespace also returns empty properties:

```bash
curl -s \
[...]

PR #4283, opened 2026-07-21
[AMORO-4240][AMS] Support namespace properties in REST catalog

johntomcat7408-cmyk:
## Summary

This is the first of two PRs for #4240. It adds namespace properties support to the Iceberg REST catalog. Stage-create support is not included and will be handled separately.

`RestCatalogService` currently ignores `CreateNamespaceRequest.properties()`, does not return properties from `getNamespace()`, and does not support updating namespace properties. The `database_metadata` table also has no column for storing them.

## Changes

- Add a `properties` column to `database_metadata`.
- Add initialization and upgrade scripts for Derby, MySQL, and PostgreSQL.
- Persist properties passed to `createNamespace()`.
- Return properties from `createNamespace()` and `getNamespace()`.
- Implement the namespace properties update endpoint.
- Serialize concurrent property updates to avoid lost updates.
- Use the namespace `location` property as the default table location. An explicit table location still takes precedence.

## Verification

- `mvn -ntp validate -pl amoro-ams -am -DskipTests -Pskip-dashboard-build`
- Namespace REST, mapper, and Derby upgrade tests passed.
- `TestInternalIcebergCatalogService`: 11 tests passed on CentOS 7 with JDK 11.

MySQL and PostgreSQL upgrade scripts were reviewed but were not run against external database instances.

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/catalog/InternalCatalog.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/catalog/InternalCatalog.java
@@ -34,7 +34,10 @@ import org.apache.amoro.server.table.internal.InternalTableHandler;
 import org.apache.amoro.table.TableIdentifier;
 import org.apache.iceberg.rest.requests.CreateTableRequest;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
@@ -56,6 +59,12 @@ public abstract class InternalCatalog extends ServerCatalog {
   }
 
   public void createDatabase(String databaseName) {
+    createDatabase(databaseName, Collections.emptyMap());
+  }
+
+  public void createDatabase(String databaseName, Map<String, String> properties) {
+    Map<String, String> databaseProperties =
+        properties == null ? Collections.emptyMap() : new HashMap<>(properties);
     if (!databaseExists(databaseName)) {
       doAsTransaction(
           // make sure catalog existed in database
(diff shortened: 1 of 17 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/catalog/InternalCatalog.java
```diff
@@ -56,6 +59,12 @@ public List<String> listDatabases() {
...
 
   public void createDatabase(String databaseName) {
+    createDatabase(databaseName, Collections.emptyMap());
+  }
+
+  public void createDatabase(String databaseName, Map<String, String> properties) {
+    Map<String, String> databaseProperties =
+        properties == null ? Collections.emptyMap() : new HashMap<>(properties);
     if (!databaseExists(databaseName)) {
       doAsTransaction(
```
zhoujinsong, line 69: **Correction**: 
Sorry, my previous comment was misleading — there’s no need to manually call `beginSession` here. We recommend using the pre-wrapped `doAsTransaction` helper instead:

```java
Map<String, String> previousProperties = new HashMap<>();

doAsTransaction(
    () -> {
      DatabaseMetadata metadata =
          getAs(
              TableMetaMapper.class,
              mapper -> mapper.selectDatabaseMetadataForUpdate(name(), databaseName));
      if (metadata == null) {
        throw new ObjectNotExistsException(getDatabaseDesc(databaseName));
      }
      Map<String, String> properties =
          metadata.getProperties() == null
              ? new HashMap<>()
              : new HashMap<>(metadata.getProperties());
      previousProperties.putAll(properties);
      requestedRemovals.forEach(properties::remove);
      properties.putAll(requestedUpdates);
      doAsExisted(
          TableMetaMapper.class,
          mapper -> mapper.updateDatabaseProperties(name(), databaseName, properties),
          () -> new ObjectNotExistsException(getDatabaseDesc(databaseName)));
    });

return previousProperties;
```

zhoujinsong (approved):
LGTM.

PR #4284, opened 2026-07-21
[AMORO-4240][AMS] Support staged table creation in REST catalog

johntomcat7408-cmyk:
## Summary

This is the second PR for #4240.

- Honor `CreateTableRequest.stageCreate()` for internal Iceberg REST catalogs.
- Return staged metadata without writing files to the table location or registering the table in AMS.
- Complete table creation when the client sends the following create commit.
- Keep direct table creation and normal metadata commits unchanged.
- Add an integration test for the staged-create transaction.

## Why

Trino expects the target table location to remain empty between staged creation and the final commit. AMS previously wrote the initial metadata file during the staged request, so Trino rejected the location as non-empty.

## Verification

- `mvn -ntp validate -pl amoro-ams -am -DskipTests -Pskip-dashboard-build`
- `mvn -ntp test -pl amoro-ams -am


amoro-ams/src/main/java/org/apache/amoro/server/table/internal/InternalTableCreator.java
```diff
@@ -25,16 +25,19 @@
 /** Interface to create an internal table. */
 public interface InternalTableCreator extends Closeable {
 
-  /**
-   * Do all things about create an internal table, and prepare the {@link TableMetadata} for {@link
-   * InternalTableManager#createTable(java.lang.String, TableMetadata)}
-   */
+  /** Build Iceberg metadata without writing files or registering the table. */
+  org.apache.iceberg.TableMetadata stage();
```
zhoujinsong, line 29: It seems the FQCN here is not necessary.

amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
```diff
@@ -358,25 +366,101 @@ public void loadTable(Context ctx) {
...
+      InternalCatalog catalog, String database, String tableName, UpdateTableRequest request) {
+    request.requirements().forEach(requirement -> requirement.validate((TableMetadata) null));
+
+    Optional<Integer> formatVersion =
+        request.updates().stream()
+            .filter(MetadataUpdate.UpgradeFormatVersion.class::isInstance)
+            .map(MetadataUpdate.UpgradeFormatVersion.class::cast)
+            .map(MetadataUpdate.UpgradeFormatVersion::formatVersion)
+            .findFirst();
+    TableMetadata.Builder builder =
```
zhoujinsong, line 411: Format version from `UpgradeFormatVersion` is not applied to the builder，
You may change the format version like:

```java
TableMetadata.Builder builder = TableMetadata.buildFromEmpty();
formatVersion.ifPresent(builder::upgradeFormatVersion);
```

amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
```diff
@@ -358,25 +366,101 @@ public void loadTable(Context ctx) {
...
+        formatVersion.map(TableMetadata::buildFromEmpty).orElseGet(TableMetadata::buildFromEmpty);
+    request.updates().forEach(update -> update.applyTo(builder));
+    TableMetadata icebergMetadata = builder.build();
+
+    CreateTableRequest createRequest =
```
zhoujinsong, line 420: The `icebergMetadata` built from `UpdateTableRequest` updates already contains everything needed (schema, spec, sort order, location, properties, and appended data files). 
The reconstructed `CreateTableRequest` is only used to construct the `InternalTableCreator` object, but `creator.create(icebergMetadata)` writes the `icebergMetadata` directly — the `CreateTableRequest` fields are not used for the actual metadata file.

This works correctly today but is confusing to read. Could you add a comment explaining that `CreateTableRequest` here is only to satisfy the current `InternalTableCreator` constructor signature, and the actual metadata written is `icebergMetadata`?

Maybe we can refactor the `InternalTableCreator` later.

amoro-ams/src/test/java/org/apache/amoro/server/TestInternalIcebergCatalogService.java
```diff
@@ -284,6 +289,29 @@ public void testCreateTableInNamespaceLocation() {
       Assertions.assertEquals(namespaceLocation + "/" + table, created.location());
     }
 
+    @Test
```
zhoujinsong, line 294: Maybe we can add a `testStagedCreateWithAppendFiles ` to verify the CTAS case.

zhoujinsong (commented):
@johntomcat7408-cmyk 
Thanks a lot for the work, I left some comments, PTAL.

amoro-ams/src/main/java/org/apache/amoro/server/table/internal/InternalTableCreator.java
johntomcat7408-cmyk: Both `org.apache.amoro.server.table.TableMetadata` and `org.apache.iceberg.TableMetadata` are used in this interface. Java cannot import both classes with the same simple name, so one of them must remain fully qualified. I kept the existing Amoro import because both `create` methods return that type; reversing the import would only move the FQCN.

amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
johntomcat7408-cmyk: Thanks for pointing this out. In Iceberg 1.7.2, `TableMetadata.buildFromEmpty()` starts at format v2, so `builder.upgradeFormatVersion(1)` would be a downgrade and fail for an explicit v1 request. `buildFromEmpty(int)` initializes the requested version; I made that overload selection explicit and added coverage for both default v2 and explicit v1.

amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
johntomcat7408-cmyk: Added the comment. It now clarifies that the reconstructed `CreateTableRequest` only satisfies the current `InternalTableCreator` construction API, while `creator.create(icebergMetadata)` persists the fully reconstructed metadata.

amoro-ams/src/test/java/org/apache/amoro/server/TestInternalIcebergCatalogService.java
johntomcat7408-cmyk: Added `testStagedCreateWithAppendFiles`. It writes real data files, appends them inside the staged transaction, verifies the table is still not registered before the final commit, and then verifies the committed snapshot and data-file paths.

johntomcat7408-cmyk:
@zhoujinsong The latest Hadoop 3 / JDK 17 CI job failed in an unrelated optimizer test: `TestDefaultOptimizingService.testPollTaskThreeTimes` expected `null` but received a task after the test optimizer expired and its task was requeued.

I confirmed that the staged-create REST catalog tests changed by this PR passed in the same job, and this PR does not modify the optimizer service or that test. I filed #4303 to track the flaky heartbeat/timing condition separately.

Could you please rerun the failed CI workflow when convenient? I do not have permission to rerun it. The rerun result will help confirm whether #4303 is intermittent or consistently reproducible on CI.

Failed job: https://github.com/apache/amoro/actions/runs/30520803888/job/90800567957

johntomcat7408-cmyk:
@zhoujinsong Thanks for rerunning the workflow. Attempt 2 passed all four Hadoop 3 matrix jobs, including JDK 17 / Spark 3.5. I updated #4303 with the result; it will track the intermittent test-timing issue separately and should not block this PR. All required CI checks on the current head are now green.

zhoujinsong (approved):
LGTM.

Thanks for the work!

Issue #4290, opened 2026-07-24
[Bug]: Possible CDC snapshot loss when Flink and Amoro Optimizer commit concurrently


huangjc7:
### What happened?

Hi Amoro team,

We observed possible CDC snapshot loss when Flink CDC and Amoro Optimizer commit to the same Iceberg table at nearly the same time.

We have confirmed the data difference and missing snapshots, but we are not yet certain about the exact root cause. We would like to provide the evidence below and ask the Amoro team to help confirm the commit behavior.

### Environment

- Amoro: 0.8.1-incubating
- Flink: 1.20.3
- Flink CDC: 3.6.0
- Flink Iceberg runtime: 1.10.1
- Amoro Optimizer: Spark 3.3.2 / Iceberg 1.6.1
- Catalog: Amoro Iceberg REST Catalog
- Iceberg V2 table with upsert enabled
- Checkpoint interval: 60 seconds
- Table volume: about 4 million rows per day
- Self-optimizing: enabled

The issue is easier to observe on this high-throughput table because CDC and Optimizer both commit frequently.

### What we observed

For one day:

- MySQL rows: 3,771,955
- Iceberg rows: 3,757,759
- Difference: 14,196 rows

We found five CDC commits whose snapshot IDs are not present in the current Iceberg snapshot metadata:

| Time | CDC snapshot ID | Added records |
|---|---:|---:|
| 15:47 | 8495459209124852563 | 140 |
| 19:36 | 3149465639900291738 | 849 |
| 21:57 | 5182638233143834397 | 97 |
| 22:01 | 954105368916944994 | 773 |
| 22:10 | 5462846769141107542 | 12,337 |

The added record count of these five commits totals 14,196, which happens to match the MySQL/Iceberg row-count difference.

### Example

For checkpoint 4693, Flink logged two committables containing 12,337 records:

    2026-07-23 22:10:14 IcebergCommitter -
    CheckpointId: 4693, AddCount: 6060

    2026-07-23 22:10:14 IcebergCommitter -
    CheckpointId: 4693, AddCount: 6277

It then logged:

    2026-07-23 22:10:16 SnapshotProducer -
    Committed snapshot 5462846769141107542

    2026-07-23 22:10:16 MergingSnapshotProducer -
    Failed to load committed snapshot: omitting sequence number from notifications

The commit report contained:

    attempts=1
    addedRecords=12337

At approximately the same time, an Amoro OPTIMIZE snapshot was committed:

[...]

PR #4291, opened 2026-07-25
[AMORO-4290][AMS] Propagate internal Iceberg table commit failures

huangjc7:
From #4290 

image
Problem:
AMS swallowed Iceberg commit exceptions in IcebergInternalTableOperations. As a result, the REST Catalog could return success even when the metadata commit had failed, preventing clients such as Flink from retrying and potentially causing snapshot loss.

Fix:
Propagate the failure as a CommitFailedException after best-effort metadata cleanup. The REST Catalog now returns HTTP 409, allowing Flink to detect the conflict and retry the commit.
The image is from my online production Kubernetes cluster, 409 error.

image

Data validation is fine.


zhoujinsong (approved):
LGTM.

Thanks for the contribution!

PR #4293, opened 2026-07-27
[AMORO-4292][AMS] Guard IcebergTableMaintainer orphan cleanup against shared table locations

### LINKED ISSUE #4292: [Bug]: IcebergTableMaintainer orphan file cleanup can corrupt other tables sharing the same location
## Bug description

Amoro's `IcebergTableMaintainer.cleanOrphanFiles()` can delete files belonging to *other* Iceberg tables when two tables share the same `LOCATION`.

It computes reachable files from only the target table's metadata, then recursively lists the whole location and deletes everything else — including a second table's data, manifest, and metadata files under the same directory.

This can be triggered inadvertently in practice: for example, a table is created by copying `SHOW CREATE TABLE` of another table but forgetting to change `LOCATION`. We hit this in production (irreversible data loss) and found 100+ shared-location pairs in our environment.

## Proposed fix

Add a default-on guard in `IcebergTableMaintainer` before orphan cleanup:

1. List `metadata/*.metadata.json` (+ `.metadata.json.gz`) under `table.location()/metadata`.
2. For each file not in the table's own version set, read its `table-uuid`:
   - Different uuid → another table shares the location → **skip cleanup** (default).
   - Unreadable (legacy/compressed/corrupt) → treat as suspicious → skip (fail-safe).
   - Same uuid → older version of this table → ignore.
3. Only proceed when no foreign uuid is found.

Add table property `clean-orphan-file.ignore-location-conflict` (default `false`) to opt out when a shared location is intentional.

### Affects Versions

all

Iceberg

AMS

### How to reproduce

1. Create Iceberg table A at `hdfs://ns/tmp/shared`.
2. Create Iceberg table B at the **same** location `hdfs://ns/tmp/shared`.
3. Configure orphan file cleanup and trigger it on A:
   - Set catalog-level table property `clean-orphan-file.enabled=true` to enable periodic cleanup.
   - Set `clean-orphan-file.min-existing-time-minutes=2` so orphan files are eligible for deletion within minutes (instead of the default 2 days), for quick observation.
4. Restart AMS so the configuration takes effect, and let the scheduled orphan file cleanup task run.
5. B's files under the shared directory are deleted → B becomes corrupted/unloadable.

### Code of Conduct

terrytlu:
Fixes #4292

IcebergTableMaintainer.cleanOrphanFiles() lists and deletes files under a table's location. When two Iceberg tables share the same location (e.g. one table is created from another's metadata directory), the cleanup of table A can delete files that actually belong to table B, silently corrupting B.

This PR adds a pre-check, `IcebergTableUtil.hasOtherTableInLocation(Table)`:
- lists the table's metadata directory via its own (storage-agnostic) `FileIO` (requires `SupportsPrefixOperations`; otherwise a `ValidationException` is thrown so the caller can decide);
- compares the `table-uuid` of every other `metadata.json` with the current table's uuid;
- an unreadable / legacy / compressed metadata file (uuid cannot be determined) is treated as a conflict (fail-safe).

If another table is detected in the same location, `cleanOrphanFiles()` now **skips cleanup by default (fail-closed)** instead of risking another table's files. A new table property `clean-orphan-file.ignore-location-conflict` (default `false`) lets users opt out of the check when they are certain the location is not shared.

Without the guard, a single misconfigured / duplicated table location leads to silent cross-table data loss during scheduled orphan-file cleanup.

## Metrics

This PR also adds two metrics to make the orphan-file-cleaning result observable:

| Metric | Type | Meaning |
|---|---|---|
| `table_orphan_file_cleaning_last_status` | Gauge<Integer> | Status of the most recent cleaning attempt: `0=SUCCESS`, `1=LOCATION_CONFLICT`, `2=LOCATION_CONFLICT_CHECK_FAILED`, `3=EXECUTION_FAILED` |
| `table_orphan_file_cleaning_last_failure_timestamp_ms` | Gauge<Long> | Epoch millis of the last real cleaning failure |

Both metrics are tagged with `catalog`, `database`, `table`.

## How was this patch tested?

- Added `TestIcebergOrphanFileLocationConflict` (5 cases × 2 Iceberg parameterizations = 10 tests): no conflict, conflict via corrupt metadata, `ValidationException` when FileIO lacks prefix support, default skip-on-conflict, and ignore-conflict-when-property-enabled proceeds with cleanup.
- `mvn test -pl amoro-ams -Dtest=TestIcebergOrphanFileLocationConflict` → Tests run: 10, Failures: 0, Errors: 0.
- `spotless:check` passes.

## Documentation

This PR introduces a new table property `clean-orphan-file.ignore-location-conflict` (default `false`), which allows users to bypass the shared-location check when they are certain the table location is not shared. The property is documented in `docs/user-guides/configurations.md` (Self-optimizing / table maintenance configurations section).


czy006:
@terrytlu Thanks for your contributor. rebase banch and resolved files to master

terrytlu:
> @terrytlu Thanks for your contributor. rebase banch and resolved files to master

Thanks @czy006 for the reminder. I've rebased onto the latest master and resolved the conflicts. Please take another look when you get a chance.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/utils/IcebergTableUtil.java
```diff
@@ -193,4 +202,95 @@ public static Set<String> getAllManifestFiles(Table table) {
...
+    String myUuid = current.uuid();
+    Set<String> myMetadataFiles = Sets.newHashSet();
+    myMetadataFiles.add(new Path(current.metadataFileLocation()).getName());
+    for (TableMetadata.MetadataLogEntry entry : current.previousFiles()) {
+      myMetadataFiles.add(new Path(entry.file()).getName());
+    }
+
+    Path metadataDir = new Path(current.metadataFileLocation()).getParent();
+
+    AuthenticatedFileIO io = (AuthenticatedFileIO) table.io();
```
czy006, line 246: **Required:** This unconditional cast makes the new test fail with `ClassCastException` when `table.io()` is a regular Iceberg `FileIO`; the expected `ValidationException` is never reached. I reproduced this in `TestIcebergOrphanFileLocationConflict.testHasOtherTableInLocationThrowsWhenFileIoLacksPrefixSupport`. Please handle plain `FileIO` instances with `instanceof` and use `AuthenticatedFileIO.asPrefixFileIO()` for authenticated wrappers instead of casting blindly.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/utils/IcebergTableUtil.java
```diff
@@ -193,4 +202,95 @@ public static Set<String> getAllManifestFiles(Table table) {
...
+    TableMetadata current = ops.current();
+
+    String myUuid = current.uuid();
+    Set<String> myMetadataFiles = Sets.newHashSet();
+    myMetadataFiles.add(new Path(current.metadataFileLocation()).getName());
+    for (TableMetadata.MetadataLogEntry entry : current.previousFiles()) {
+      myMetadataFiles.add(new Path(entry.file()).getName());
+    }
+
+    Path metadataDir = new Path(current.metadataFileLocation()).getParent();
```
czy006, line 244: **Critical:** The guard derives the scan directory from `current.metadataFileLocation()`, while `IcebergTableMaintainer.clearInternalTableMetadata()` deletes from `table.location()/metadata`. With a custom metadata location, or with two tables sharing the data location but using different metadata directories, this check can miss the other table and the cleanup can still delete its files. Please make the guard use the same canonical metadata location as the cleanup path and add a regression test for `write.metadata.location`.

amoro-ams/src/test/java/org/apache/amoro/server/table/TestTableOrphanFilesCleaningMetrics.java
```diff
@@ -0,0 +1,110 @@
...
+
+import org.apache.amoro.ServerTableIdentifier;
+import org.apache.amoro.TableFormat;
+import org.apache.amoro.metrics.MetricRegistry;
+import org.apache.amoro.maintainer.MaintainerMetrics.CleanFailureReason;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
```
czy006, line 29: This import order currently fails the project's Spotless check. The normal Maven build stops before running tests with a format violation in this file. Please run Spotless and commit the formatted import order.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/maintainer/IcebergTableMaintainer.java
```diff
@@ -144,22 +158,84 @@ public Map<String, String> cleanOrphanFiles() {
...
+      table.refresh();
 
-    // refresh
-    table.refresh();
+      // clear metadata files
+      metadataDeleted = cleanMetadata(System.currentTimeMillis() - keepTime, metrics);
 
-    // clear metadata files
-    int metadataDeleted = cleanMetadata(System.currentTimeMillis() - keepTime, metrics);
+      // cleanup completed — reset status to SUCCESS
```
czy006, line 187: In a mixed table, both the change and base maintainers share the same metrics instance and execute this success update. If the change table records `LOCATION_CONFLICT` and the base table then succeeds, this line overwrites the final status to `SUCCESS`, hiding the skipped cleanup. Please aggregate child outcomes in `MixedTableMaintainer` and record one final status after both children complete.

amoro-common/src/main/java/org/apache/amoro/maintainer/MaintainerMetrics.java
```diff
@@ -40,6 +40,36 @@ public interface MaintainerMetrics {
    */
   void recordOrphanMetadataFilesCleaned(int expected, int cleaned);
 
+  /** Record a successful orphan-file-cleaning run. */
+  void recordSuccess();
```
czy006, line 44: Adding new abstract methods to this public interface can break external implementations at source or runtime (`AbstractMethodError`). Consider default no-op methods or a separate optional interface for cleanup-status metrics.

czy006 (changes requested):
Request changes. I reviewed the latest PR head fc22a661d395ef245da9f25759ec32c048025d01 against the current base. The overall direction is reasonable, but the current patch has reproducible test/build failures and several correctness gaps that should be addressed before merging.

Verification:
- The normal Maven build fails Spotless validation on TestTableOrphanFilesCleaningMetrics.java.
- With Spotless skipped, the metrics tests pass, but TestIcebergOrphanFileLocationConflict has 2 errors caused by ClassCastException.

Please address the inline comments below, add regression coverage for custom metadata locations and mixed-table status aggregation, then rerun the relevant tests and formatting checks.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/utils/IcebergTableUtil.java
```diff
@@ -193,4 +202,95 @@ public static Set<String> getAllManifestFiles(Table table) {
...
+
+    SupportsPrefixOperations prefixIo = (SupportsPrefixOperations) io;
+    for (FileInfo info : prefixIo.listPrefix(prefix)) {
+      String name = new Path(info.location()).getName();
+      if (!isMetadataJson(name) || myMetadataFiles.contains(name)) {
+        continue;
+      }
+
+      String otherUuid = readTableUuid(table, info.location());
+      if (otherUuid == null || !otherUuid.equals(myUuid)) {
```
j1wonpark, line 269: (blocking) A table created with `register_table` from another table's metadata keeps that table's uuid (`registerTable` commits the parsed metadata as-is, and later commits keep `base.uuid`). That is the "created from another's metadata directory" case in the PR description, and this check returns `false` for it, so the other table's newer files would still be deleted.

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/maintainer/IcebergTableMaintainer.java
```diff
@@ -144,22 +158,84 @@ public Map<String, String> cleanOrphanFiles() {
       return Maps.newHashMap();
     }
 
+    LocationConflictCheckResult checkResult = checkLocationConflict(tableConfiguration);
+
+    switch (checkResult) {
+      case CONFLICT_DETECTED:
```
j1wonpark, line 164: question: This also fires for a table that was dropped without purge and re-created at the same location — the old uuid's `metadata.json` files remain, so cleanup for such tables stops permanently with only a WARN log. Is that acceptable as a default, or should the release notes at least call it out with the opt-out property?

j1wonpark (changes requested):
Thanks for the contribution and the detailed write-up in #4292 — cross-table deletion is clearly worth preventing.

Before going further, I'd like to question the approach. Iceberg treats this as a location-ownership problem: tables that don't own their location set `gc.enabled=false`, and `ExpireSnapshots` / `DeleteOrphanFiles` refuse to run for them (apache/iceberg#4159). Amoro already honours that — `TableConfigurations` gates orphan cleanup, snapshot expiration and data expiration on `gc.enabled`. So the enforcement exists; what was missing in #4292 was *detecting* which tables share a location.

I think that detection belongs in AMS rather than in the maintainer. AMS loads every managed table anyway, so it can record `table.location()` (internal-catalog tables already persist it; external ones would need a column) and check uniqueness with one query — no file listing, no metadata parsing, and it also catches the `register_table` case where both tables share a uuid.

Two further concerns with the scan itself, on top of @czy006's comments: under Iceberg's default metadata retention (`delete-after-commit.enabled=false`, `previous-versions-max=100`) every metadata file beyond the last 100 versions is fully parsed on every cleanup cycle, and the fail-closed default silently disables cleanup for tables dropped and re-created at the same location.

Suggestion:
1. Split the two new gauges into their own small PR — they're useful regardless of how detection is done.
2. Move detection to an AMS-level location-uniqueness check and rely on `gc.enabled` / `clean-orphan-file.enabled` for enforcement.

@czy006 @zhoujinsong — would appreciate your view on the direction before the author invests more here.

PR #4298, opened 2026-07-28
[AMORO-4297][AMS] Persist failure reasons in task quota history

### LINKED ISSUE #4297: [Bug]: Optimizing task quota history does not persist failure reasons
### What happened?

`optimizing_task_quota.fail_reason` is always persisted as `NULL` for failed optimizing task attempts, even though `TaskRuntime` contains the failure reason.

This prevents users from identifying the cause of each retry attempt.

### Affects Versions

master (since v0.5.0)

### How to reproduce

1. Complete an optimizing task with a non-null error message.
2. Let AMS persist the task quota.
3. Check `optimizing_task_quota.fail_reason`.

The value is `NULL` instead of the task's failure reason.

### Anything else

`TaskRuntime.TaskQuota` has a `failReason` field, and the mapper persists it. However, its constructor does not copy `task.getFailReason()`.

https://github.com/apache/amoro/blob/ea4ea4cd31ecab35a3d8b7ebf39571d266cac472/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/TaskRuntime.java#L392-L435

### Code of Conduct

j1wonpark:

`TaskRuntime` records the error message when an optimizing task fails, but the failure reason was not copied to `TaskQuota`. As a result, `optimizing_task_quota.fail_reason` was persisted as `NULL`.

Fixes #4297.

## Brief change log

- Propagate the failure reason from `TaskRuntime` to `TaskQuota`.
- Add a regression test for the conversion.

## How was this patch tested?

- [x] Added a focused regression test.
- [x] Ran the related optimizing queue tests locally.

```shell
./mvnw test -pl amoro-ams -Dtest=TestTaskRuntime
./mvnw test -pl amoro-ams -Dtest=TestTaskRuntime,TestOptimizingQueue
```

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/TaskRuntime.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/TaskRuntime.java
@@ -408,6 +408,7 @@ public class TaskRuntime<T extends StagedTaskDescriptor<?, ?, ?>> extends Stated
       this.taskId = task.getTaskId().getTaskId();
       this.tableId = task.getTableId();
       this.retryNum = task.getRetry();
+      this.failReason = task.getFailReason();
     }
 
     public long getStartTime() {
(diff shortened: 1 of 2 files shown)


zhoujinsong (approved):
LGTM.

Thanks for the work!

PR #4302, opened 2026-07-30
[AMORO-4301][ams] Fix Kubernetes executor label prefix

### LINKED ISSUE #4301: [Bug]: Spark optimizer does not add Amoro labels to Kubernetes executor pods
### What happened?

When AMS builds the Spark submit command for a Kubernetes-native optimizer, `KUBERNETES_EXECUTOR_LABEL_PREFIX` is set to `spark.kubernetes.driver.label.`, the same value as the driver prefix.

`addKubernetesProperties()` writes three labels for the driver and then three intended labels for executors. Because Spark options are stored in a map, the second set overwrites the same driver keys. The generated command contains only `spark.kubernetes.driver.label.*` options and no `spark.kubernetes.executor.label.*` options.

As a result, executor pods do not receive Amoro's structured optimizer labels: `optimizer-group`, `optimizer-implementation`, and `optimizer-id`. This does not prevent optimizer startup, execution, or release, but it affects observability and label-based operations.

Expected behavior: emit both driver and executor label options using Spark's separate documented prefixes.

### Affects Versions

master (ec28554a2)

_No response_

Optimizer, Spark

### How to reproduce

1. Create a `SparkOptimizerContainer` whose `optimizingContainerProperties.master` uses a `k8s://` URL.
2. Call `generateStartupArgs(...)` for an optimizer resource.
3. Inspect the generated `--conf` arguments.
4. Observe that the three `spark.kubernetes.driver.label.*` options are present, while the corresponding `spark.kubernetes.executor.label.*` options are absent.

### Relevant log output

```shell

```

### Anything else

Spark documents separate driver and executor label prefixes: https://spark.apache.org/docs/3.5.8/running-on-kubernetes.html#spark-properties

The incorrect prefix appears to have been introduced with the Kubernetes-native Spark optimizer support. A one-line constant correction plus regression coverage for both driver and executor startup arguments should address the issue.

### Code of Conduct

csurong:

`KUBERNETES_EXECUTOR_LABEL_PREFIX` currently uses the driver label prefix. Because Spark options are stored by key in a map, the intended executor entries overwrite the same driver keys and no `spark.kubernetes.executor.label.*` options are emitted. Executor pods therefore miss Amoro's optimizer labels for observability and label-based operations.

Closes #4301.

## Brief change log

- Use `spark.kubernetes.executor.label.` for executor pod labels.
- Add regression coverage for all optimizer labels on both driver and executor startup arguments.

## How was this patch tested?

- [x] Added a regression test that fails before the prefix correction and passes after it.
- [x] `mvn test -pl amoro-ams -am -Dtest=TestSparkOptimizerContainer -Dsurefire.failIfNoSpecifiedTests=false -Pskip-dashboard-build` (JDK 11; 5 tests passed).
- [x] `mvn validate -pl amoro-ams` (JDK 11; Spotless and Checkstyle passed).
- [ ] Screenshots for manual tests (not applicable; no UI change).

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/manager/SparkOptimizerContainer.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/manager/SparkOptimizerContainer.java
@@ -340,7 +340,8 @@ public class SparkOptimizerContainer extends AbstractOptimizerContainer {
     public static final String KUBERNETES_NAMESPACE = "spark.kubernetes.namespace";
     public static final String KUBERNETES_SUBMISSION_WAIT_APP_COMPLETION =
         "spark.kubernetes.submission.waitAppCompletion";
-    public static final String KUBERNETES_EXECUTOR_LABEL_PREFIX = "spark.kubernetes.driver.label.";
+    public static final String KUBERNETES_EXECUTOR_LABEL_PREFIX =
+        "spark.kubernetes.executor.label.";
     public static final String KUBERNETES_DRIVER_LABEL_PREFIX = "spark.kubernetes.driver.label.";
     public static final String KUBERNETES_DRA_ENABLED = "spark.dynamicAllocation.enabled";
     public static final String KUBERNETES_DRA_MAX_EXECUTORS =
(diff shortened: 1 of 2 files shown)


zhoujinsong (approved):
LGTM.

Thanks for the fix.

xxubai (approved):

PR #4306, opened 2026-07-31
[Hotfix][tools] Include Helm chart in change-version script

xxubai:

The `tools/change-version.sh` script currently updates only Maven POM files. As a result, the Helm chart `appVersion` can remain stale after a project version change, requiring a follow-up hotfix to align the chart with the released application version.

The root cause is that the script searches only for `pom.xml` files and does not include the Helm chart as a project-version carrier.

## Brief change log

- Update the Helm chart `appVersion` together with Maven POM versions.
- Keep the release guide's embedded script example in sync with the implementation.

## How was this patch tested?

  - `bash -n tools/change-version.sh`
  - Isolated fixture verification for POM and Helm chart version replacement, including idempotency
  - `./mvnw validate`
  - `helm lint charts/amoro`

## Documentation


zhoujinsong (approved):
LGTM.

Thanks for the work!

PR #4307, opened 2026-07-31
[AMORO-4303][AMS] Stabilize optimizer heartbeat tests

### LINKED ISSUE #4303: [Bug]: TestDefaultOptimizingService.testPollTaskThreeTimes can fail after optimizer heartbeat expiration
## What happened?

`TestDefaultOptimizingService.testPollTaskThreeTimes` failed intermittently in the Hadoop 3 / JDK 17 / Spark 3.5 CI job while validating PR #4284.

The test expects the final `pollTask` call to return `null`, but it returned the same `OptimizingTask` after the optimizer was considered expired and the task was put back into the retry queue.

This failure is unrelated to the staged table creation changes in PR #4284. The REST catalog tests changed by that PR passed in the same job, and the parent `master` commit passed the equivalent JDK 17 job.

Expected behavior: tests that do not exercise optimizer expiration should keep the test optimizer alive deterministically and should not fail because of CI scheduling delays.

## Affects Versions

`master` at/after `b7bf0d39d`; observed while testing PR #4284 at `3844f9ed73f2c3a33802fefe5e6559e806bd6efd`.

Iceberg test fixture; the failure is in AMS optimizer test infrastructure rather than format-specific behavior.

AMS / Optimizer

## How to reproduce

The failure is intermittent and currently observed once in the full Hadoop 3 CI matrix:

1. Use JDK 17 and Spark 3.5.
2. Run the Hadoop 3 build/test workflow, or repeatedly run:

   `./mvnw -pl amoro-ams -am -Dtest=TestDefaultOptimizingService#testPollTaskThreeTimes test`

3. Under sufficient CI scheduling or lock contention, the background `Toucher` may not complete a heartbeat before the 800 ms test timeout.
4. The optimizer keeper expires the optimizer and requeues its task; the assertion at line 236 then receives a task instead of `null`.

Failed job: https://github.com/apache/amoro/actions/runs/30520803888/job/90800567957

Successful equivalent job on the parent `master` commit: https://github.com/apache/amoro/actions/runs/30519308937

## Relevant log output

```text
[ERROR] TestDefaultOptimizingService.testPollTaskThreeTimes
expected: <null> but was: <OptimizingTask(...)>
at org.apache.amoro.server.TestDefaultOptimizingService.testPollTaskThreeTimes(TestDefaultOptimizingService.java:236)

INFO DefaultOptimizingService: Optimizer ... has been expired, unregister it
INFO DefaultOptimizingService: Task ... is suspending, since its optimizer is expired, put it to retry queue
```

## Analysis

The test setup currently uses:

- `OPTIMIZER_HB_TIMEOUT = 800 ms` in `AMSServiceTestBase`.
- A background `Toucher` that sleeps 300 ms between heartbeat attempts.

That margin is vulnerable to CI thread scheduling and database/lock delays. Once the heartbeat misses the 800 ms window, the production retry-on-expiration behavior is valid, but it invalidates this test's assumption that the optimizer remains alive throughout the final blocking poll.

This issue tracks stabilization of the test fixture or timing control; it does not assert a production regression yet. A rerun of the failed CI job is requested to determine whether the failure reproduces.

Yes, after confirming the failure or agreeing on the preferred deterministic test approach.

## Code of Conduct

I agree to follow this project's Code of Conduct.

johntomcat7408-cmyk:

Close #4303.

`AMSServiceTestBase` applied an 800 ms optimizer heartbeat timeout to every optimizer test, while its background toucher only attempted a heartbeat every 300 ms. Under CI scheduling or database delays, tests that were not exercising expiration could lose the optimizer, requeue its task, and fail nondeterministically.

## Brief change log

- Use a 10-second heartbeat timeout for ordinary optimizer tests while retaining an explicit 800 ms timeout in expiration-specific tests.
- Touch every 100 ms and replace fixed sleeps with condition-based waits for task and optimizer state transitions.
- Fully dispose and restore test services around custom-timeout scenarios so background keepers cannot leak between tests.
- Ignore keeper failures only after shutdown has begun, while preserving runtime error logging.

## How was this patch tested?

- [x] Update the existing optimizer service tests to cover normal and expiration-specific timing independently.
- [ ] Add screenshots for manual tests if appropriate (not applicable).
- [x] Run tests locally before making a pull request.

Validation performed against commit `7efadacadc3561b60ec7c10cc28e91964c29e5a8`:

- JDK 11: full `TestDefaultOptimizingService` suite (22 tests) passed, followed by 20/20 successful repetitions of `testPollTaskThreeTimes`.
- JDK 17: full `TestDefaultOptimizingService` suite (22 tests) passed, followed by 20/20 successful repetitions of `testPollTaskThreeTimes`.
- Spotless, Checkstyle, production compilation, and test compilation passed.

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
@@ -608,7 +608,9 @@ public class DefaultOptimizingService extends StatedPersistentBase
           }
         } catch (InterruptedException ignored) {
         } catch (Throwable t) {
-          LOG.error("{} has encountered a problem.", this.getClass().getSimpleName(), t);
+          if (!stopped) {
+            LOG.error("{} has encountered a problem.", this.getClass().getSimpleName(), t);
+          }
         }
       }
     }
--- a/amoro-ams/src/test/java/org/apache/amoro/server/AMSServiceTestBase.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/AMSServiceTestBase.java
@@ -33,16 +33,24 @@ import org.junit.BeforeClass;
 import java.time.Duration;
 
 public abstract class AMSServiceTestBase extends AMSManagerTestBase {
+  // Normal optimizer tests may block in pollTask for 3 seconds. Keep the heartbeat timeout well
+  // outside that window so CI scheduling delays do not turn unrelated tests into expiry tests.
+  private static final Duration DEFAULT_OPTIMIZER_HEARTBEAT_TIMEOUT = Duration.ofSeconds(10);
+
   private static DefaultTableService TABLE_SERVICE = null;
   private static DefaultOptimizingService OPTIMIZING_SERVICE = null;
   private static ProcessService PROCESS_SERVICE = null;
 
   @BeforeClass
   public static void initTableService() {
+    initTableService(DEFAULT_OPTIMIZER_HEARTBEAT_TIMEOUT);
+  }
+
+  protected static void initTableService(Duration optimizerHeartbeatTimeout) {
     DefaultTableRuntimeFactory runtimeFactory = new DefaultTableRuntimeFactory();
     try {
       Configurations configurations = new Configurations();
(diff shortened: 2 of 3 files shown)


zhoujinsong (approved):
LGTM.

Thanks for the fix.

PR #4310, opened 2026-08-04
[AMORO-4139] Delete resource and optimizer atomically when releasing optimizer

### LINKED ISSUE #4139: [Improvement]: NPE when releasing optimizer due to non-atomic deleteResource + deleteOptimizer
### Search before asking

When releasing an optimizer via `DELETE /api/ams/v1/optimize/optimizers/{jobId}`, a `NullPointerException` is thrown if the `resource` table entry is missing.

### Root cause
 `deleteResource` and `deleteOptimizer` are executed as two separate DB operations without a transaction:

  ```java
  optimizerManager.deleteResource(resourceId);                           // step 1
  optimizerManager.deleteOptimizer(resource.getGroupName(), resourceId); // step 2
```
If step 1 succeeds but step 2 fails, the resource row is deleted while the optimizer row remains. On the next release attempt, getResource() returns null and the following line throws NPE:

```
Resource resource = optimizerManager.getResource(resourceId); // null
resource.getProperties().putAll(...);                         // NPE
```

### How should we improve?

 Both operations should be executed atomically in a single transaction to prevent inconsistent state. If resource is already missing due to a previous partial failure, the release should handle it gracefully using OptimizerInstance fields instead of failing with NPE.

### Subtasks

_No response_

### Code of Conduct

### PR DESCRIPTION
**Pull request by seoyoniee**
<!--
Please clarify why the changes are needed. For instance,
  1. If you add a feature, you can talk about its use case.
  2. If you fix a bug, you can clarify why it is a bug.
  3. Use Fix/Resolve/Close #{ISSUE_NUMBER} to link this PR to its related issue
-->

Close #4139 

`OptimizerController#releaseOptimizer` deleted the resource record and the optimizer records through two independent transactions:                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                         
```java                                                                                                                                                                                                                                
  optimizerManager.deleteResource(resourceId);                                                                                                                                                                                           
  optimizerManager.deleteOptimizer(resource.getGroupName(), resourceId);                              
```                                                                   

Both records describe the same optimizer, so deleting them in separate transaction can leave the resource gone while its optimizer rows survive - a state the release API can no longer recover from.                                                                
                                                                                                                                                                                                                                         
If the first call committed and the second one failed, the resource row was deleted while the matching optimizer rows survived. That state cannot be recovered through the API:                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                         
- getResource(resourceId) returns null on the next release attempt, so resource.getProperties().putAll(...) throws a NullPointerException.                                                                                                                                                                    
- The container startup stats (yarn-application-id, the Kubernetes namespace and name, ...) are persisted only in the resource properties, so once that row is gone the underlying optimizer process cannot be released either.                                                                              

## Brief change log
<!--
Clearly describe the changes made in modules, classes, methods, etc.
-->

- `OptimizerManager`: rename deleteOptimizer(String, String) to                                                                                                                                                                          
  deleteOptimizerAndResource(String, String), since the operation now owns both                                                                                                                                                          
  records. OptimizerController#releaseOptimizer was its only caller;                                                                                                                                                                     
  DefaultOptimizingService#deleteOptimizer is a separate method and is left untouched.                                                                                                                                                   
 - `DefaultOptimizerManager#deleteOptimizerAndResource`: run the resource deletion and                                                                                                                                                    
  the optimizer deletion inside a single doAsTransaction(...). The existing deletion                                                                                                                                                     
  logic is unchanged, only wrapped. Because NestedSqlSession keeps one SQL session per                                                                                                                                                   
  thread, the nested doAs commits become no-ops and the outermost transaction commits                                                                                                                                                    
  or rolls back both deletions together.                                                                                                                                                                                                 
- `OptimizerController#releaseOptimizer`: replace the two calls with the single                                                                                                                                                          
  transactional one.

## How was this patch tested?

Should I add a `TestDefaultOptimizerManager` to this PR?

## Documentation


johntomcat7408-cmyk (commented):
Thanks for working on this. I traced the transaction flow through PersistentBase and NestedSqlSession: the outer doAsTransaction keeps one thread-local SQL session, so the nested deleteResource/getAs/unregisterOptimizer calls do not commit independently; this is a sound way to prevent future partial deletion. I do think this PR should add regression coverage before merge: (1) the success path deletes the resource and every optimizer row for the resource ID, and (2) a failure during optimizer deletion rolls back the preceding resource deletion, proving the intended atomicity rather than only the final happy-path state. There is also one scope point to clarify. The existing inconsistent state described in #4139 is not recovered by this patch: OptimizerController still dereferences resource.getProperties() immediately after getResource(resourceId), so a missing resource row fails before deleteOptimizerAndResource is reached. If container startup-only properties make terminating the underlying process impossible, could the PR define and test a graceful non-NPE behavior for that state (and document any cleanup limitation), or explicitly agree with maintainers that recovery will be a separate follow-up? Finally, the group argument is still unused; since the API is already being renamed, a resourceId-only signature would be clearer unless group is intentionally retained for future validation.

zhoujinsong (approved):
This PR look good to me.

And thanks for the review! @johntomcat7408-cmyk 

On the scope point regarding the pre-existing inconsistent state, I agree this PR doesn't recover from it. 

I'd prefer to keep historical recovery as a separate follow-up rather than expanding this PR's scope, but I'll make the limitation explicit in the PR description and link the follow-up issue. Let me know if you'd prefer otherwise.

On the unused `group` argument: agreed. We can polish the code in another PR.

PR #4312, opened 2026-08-12
[Hotfix][AMS] Unify authentication handling for dashboard requests

xxubai:

Dashboard requests containing the legacy single-page token parameter follow a separate authentication path. This causes authentication behavior to differ between dashboard requests.

This change removes the legacy single-page token handling and ensures that all dashboard requests use the standard authentication flow.

## Brief change log

- Route all dashboard requests through the standard request authentication handler.
- Remove the legacy single-page token generation and validation methods.
- Remove the unused table-page token controller method.
- Add tests covering catalog, terminal, and web-session authentication behavior.

## How was this patch tested?


zhoujinsong (approved):
LGTM.

Thanks for the fix.

PR #4296, opened 2026-07-28
[AMORO-4295] AIP-5 Phase 3+4: idle-driven scale-down with graceful drain for dynamic allocation groups

j1wonpark:
## Why are the changes needed?

Close #4295.

Third and fourth implementation phase of **AIP-5: Dynamic Resource Allocation for Optimizer** (#4191), building on the scale-up merged in #4272: per-optimizer idle tracking and idle-driven **scale-down with graceful drain** for dynamic-allocation-enabled groups. One PR because idle tracking's only consumer is the scale-down decision. Opt-in as before: groups without `dynamic-allocation.enabled = true` are unaffected.

One deliberate deviation from the AIP page: idle tracking is derived from the per-round load snapshot (per-token in-flight counts) instead of the sketched event-based `OptimizerCounters`. Several task-reclamation paths offer no service-level release hook (`resetStaleTasksForThread`, the suspending predication reclaiming timed-out tasks of live optimizers, process close), so an event counter would silently drift — and one leaked increment makes an optimizer permanently non-idle, i.e. a permanent pod leak. Snapshot-derived state is re-derived every round and immune to all of them, consistent with the state-derived position taken in #4272. The AIP page will be revised alongside this PR.

Drain is an efficiency layer only — correctness stays with the existing retry-on-expiry machinery. A draining optimizer stops receiving tasks (`pollTask` returns `null`), finishes in-flight work, and is removed; a `drain-timeout` force removal falls back to the existing orphan recovery.

## Brief change log

- `OptimizingQueue.collectDynamicAllocationLoad()` additionally aggregates per-token in-flight counts; recovered tasks keep their tokens, so the first post-restart snapshot is accurate.
- `DynamicAllocationState`: `observe()` tracks last-busy time per token (new tokens are seeded idle, so an instance that never gets a task is still reclaimed); `computeScaleDown()` applies cooldown → idle filter → floor (`registered − draining − candidate ≥ min-parallelism`) → longest-idle selection, one per round, and runs only in rounds with no scale-up demand signal (`wasDemandActive()`). New validation: `sustained-backlog-timeout ≤ executor-idle-timeout / 2` (observation resolution equals the keeper cadence).
[...]

PR #4314, opened 2026-08-13
[AMORO-4313] AIP-5 Phase 5: observability metrics for dynamic allocation

### LINKED ISSUE #4313: [Subtask]: AIP-5 Phase 5 — Observability metrics for dynamic allocation
### Parent issue

#4191 (AIP-5: Dynamic Resource Allocation for Optimizer)

### Description

Final phase of [AIP-5](https://cwiki.apache.org/confluence/display/AMORO/AIP-5%3A+Dynamic+Resource+Allocation+for+Optimizer): expose the scaling behavior of dynamic allocation groups as per-group metrics (all tagged `group`).

**Gauges**
- `optimizer_group_idle_optimizers` — instances with zero in-flight tasks
- `optimizer_group_config_invalid` — `1` while an opted-in group's dynamic allocation config is invalid and the startup fail-safe keeps it disabled
- `optimizer_group_pending_removal_optimizers` — instances in graceful drain
- `optimizer_group_effective_threads` — registered threads plus threads of optimizers pending registration
- `optimizer_group_backlog_duration_ms` — duration since demand first exceeded capacity

**Counters**
- `optimizer_group_scale_up_total` / `optimizer_group_scale_down_total` — cumulative scaling actions, one per scale-out round / drain start, so thrashing-style alerts become expressible with `increase(...[1h])`

The first two gauges register with the existing `OptimizerGroupMetrics` for every group; the other five live with the scale keeper's watch on the group, since their sources (drain set, boot-window accounting, backlog timer) exist only while dynamic allocation is enabled.

Also documents the new metrics in `docs/user-guides/metrics.md`.

Depends on #4295.

j1wonpark:

Close #4313.

Final phase of [AIP-5](https://cwiki.apache.org/confluence/display/AMORO/AIP-5%3A+Dynamic+Resource+Allocation+for+Optimizer). The scaling behavior from #4272 and #4296 is currently observable only through INFO logs; this exposes it as metrics so operators can alert on it — in particular resource thrashing ("N scale-downs in the last hour", `increase(optimizer_group_scale_down_total[1h])`) and the config fail-safe fallback that would otherwise be silent.

## Brief change log

New per-group metrics (all tagged `group`), registered in two places by ownership of their source state:

- **`OptimizerGroupMetrics` (every group, queue lifecycle):**
  - `optimizer_group_idle_optimizers` (gauge) — instances with zero in-flight (SCHEDULED/ACKED) tasks
  - `optimizer_group_config_invalid` (gauge) — `1` while an opted-in group's DRA config is invalid. Registered for every group because an invalid config makes the group *not* effectively enabled — tying this gauge to the DRA watch would remove it exactly when it should alarm.
- **New `DynamicAllocationMetrics` (DRA groups only, scale-keeper watch lifecycle):**
  - `optimizer_group_pending_removal_optimizers` (gauge) — instances in graceful drain
  - `optimizer_group_effective_threads` (gauge) — registered threads + threads of optimizers pending registration, the raw observable as defined in the AIP (the keeper's round arithmetic additionally subtracts draining instances internally)
  - `optimizer_group_backlog_duration_ms` (gauge) — duration since demand first exceeded capacity, derived at scrape time (keeper rounds are seconds apart, pushing would go stale between scrapes)
  - `optimizer_group_scale_up_total` / `optimizer_group_scale_down_total` (counters) — one *attempted* scale-out round (regardless of instance count or request outcome) and one drain start each count 1; instance-count trends are already covered by `optimizer_group_optimizer_instances`, and a rising counter without rising instances sustained beyond the pod boot window signals failing resource requests

Naming follows the Prometheus/OpenMetrics conventions (`_total` counter suffix, `_ms` unit suffix) as published in the AIP, rather than the pre-reporter `_count`/`_mills` suffixes of the existing table metrics.

Metric lifecycle is hardened against the failure paths it introduces: registration failures roll back partially registered metrics so the group stays rewatchable, the keeper's dispose unregisters its metrics (the global registry outlives the service across HA leader hand-offs), and a watch arriving after dispose is a no-op (`stopped` guard, the same pattern as `OptimizerGroupKeeper.keepInTouch`). Watch/unwatch are serialized, and a disable unwatches on the config-entry path itself — the round-driven unwatch runs on the leader only, so followers would otherwise keep exporting the group's DRA metrics until failover.

Supporting changes: `DynamicAllocationState.backlogDurationMs()` read accessor, `DynamicAllocationConfig.isConfigInvalid()` (a group that never opted in is never invalid, whatever its leftover properties parse to), and the metrics table in `docs/user-guides/metrics.md`.

## Known trade-offs

- Disabling and re-enabling dynamic allocation re-registers the keeper-scoped metrics, resetting the counters to 0. Prometheus `increase()`/`rate()` are counter-reset tolerant, so alerts survive this.
- On HA standby nodes the gauges are registered (as the existing group metrics are) but the counters stay 0, since only the leader evaluates scaling.
- The `backlog_duration_ms` gauge wiring is covered by unit tests on the accessor plus the keeper lambda; an end-to-end assertion under real demand would need the heavy real-table demand harness already declared out of scope in the previous phases.

## Documents & Website changes

`docs/user-guides/metrics.md`: 2 rows added to the optimizer group table, plus a dynamic-allocation metrics table with the counter semantics.

## How was this patch tested?

- `TestDynamicAllocationMetrics` (new): gauge/counter registration, source-backed values, unregistration, rollback of partial registration on failure.
- `TestOptimizerScaleKeeper`: metrics registered on watch and removed on disable; a floor round counts one scale-up action; a drain start counts one scale-down action; a failed watch leaves the group rewatchable; dispose unregisters the metrics and a watch arriving after dispose registers nothing; an enabled-to-enabled config update re-enters watch idempotently.
- `TestOptimizingQueue`: idle-optimizer gauge flips with an in-flight task; config-invalid gauge flips on an invalid runtime update.
- `TestDynamicAllocationConfig` / `TestComputeScaleUp`: `isConfigInvalid` and `backlogDurationMs` unit coverage.

All new code was written test-first; full DRA suite green (184 tests, Phase 1–4 tests unmodified).

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
@@ -41,11 +41,13 @@ import org.apache.amoro.server.catalog.CatalogManager;
 import org.apache.amoro.server.dashboard.model.OptimizerResourceInfo;
 import org.apache.amoro.server.ha.HighAvailabilityContainer;
 import org.apache.amoro.server.manager.AbstractOptimizerContainer;
+import org.apache.amoro.server.manager.MetricManager;
 import org.apache.amoro.server.optimizing.OptimizingProcess;
 import org.apache.amoro.server.optimizing.OptimizingQueue;
 import org.apache.amoro.server.optimizing.OptimizingStatus;
 import org.apache.amoro.server.optimizing.TaskRuntime;
 import org.apache.amoro.server.optimizing.dra.DynamicAllocationConfig;
+import org.apache.amoro.server.optimizing.dra.DynamicAllocationMetrics;
 import org.apache.amoro.server.optimizing.dra.DynamicAllocationState;
 import org.apache.amoro.server.optimizing.dra.PendingRegistrations;
 import org.apache.amoro.server.persistence.StatedPersistentBase;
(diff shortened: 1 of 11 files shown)


j1wonpark:
Hi @zhoujinsong @xxubai @czy006, AIP-5 Phase 5 (the final phase, observability metrics) is ready for review. It includes the config-invalid gauge and alertable scale events suggested in the design review. CI is green. Thanks!

zhoujinsong (approved):
Looks good — the Source-backed gauge pattern, lifecycle management, and defensive error handling in `config_invalid` are all well done. No issues with the 7 metrics here; they cover the core "is DRA working?" question well.

A few additional metrics worth considering for follow-up — they'd help operators diagnose failures and spot config limits. None of these block merging.

### 1. `scale_up_failed_total` (Counter) — high priority

`scale_up_total` increments unconditionally before the `requestResource`/`createResource` loop. The code already distinguishes success vs failure in the catch block, but only logs it — a `scale_up_failed_total` incremented there would let operators directly see failing scale-ups without correlating against `optimizer_instances` (which is ambiguous during the pending-registration window).

### 2. `scale_down_failed_total` (Counter)

`executeRemoval` can fail on container release and retry silently. A failure counter would distinguish "drain stuck because tasks won't finish" from "drain stuck because releases keep failing."

### 3. `at_max_capacity` / `at_min_capacity` (Gauge 0/1)

`1` when `effectiveThreads >= maxParallelism` / `<= minParallelism`, else `0`. Directly answers "is this group maxed/floor-bound?" without operators needing to know the configured limits. Combined with `backlog_duration_ms > 0`, cleanly identifies "needs manual `max-parallelism` increase" — currently requires inference from "backlog growing but `scale_up_total` stopped."

PR #4315, opened 2026-08-13
[Docs] Fix the broken Star History chart in the README

OctoBored:
The Star History chart in the README is currently broken and no longer renders correctly. This change points the chart to a working mirror so the project's star history is displayed again. This is a small documentation-only fix.


xxubai (approved):
Thanks for the contribution

PR #4317, opened 2026-08-15
[AMORO-4316][AMS] Support monthly period for auto-creating Iceberg tags

### LINKED ISSUE #4316: [Feature]: Support monthly period for auto-creating Iceberg tags
### Description

`tag.auto-create.trigger.period` currently supports only `daily` and `hourly` (`TagConfiguration.Period`). This proposes adding `monthly` as a third period.

### Use case/motivation

Auto-created tags are useful as long-lived, point-in-time checkpoints — for archival, audit, and recovering data that no longer exists in the current snapshot (for example, tables whose upstream source periodically purges old rows, so the live table only ever holds a recent window).

For that class of retention requirement, a calendar month is the natural granularity. But `tag.auto-create.trigger.period` accepts only `hourly` and `daily`, so there is no way to express this.

### Describe the solution

Add `MONTHLY("monthly")` to `TagConfiguration.Period`.

One part is not a plain enum-constant addition: `periodDuration()` must widen from `Duration` to `TemporalAmount`, because `generateTagName` derives the tag name as `tagTime.minus(periodDuration())` and a calendar month is not a fixed-length duration.

### Subtasks

_No response_

### Related issues

_No response_

### Code of Conduct

seoyoniee:

Close #4316 

Auto-created tags are useful as long-lived checkpoints for archival and audit. For that, a calendar month is often the right granularity — `daily` creates ~365 refs per table per year, and each one pins a snapshot from expiration.

`tag.auto-create.trigger.period` currently accepts only `hourly` and `daily`, so there is no way to express this without giving up auto-create and driving `createTag()` from an external scheduler.

## Brief change log
- `TagConfiguration.Period`: add `MONTHLY("monthly")`.
- `TagConfiguration.Period#periodDuration()`: widen `Duration` → `TemporalAmount`.
  A calendar month is not a fixed-length duration, so `java.time.Period` is needed here. With `Duration.ofDays(30)` the February tag would be named `tag-202601`, collide with January's, and never be created.
  `DAILY`/`HOURLY` keep returning `Duration` (covariant override).
- `TableProperties`: add `AUTO_CREATE_TAG_FORMAT_MONTHLY_DEFAULT = "'tag-'yyyyMM"`.
- `TableConfigurations#parseTagConfiguration`: add the `MONTHLY` case to the default-format switch (it throws on unknown periods). The same switch is duplicated in `TestAutoCreateIcebergTagAction`'s helper and updated too.
- Docs updated.

## How was this patch tested?

- [x]  Add some test cases that check the changes thoroughly including negative and positive cases if possible
  `TestAutoCreateIcebergTagAction`: `testCreateMonthlyTag`,
  `testMonthlyTagNameAcrossMonthBoundary` (Feb/Mar, leap year, year boundary), and `monthly` cases in `testTriggerTimePeriod`.

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurations.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/TableConfigurations.java
@@ -389,6 +389,9 @@ public class TableConfigurations {
       case HOURLY:
         defaultFormat = TableProperties.AUTO_CREATE_TAG_FORMAT_HOURLY_DEFAULT;
         break;
+      case MONTHLY:
+        defaultFormat = TableProperties.AUTO_CREATE_TAG_FORMAT_MONTHLY_DEFAULT;
+        break;
       default:
         throw new IllegalArgumentException(
             "Unsupported trigger period: " + tagConfig.getTriggerPeriod());
--- a/amoro-common/src/main/java/org/apache/amoro/config/TagConfiguration.java
+++ b/amoro-common/src/main/java/org/apache/amoro/config/TagConfiguration.java
@@ -26,6 +26,7 @@ import java.time.Duration;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
+import java.time.temporal.TemporalAmount;
 
 /** Configuration for auto creating tags. */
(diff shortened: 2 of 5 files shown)


zhoujinsong (approved):
LGTM.

czy006 (approved):

PR #4320, opened 2026-08-17
[hotfix][AMS] Clean up orphan optimizer records at startup

czy006:
## Brief change log

Clean up optimizer records that no longer have a corresponding runtime when AMS starts. This prevents a null dereference during recovery and allows startup to complete normally.

## How was this patch tested?

- [x] Add test coverage for orphan optimizer cleanup during optimizer-service startup.
- [x] Run `TestDefaultOptimizingService` locally with JDK 11 before creating this pull request.

## Documentation

- no
- not applicable

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
@@ -239,7 +239,17 @@ public class DefaultOptimizingService extends StatedPersistentBase
 
   private void registerOptimizer(OptimizerInstance optimizer, boolean needPersistent) {
     if (needPersistent) {
-      doAs(OptimizerMapper.class, mapper -> mapper.insertOptimizer(optimizer));
+      doAsTransaction(
+          () -> {
+            String groupName =
+                getAs(
+                    ResourceMapper.class,
+                    mapper -> mapper.selectResourceGroupNameForUpdate(optimizer.getGroupName()));
+            if (groupName == null) {
+              throw new ObjectNotExistsException("Optimizer group " + optimizer.getGroupName());
+            }
+            doAs(OptimizerMapper.class, mapper -> mapper.insertOptimizer(optimizer));
+          });
     }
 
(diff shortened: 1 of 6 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
```diff
@@ -250,6 +250,26 @@ private void registerOptimizer(OptimizerInstance optimizer, boolean needPersiste
...
 
+  /**
+   * Registers optimizers recovered from persistence at startup. An optimizer record referencing a
+   * resource group that no longer exists (e.g. the group was dropped while AMS was down) is removed
+   * instead of failing the whole initialization, mirroring the tolerant handling of the
+   * follower-sync path in {@code registerOptimizerWithoutPersist}.
+   */
+  void registerOptimizers(List<OptimizerInstance> optimizers) {
+    for (OptimizerInstance optimizer : optimizers) {
+      if (optimizingQueueByGroup.containsKey(optimizer.getGroupName())) {
```
xxubai: In master-slave mode, another AMS instance may create a resource group and register an optimizer between these queries. Since the local queue map does not yet contain the new group, this code could incorrectly delete a valid optimizer record from the shared database.

j1wonpark:
Thanks for the fix. The NPE is real and this stops it.

One thought on the test: it calls `registerOptimizers` directly, so it skips `loadOptimizingQueues` where the NPE actually happens. I tried restoring `optimizers.forEach(...)` while keeping the new method, and the bug comes back but the test still passes.

amoro-ams/src/main/java/org/apache/amoro/server/DefaultOptimizingService.java
czy006: Thanks for the feedback. I need to further verify whether there are any additional regressions in the high-availability setup.

j1wonpark (approved):
LGTM, +1 (non-binding). The new test covers the real startup path now.

zhoujinsong (approved):
LGTM.

PR #4324, opened 2026-08-17
[hotfix][Optimizer] Guard invalid fragment ratio and target quota

czy006:
## Brief change log

- Clamp per-process quota and target quota values to at least one so invalid quota values cannot corrupt scheduling weights or starve a table.
- Normalize a zero or negative `self-optimizing.fragment-ratio` to 1 in `OptimizingConfig.setFragmentRatio()`; the configured default remains 8 in `TableProperties`.
- Reuse `OptimizingConfig.maxFragmentSize()` in `CommonPartitionEvaluator` so the production planner no longer performs a raw division by an invalid ratio.
- Keep the persisted table property unchanged; normalization applies only to the runtime `OptimizingConfig`.

## How was this patch tested?

- [x] `./mvnw test -pl amoro-common -Dtest=TestOptimizingConfig` (2 tests passed)
- [x] `./mvnw test -pl amoro-format-iceberg -am -Dtest=TestCommonPartitionEvaluator` (1 test passed)
- [x] `./mvnw test -pl amoro-ams -am -Dtest="TestOptimizingQueue#testQuotaSchedulePolicy"` (1 test passed; 15 reactor modules succeeded)
- [x] `./mvnw validate` (33 reactor modules succeeded)
- [x] Backend-only change; screenshots are not applicable.

## Documentation

### DIFF
--- a/amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
+++ b/amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
@@ -212,7 +212,7 @@ public class OptimizingConfig {
   }
 
   public OptimizingConfig setFragmentRatio(int fragmentRatio) {
-    this.fragmentRatio = fragmentRatio;
+    this.fragmentRatio = Math.max(1, fragmentRatio);
     return this;
   }
 
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
@@ -622,9 +622,11 @@ public class OptimizingQueue extends PersistentBase {
 
     private int getQuotaLimit() {
       double targetQuota = tableRuntime.getOptimizingConfig().getTargetQuota();
+      // A non-positive target quota must not starve the table to zero schedulable slots;
+      // clamp to 1 like getAvailableCore does for the group quota.
       return targetQuota > 1
           ? (int) targetQuota
-          : (int) Math.ceil(targetQuota * getAvailableCore());
+          : (int) Math.max(1, Math.ceil(targetQuota * getAvailableCore()));
     }
 
     @Override
--- a/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/table/DefaultTableRuntime.java
@@ -228,8 +228,12 @@ public class DefaultTableRuntime extends AbstractTableRuntime {
 
   public double calculateQuotaOccupy() {
     double targetQuota = getOptimizingConfig().getTargetQuota();
+    // Guard against a non-positive target quota (misconfigured property): a zero limit would
+    // turn the weight into Infinity/NaN and corrupt quota-based sorting.
     int targetQuotaLimit =
-        targetQuota > 1 ? (int) targetQuota : (int) Math.ceil(targetQuota * getThreadCount());
+        targetQuota > 1
+            ? (int) targetQuota
+            : (int) Math.max(1, Math.ceil(targetQuota * getThreadCount()));
     return (double) getQuotaTime() / AmoroServiceConstants.QUOTA_LOOK_BACK_TIME / targetQuotaLimit;
   }
 
--- a/amoro-common/src/test/java/org/apache/amoro/config/TestOptimizingConfig.java
+++ b/amoro-common/src/test/java/org/apache/amoro/config/TestOptimizingConfig.java
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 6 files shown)


amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
```diff
@@ -226,7 +226,9 @@ public OptimizingConfig setMinTargetSizeRatio(double minTargetSizeRatio) {
   }
 
   public long maxFragmentSize() {
-    return targetSize / fragmentRatio;
+    // A non-positive fragment ratio (misconfigured property or legacy persisted value) means
+    // "no fragmenting": fall back to the target size instead of dividing by zero in planners.
+    return fragmentRatio <= 0 ? targetSize : targetSize / fragmentRatio;
```
xxubai: the default `target-size` is 134217728 which is too large

czy006: Good catch. I updated the fallback to use the documented default fragment ratio of 8, so the default 128 MiB target size yields a 16 MiB fragment threshold. CommonPartitionEvaluator now calls OptimizingConfig.maxFragmentSize() instead of performing the raw division, and a planner-level regression test covers both zero and negative ratios.

amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
```diff
@@ -26,6 +26,8 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class OptimizingConfig {
 
+  public static final int DEFAULT_FRAGMENT_RATIO = 8;
```
xxubai: Why define the defaut value there

amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
```diff
@@ -226,7 +228,8 @@ public OptimizingConfig setMinTargetSizeRatio(double minTargetSizeRatio) {
   }
 
   public long maxFragmentSize() {
-    return targetSize / fragmentRatio;
+    int effectiveFragmentRatio = fragmentRatio > 0 ? fragmentRatio : DEFAULT_FRAGMENT_RATIO;
+    return targetSize / effectiveFragmentRatio;
```
xxubai:
```suggestion
    return targetSize / Math.max(1, fragmentRatio);
```

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
```diff
@@ -104,7 +105,8 @@ private TableProperties() {}
   public static final long SELF_OPTIMIZING_MAX_TASK_SIZE_DEFAULT = 134217728; // 128 MB
 
   public static final String SELF_OPTIMIZING_FRAGMENT_RATIO = "self-optimizing.fragment-ratio";
-  public static final int SELF_OPTIMIZING_FRAGMENT_RATIO_DEFAULT = 8;
+  public static final int SELF_OPTIMIZING_FRAGMENT_RATIO_DEFAULT =
```
xxubai:
```suggestion
  public static final int SELF_OPTIMIZING_FRAGMENT_RATIO_DEFAULT = 8;
```

amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
xxubai: ignore

amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
xxubai: I think we can also add a upper limitation for fragmentRatio. it should be less or equals than `targetSize`

xxubai: I believe the param would be validated in `setFragmentRatio` instead of getter method. WDYT

amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
czy006: Agreed. I removed the default constant from `OptimizingConfig`; the configured default remains owned by `TableProperties`.

amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
czy006: Thanks. Following the later setter-validation suggestion, the lower bound is now applied once in `setFragmentRatio()` instead of inside `maxFragmentSize()`.

amoro-format-iceberg/src/main/java/org/apache/amoro/table/TableProperties.java
czy006: Done. `SELF_OPTIMIZING_FRAGMENT_RATIO_DEFAULT` is restored to `8` in `TableProperties`.

amoro-common/src/main/java/org/apache/amoro/config/OptimizingConfig.java
czy006: Agreed and updated. `setFragmentRatio()` now normalizes zero and negative values to `1`, and `maxFragmentSize()` remains a simple derived calculation. The evaluator also reuses this method.

czy006: I have not added the upper bound in this hotfix. `fragmentRatio` is dimensionless while `targetSize` is a byte size, so this would introduce a new cross-field invariant. Also, `TableConfigurations` currently calls `setFragmentRatio()` before `setTargetSize()`, so the setter cannot validate against the final target size reliably. A very large positive ratio produces a zero fragment threshold but does not cause the division failure addressed here. I suggest handling an upper-bound policy separately, with defined semantics and validation after both values are available.

zhoujinsong (approved):
LGTM.

PR #4325, opened 2026-08-17
[hotfix][Optimizer] Preserve keeper visibility on lookup failure

czy006:
## Brief change log

Make optimizer touch-time visibility explicit and keep resource groups under observation when a container lookup fails transiently, so the keeper can retry on a later cycle.

## How was this patch tested?

- [x] Add keeper test coverage for touch-time visibility and lookup-failure requeueing.
- [x] Run `TestOptimizerGroupKeeper` locally with JDK 11 before creating this pull request.

## Documentation

- no
- not applicable


zhoujinsong (approved):
LGTM.

PR #3486, opened 2025-03-25
[AMORO-3485] Introduce scheduler module and external resource container

majin1102:
## Why are the changes needed?

Close #3485.

## Brief change log

Document: https://docs.google.com/document/d/1pEKneQXI_OIY7EU_XX0xwHeRFG0HHi37bb3SaNucI00/edit?tab=t.0

## How was this patch tested?

- [ ] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [ ] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (yes / no)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #3257, opened 2024-10-15
[AMORO-3066] Optimizing the efficiency for optimizing-processes rest api

klion26:
## Why are the changes needed?

Close #3066.

## Brief change log

- remove the ServerCatalog#tableExists(db, table) in TableController#getOptimizingProcesses because we can reuse the table load logic in the following logic
- push the offset and limit in mapper#selectOptimizingProcesses(MixedAndIcebergTableDescriptor#getOptimizingProcessesInfo) to db (this can limit the items retrieved from db)

## How was this patch tested?

- [x] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [x] Run test locally before making a pull request
WeChatWorkScreenshot_b2bc81ed-0dc4-41cc-92bb-f3fd48abde4f
WeChatWorkScreenshot_5f2c49bd-20a8-4d4e-9218-eba80d441ddd

## Documentation

- Does this pull request introduce a new feature? (yes / no)
- If yes, how is the feature documented? (not applicable / docs / JavaDocs / not documented)

PR #4331, opened 2026-08-17
[hotfix][Optimizer] Recognize CLOSED optimizing processes

czy006:
## Brief change log

Make `OptimizingProcess.isClosed()` recognize the `CLOSED` status set by `close()`, in addition to `KILLED`. This restores the guard against accepting late results for a closed optimizing process.

## How was this patch tested?

- [x] Extend process-close queue coverage to assert the closed predicate.
- [x] Run `TestOptimizingQueue#testProcessCloseKeepsLastOptimizedSnapshotId` locally with JDK 11 before creating this pull request.

## Documentation

- no
- not applicable

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
@@ -775,7 +775,10 @@ public class OptimizingQueue extends PersistentBase {
 
     @Override
     public boolean isClosed() {
-      return status == ProcessStatus.KILLED;
+      // close() sets CLOSED (KILLED is reserved for kill flows); checking only KILLED made
+      // this predicate permanently false, so the acceptResult guard against late results on a
+      // closed process could never fire.
+      return status == ProcessStatus.CLOSED || status == ProcessStatus.KILLED;
     }
 
     @Override
--- a/amoro-ams/src/test/java/org/apache/amoro/server/optimizing/TestOptimizingQueue.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/optimizing/TestOptimizingQueue.java
@@ -858,6 +858,8 @@ public class TestOptimizingQueue extends AMSTableTestBase {
     // Close process without success (simulates group change / forced termination)
     process.close(false);
 
+    Assert.assertTrue(process.isClosed());
+
     // lastOptimizedSnapshotId and lastOptimizedChangeSnapshotId should NOT be updated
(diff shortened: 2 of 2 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingQueue.java
```diff
@@ -775,7 +775,10 @@ private void resetTask(TaskRuntime<RewriteStageTask> taskRuntime) {
 
     @Override
     public boolean isClosed() {
```
xxubai, line 777: I'm wondering why we need `isClosed()` if we already have `getStatus()`. If it represents something other than the process status, we should either rename it to make its semantics clearer or remove it and use `getStatus()` instead.

czy006: Thanks for the question. `isClosed()` is not a shortcut for `getStatus() == ProcessStatus.CLOSED` — it is the predicate "this process was terminated out-of-band and must reject late task results", and it intentionally spans two statuses:

- `CLOSED` — written by `close(false)` (e.g. the table is released from the queue / the optimizer group changes) and by the partial-commit completion path in `commit()`;
- `KILLED` — the terminal state this class already treats as "process is gone" in `poll()` (`status != KILLED && status != FAILED`), and the recovery constructor restores the status from the persisted `TableProcessMeta`, so it can carry whatever the process framework persisted.

That is why simply removing it and comparing `getStatus()` at the call site does not express the intent:

- comparing `getStatus() == ProcessStatus.CLOSED` would silently lose the `KILLED` arm;
- comparing `getStatus() == CLOSED || getStatus() == KILLED` just re-creates this predicate inline, without a name.

For context, this is also exactly how the bug was introduced: #3257 migrated the old `OptimizingProcess.Status` check correctly (`status == ProcessStatus.CLOSED`), then #3486 changed the *readers* (`poll()`, the recovery check, `isClosed()`) from `CLOSED` to `KILLED` while leaving `close()` writing `CLOSED`. Since nothing in this class assigns `KILLED`, `isClosed()` has been permanently false and the guard in `acceptResult()` dead ever since. Keeping one named predicate for "results are no longer accepted" — instead of scattered raw status comparisons — is what prevents this kind of writer/reader drift from recurring.

On the naming — fair point that `isClosed()` covering `KILLED` can read as a `CLOSED` shortcut at first glance. I'd like to keep this hotfix minimal (just restoring the dead guard) and treat any rename (e.g. `isTerminated()`) as a follow-up if you think it is worth it, since it would touch the public `OptimizingProcess` interface.

xxubai (approved):

PR #4332, opened 2026-08-18
[Hotfix][AMS] Stabilize optimizer scale-down floor test

j1wonpark:

`TestOptimizerScaleKeeper.testScaleDownRespectsFloor` intermittently failed with `expected: <1> but was: <2>`.

The test created a DRA group with an unmet `min-parallelism` floor and then manually registered an optimizer. Group creation queues an immediate scale-keeper round, and floor recovery is intentionally not delayed by the configured scaling cadence. Depending on thread scheduling, the background keeper could register one optimizer before the test registered another.

This is a test setup race rather than a failure in the production scale-down logic.

## Brief change log

- Create the test DRA group with `min-parallelism=0`.
- Raise the floor to `1` after the test optimizer is registered.
- Assert that the live scale keeper does not trigger an unexpected scale-out during setup.

## How was this patch tested?

  - Reproduced the original failure deterministically: `expected: <1> but was: <2>`.
  - `TestOptimizerScaleKeeper`: 20 tests passed.
  - `testScaleDownRespectsFloor`: passed 10 consecutive runs.

  Not applicable because this is a test-only change.

  - `./mvnw -o -pl amoro-ams -DskipTests validate`
  - Spotless and Checkstyle passed.

## Documentation

### DIFF
--- a/amoro-ams/src/test/java/org/apache/amoro/server/TestOptimizerScaleKeeper.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/TestOptimizerScaleKeeper.java
@@ -458,15 +458,25 @@ public class TestOptimizerScaleKeeper extends AMSTableTestBase {
             .anyMatch(r -> optimizer.getResourceId().equals(r.getResourceId())));
   }
 
-  /** The min-parallelism floor keeps the last instance no matter how long it idles. */
+  /**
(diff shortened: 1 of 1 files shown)


j1wonpark:
Hi @czy006  Could you please take a look at this DRA test hotfix? Thanks!

czy006 (approved):

Issue #4268, opened 2026-07-10
Feature: HDFS commits for optimizing/maintenance as table owner (Hadoop proxy impersonation)


myloginid:
### Summary

On secured Hadoop clusters, AMS and optimizers typically run as a **dedicated service principal** (e.g. `amoro`). Iceberg/Hive warehouse paths are often owned by the **table creator** (e.g. `lily`). Optimizing **tasks** can succeed while the **AMS commit** fails with HDFS `AccessControlException` on `data/` or `metadata/` because all catalog file IO runs as the service user, not the table owner.

Configuring `hadoop.proxyuser.amoro.*` on the cluster is **necessary but not sufficient**: nothing in the AMS optimizing commit path uses Hadoop proxy impersonation today.

### Environment (example)

- Amoro 0.9.x (AMS + external optimizer)
- Kerberos, Hive Metastore + Iceberg tables on HDFS
- Ranger HDFS/Hive policies for the service user
- Tables under e.g. `/warehouse/tablespace/external/hive/<db>/<table>/` with owner = end user, mode `drwxr-xr-x`

### Observed behavior

1. Optimizer finishes MINOR (or similar) work; UI shows tasks **1/1 success**.
2. AMS commit fails; optimizing process status **FAILED**.
3. AMS logs show `OptimizingCommitException` / Iceberg commit failure, root cause similar to:

   `Permission denied: user=amoro, access=WRITE, inode=".../mor_events/metadata":lily:hive:drwxr-xr-x`

Ranger grants for `amoro` on `/*` and cluster `hadoop.proxyuser.amoro.{hosts,groups,users}=*` did **not** change this, because commits still execute as `amoro`.

Workarounds that work operationally but do not scale well:

- Per-table or per-DB HDFS ACLs for the service user on `data/` and `metadata/`
- Warehouse-wide default ACLs
- Treating the service user as a broad HDFS/Ranger “power user”

These are painful in multi-tenant warehouses where table ownership follows the creating user (as with Hive).

### Expected behavior

When a catalog property (or table property) enables **owner impersonation**, AMS and optimizers should perform HDFS reads/writes for optimizing commits (and related maintenance) **as the table owner** (from HMS/Iceberg metadata), using the catalog’s Kerberos principal as the **proxy user**, consistent with Hive/Impala-style deployments.

Cluster admins would configure:

```xml
<property><name>hadoop.proxyuser.amoro.hosts</name><value>*</value></property>
[...]

PR #4333, opened 2026-08-22
[AMORO-4268][AMS] Support HDFS owner impersonation for optimizing commits

johntomcat7408-cmyk:

AMS currently constructs optimizing commits with tables whose Hadoop FileIO runs as the catalog service identity. In secured, table-owner-controlled warehouses, optimizer tasks can succeed while the AMS commit fails to write metadata or clean up files.

This is the first scoped part of #4268. It covers AMS-side optimizing commits only; optimizer task writes and other maintenance operations are unchanged.

## Brief change log

- Add opt-in `hdfs.impersonation.enabled` configuration, disabled by default. An explicit table value overrides the direct catalog value and the catalog `table.` default.
- Scope owner impersonation to table loading for AMS optimizing commits so the resulting FileIO retains the owner identity for commit-time reads, writes, deletes, and executor-thread cleanup.
- Resolve the current owner from Hive Metastore for Hive-backed Iceberg, Mixed-Iceberg, and Mixed-Hive tables, and from refreshed Iceberg metadata for other Iceberg, Mixed-Iceberg, and internal tables.
- Require SIMPLE or KERBEROS authentication, HadoopFileIO, and a non-empty owner. Proxy and configuration failures are propagated without falling back to the catalog service user.
- Keep ordinary table loads unchanged and document the Hadoop proxy-user and authorization requirements.

## How was this patch tested?

- [x] Added positive and negative tests for configuration precedence, owner resolution, caller and executor identities, concrete FileIO entry points, unsupported configurations, missing owners, and no-fallback behavior.
- [ ] Screenshots are not applicable because this change has no UI surface.
- [x] Ran focused tests and validation locally.

Commands and results:

- `.\mvnw.cmd test -pl amoro-format-iceberg -am -Dtest=TestAuthenticatedHadoopFileIO,TestTableOwnerResolver -Dsurefire.failIfNoSpecifiedTests=false` — 17 tests passed.
- `.\mvnw.cmd test -pl amoro-format-mixed/amoro-mixed-hive -am -Dtest=TestMixedHiveTablesImpersonation -Dsurefire.failIfNoSpecifiedTests=false` — 2 tests passed.
- `TestInternalIcebergHandlerImpersonation` — 1 test passed in the AMS-focused run.
- `.\mvnw.cmd validate -pl amoro-format-iceberg,amoro-format-mixed/amoro-mixed-hive,amoro-ams -am` — all 15 reactor modules passed Spotless and Checkstyle validation.
- `git diff HEAD^ HEAD --check` — passed.

The AMS-focused run could not initialize `TestOptimizingQueue`'s Hive metastore on Windows because `HADOOP_HOME`/`winutils.exe` is unavailable. A real Kerberos/MiniDFS/HMS environment and the full test suite were not run locally.

## Compatibility and risks

- The feature is opt-in and disabled by default.
- There are no Thrift, REST, database schema, or dependency changes.
- When enabled, an invalid owner, unsupported authentication/FileIO, or Hadoop proxy authorization failure intentionally fails the optimizing commit instead of retrying as the service user.
- The conventional `hdfs` owner is not special-cased; cluster proxy-user policy remains authoritative.

## Provenance and contribution boundaries

The implementation is original to this PR. It builds on the existing `TableMetaStore#doAsImpersonating` and authenticated FileIO patterns already present in upstream commit `96a4d781aac60c4080b3f4a41322848e2308649a`. No code was copied from open or unmerged pull requests or third-party sources.

## Documentation

- If yes, how is the feature documented? `docs/admin-guides/managing-catalogs.md`


sivakumar-mahalingam (commented):
Could we avoid unconditional icebergTable.refresh() when HDFS impersonation is not enabled?

buildCommit() now loads the table inside withOptimizingCommitImpersonation(), so the TableOwnerResolver.resolve() appears to refresh the Iceberg table on each optimizing commit before checking hdfs.impersonation.enabled

It will add extra metadata I/O/failure point even when impersonation is disabled, which is the default behavior. Could we move the refresh after the hdfs.impersonation.enabled check so it will only run when impersonation is enabled?

PR #4335, opened 2026-08-24
[AMORO] Fix SLF4J 2.x logging not working due to incompatible log4j-slf4j-impl binding

felixhzhu:
What
Replace log4j-slf4j-impl with log4j-slf4j2-impl across all modules.

Why
PR #4046 upgraded slf4j-api from 1.7.x to 2.0.17, but did not update the Log4j2 binding accordingly. log4j-slf4j-impl uses the StaticLoggerBinder mechanism which is only compatible with SLF4J 1.7.x. SLF4J 2.x uses ServiceLoader to discover providers, and since log4j-slf4j-impl does not register as a SLF4JServiceProvider, SLF4J falls back to NOP (no-operation) logger.

This results in zero log output in amoro-optimizer (and potentially other modules depending on classpath order).

Fix
log4j-slf4j2-impl is the correct Log4j2 → SLF4J 2.x bridge that implements SLF4JServiceProvider.

How to verify
Build amoro-optimizer-standalone
Run optimizer — previously showed No SLF4J providers were found, now shows normal Log4j2 output


zhoujinsong:
fixed by #4336

PR #4336, opened 2026-08-25
[Hotfix] Change log4j impl dependency to `log4j-slf4j2-impl`

zhoujinsong:

PR #4046 upgraded `slf4j-api` from `1.7.x` to `2.0.17`, but did not update the Log4j2 binding accordingly. `log4j-slf4j-impl` uses the StaticLoggerBinder mechanism, which is only compatible with SLF4J 1.7.x. SLF4J 2.x uses ServiceLoader to discover providers, and since `log4j-slf4j-impl` does not register as a `SLF4JServiceProvider`, SLF4J falls back to a NOP (no-operation) logger.

This PR changes the impl to `log4j-slf4j2-impl` to match the new `slf4j-api`.

## Brief change log

- As title

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/pom.xml
+++ b/amoro-ams/pom.xml
@@ -267,8 +267,13 @@
         </dependency>
 
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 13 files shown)


czy006 (approved):

zhoujinsong:
Thanks for the review! @czy006

PR #4338, opened 2026-08-26
[AMORO-4337] Upgrade Iceberg dependency to 1.8.1

### LINKED ISSUE #4337: [Improvement]: Upgrade Iceberg dependency to 1.8.1
### Search before asking

Amoro currently depends on Iceberg 1.7.2. Iceberg has since released the 1.8, 1.9, 1.10 and 1.11 lines, and each of those lines removes engine modules that mixed-format connectors build on:

| Iceberg | Engine modules removed |
|---|---|
| 1.8.x | none |
| 1.9.0 | `iceberg-spark-3.3` |
| 1.10.0 | `iceberg-flink-1.18` |
| 1.11.0 | `iceberg-flink-1.19` (and the build baseline moves to JDK 17) |

This makes **1.8.1 the only upgrade in this range that costs no engine support**: mixed Spark 3.3/3.4/3.5 and mixed Flink 1.18/1.19/1.20 all remain buildable.

Staying on 1.7.2 also means the gap to the current Iceberg line keeps widening, and any later upgrade has to absorb both the API changes and the engine drops in a single change. Taking the free hop first separates those two concerns.

### How should we improve?

Bump `iceberg.version` from 1.7.2 to 1.8.1.

### Subtasks

- [ ] Bump `iceberg.version` to 1.8.1
- [ ] Migrate `RowDataFileScanTaskReader` to `FlinkPlannedAvroReader`
- [ ] Verify full build and test suites, including the `spark-3.3` and `spark-3.4` profiles
- [ ] Update documentation referencing the Iceberg version

### Code of Conduct

j1wonpark:

Close #4337.

Amoro depends on Iceberg 1.7.2. Among the Iceberg lines released since, 1.8.x is the only one that removes no engine module: `iceberg-spark-3.3` is dropped in 1.9.0, `iceberg-flink-1.18` in 1.10.0, and `iceberg-flink-1.19` in 1.11.0. Upgrading to 1.8.1 therefore keeps the full mixed-format engine matrix (Spark 3.3/3.4/3.5, Flink 1.18/1.19/1.20) intact while closing part of the gap to current Iceberg.

Doing this hop on its own also keeps the API migration separate from the engine-support decisions that later upgrades will require.

## Brief change log

- Bump `iceberg.version` from 1.7.2 to 1.8.1.
- Migrate `RowDataFileScanTaskReader` in `amoro-mixed-flink-common-iceberg-bridge` from `FlinkAvroReader`, removed in Iceberg 1.8.0, to `FlinkPlannedAvroReader`. The class is a copy of the Iceberg class of the same name, so this mirrors the exact change Iceberg 1.8 made to its own copy (including `path().toString()` → `location()`); the only remaining deviations are the pre-existing mixed-format ones (`AdaptHiveFlinkParquetReaders`, shaded guava).

- Fix the position-delete read schema in `IcebergRewriteExecutorTest`: it projected delete files with the virtual metadata columns `_file`/`_pos` instead of the `file_path`/`pos` columns the files actually store. Iceberg 1.7 tolerated this by synthesizing row positions and ignoring the missing `_file`, so the assertions passed by accident; Iceberg 1.8 rejects projecting a required field that is absent from the file.

- Regenerate the dependency manifests (`dev/deps`) via `./dev/dependencies.sh --replace`: Iceberg artifacts move to 1.8.1 and the transitive `httpclient5`/`httpcore5` move to 5.4.1/5.3.1.

No other production source changes are required.

## How was this patch tested?

- Full `clean test-compile` passes against Iceberg 1.8.1 across all modules, including the `spark-3.3` and `spark-3.4` profiles.
- `amoro-mixed-flink-common` + `amoro-mixed-flink-common-iceberg-bridge`: 314 tests pass (Kafka testcontainers included).
- `amoro-mixed-hive`: 188 tests pass.
- `amoro-format-iceberg`: `IcebergRewriteExecutorTest` passes across all format/partition combinations after the schema fix.
- `spotless:check` and `checkstyle:check` pass.
- Verified the bridge `RowDataFileScanTaskReader` against the Iceberg 1.8.1 source: after this change, the only remaining differences from the upstream class are the pre-existing mixed-format adaptations (`AdaptHiveFlinkParquetReaders`, shaded guava imports).
- Built the shaded `amoro-mixed-flink-runtime-1.20` jar and confirmed it bundles the Iceberg 1.8.1 classes.


zhoujinsong (approved):
LGTM.

zhoujinsong:
Besides, perhaps we could start a discussion about upgrading Iceberg to a newer version. While this might require us to drop support for some mixed-format engine versions, maintaining compatibility with newer versions remains necessary.

Issue #3670, opened 2025-07-16
[Feature]: Support AMS horizontal performance expansion capabilities


wardlican:
### Description

Currently, AMS uses a master-slave mode. If a large number of Iceberg tables are managed, the master-slave mode will have performance bottlenecks, and a solution that can support horizontal performance expansion is needed.

### Use case/motivation

Multiple catalogs need to be registered on amoro. Under each catalog, there are many Iceberg tables that need to be managed automatically. If only a single-node AMS is working, the single node will consume a lot of resources. It is necessary to support a horizontally scalable performance expansion solution. By expanding the capacity of AMS nodes, each node can be responsible for processing a part of the Iceberg table. In this way, the resources of each node can be controlled and horizontal performance expansion can be achieved.

### Describe the solution

Change the master-slave mode to the master-slave mode. Each AMS node is only responsible for processing a part of the Iceberg table. It supports dynamic expansion and contraction of the number of AMS nodes and automatic migration of the Iceberg table. The optimizer will pull tasks from each AMS node for processing. The AMS master node is responsible for evenly distributing the Iceberg table to all AMS nodes in the cluster and managing the life cycle of the AMS nodes in the cluster.

### Subtasks

-[]#1 Change tableService to SPI mode to support expansion of other implementation methods.
-[]#2 Implement the master-slave mode in the extended tableService. The master node is responsible for synchronizing table information from the catalog.
-[]#3 To extend the implementation of ZK, in addition to being responsible for the election of the master, it is also necessary to implement functions such as the registration of ordinary AMS nodes
-[]#4 Modify the optimizer to support obtaining tasks from each AMS node for processing.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://www.apache.org/foundation/policies/conduct)

wardlican:
Please help with the code review. @zhoujinsong @baiyangtx @xxubai

PR #4339, opened 2026-08-26
[Refactor] Unify the main process of amoro service container

zhoujinsong:

#3670 introduced a new master-slave mode for Amoro. However, many implementations adopted separate code paths for this mode. While this reduced intrusion into the legacy code, it is not conducive to long-term maintenance. This PR refactors the main process of `AmoroServiceContainer` to unify the handling logic.

## Brief change log

- As title

## How was this patch tested?

## Documentation


amoro-ams/src/main/java/org/apache/amoro/server/AmoroServiceContainer.java
```diff
@@ -316,11 +348,6 @@ public void disposeOptimizingService() {
       LOG.info("Stopping optimizing server[serving:{}] ...", optimizingServiceServer.isServing());
       optimizingServiceServer.stop();
     }
-    if (amsAssignService != null) {
```
xxubai, line 319: `dispose()` still calls `disposeOptimizingService()` directly from the shutdown hook. Will it cause the `ams-assign-scheduler` running while the other services are being disposed?

zhoujinsong: Yes, we may stop the leader service first, if necessary, when disposing.
I have made some changes, PTAL.

xxubai (approved):
The previous disposal issue is fixed by stopping leader services before disposing the remaining services.

czy006 (approved):

PR #4342, opened 2026-08-28
[Hotfix] Remove slf4j classes from kyuubi-hive-jdbc-shaded

zhoujinsong:

`kyuubi-hive-jdbc-shaded` shade the slf4j 1.x classes which are conflicts with slf4j 2.x in ams.
This PR removed them from the jar to avoid conflicts.

## Brief change log

- As title

## How was this patch tested?

## Documentation

### DIFF
--- a/amoro-ams/pom.xml
+++ b/amoro-ams/pom.xml
@@ -608,6 +608,33 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Strip shaded SLF4J 1.x classes from kyuubi-hive-jdbc-shaded to prevent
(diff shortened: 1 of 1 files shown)


j1wonpark:
Reproduced with the 0.9.0 RC5 binary and on our k8s deployment. AMS and optimizer logs are empty (`StaticLoggerBinder` NOP fallback). Putting `slf4j-api` first on the classpath brings them back, so the fix looks right. RC5 doesn't include this, though.

j1wonpark (approved):
LGTM. Verified in our deployment: the stripped jar no longer contains org/slf4j/** classes and AMS logging works fine without binding conflicts.

PR #4347, opened 2026-08-31
[AMORO-1523][AMS] Auto-install discovered plugins when no plugin configs exist

### LINKED ISSUE #1523: [Bug]: When I start AMS on IntelliJ IDEA, I cannot start the Local Optimizer.
### What happened?

When I start AMS on IntelliJ IDEA, I cannot start the Local Optimizer.

### Affects Versions

0.5.0

AMS

### How to reproduce

To start the Local Optimizer in IntelliJ IDEA after starting AMS:

### Relevant log output

_No response_

### Anything else

When you start AMS directly from IntelliJ IDEA, the CLASSPATH environment variable is empty. As a result, the localOptimize.sh startup script cannot access the CLASSPATH and the Local Optimizer cannot be launched.

### Code of Conduct

zhang-arvin:

Fixes #1523: When starting AMS in IntelliJ IDEA, the Local Optimizer cannot be started because the `execute-engines.yaml` plugin configuration file is missing from the default IDE setup.

### Root Cause

`AbstractPluginManager.initialize()` discovers plugins via ServiceLoader but then only installs those with matching YAML configurations. When started from an IDE without the `plugins/` config directory, no configurations are loaded, so discovered plugins like `LocalExecutionEngine` are never installed.

### Fix

When `loadPluginConfigurations()` returns an empty list but `foundAvailablePlugins()` has discovered plugins, auto-create default configurations for all discovered plugins. This ensures the Local Optimizer works out-of-the-box in IDE development environments.

## How was this patch tested?

- Verified the fix compiles
- The `emptyConfig()` method already exists in `PluginConfiguration` for this purpose

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/manager/AbstractPluginManager.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/manager/AbstractPluginManager.java
@@ -85,6 +85,19 @@ public abstract class AbstractPluginManager<T extends ActivePlugin> implements P
     foundAvailablePlugins();
 
     List<PluginConfiguration> pluginConfigs = loadPluginConfigurations();
+    // If no plugin configs were loaded from file but plugins were found via ServiceLoader,
+    // create default configurations so they can be installed. This handles the case where
+    // AMS is started from an IDE without the full plugins directory structure.
+    if (pluginConfigs.isEmpty() && !foundedPlugins.isEmpty()) {
+      pluginConfigs =
+          foundedPlugins.keySet().stream()
+              .map(PluginConfiguration::emptyConfig)
+              .collect(Collectors.toList());
+      LOG.info(
+          "No {} plugin configurations found, auto-installing discovered plugins: {}",
+          pluginCategory(),
+          foundedPlugins.keySet());
+    }
     pluginConfigs.stream()
         .sorted(Comparator.comparing(PluginConfiguration::getPriority))
         .forEach(


czy006:
Hi @zhang-arvin, thanks a lot for picking this up, and for the solid investigation in #1523 — your root-cause analysis (plugins are discovered via ServiceLoader but never installed because the `conf/plugins/` directory is absent in IDE setups) is spot on. Making the local optimizer work out-of-the-box from an IDE is definitely worth fixing. 👍

Unfortunately I found one blocking issue with the current trigger condition that we'll need to address before merging. It's quite contained to fix though, and I've sketched a direction below — happy to help review the next iteration!

### Blocking: default binary deployments would fail to start

The shipped `conf/plugins/metric-reporters.yaml` on master has its plugin list fully commented out, so `loadPluginConfigurations()` returns an **empty list in every default deployment** — not just in IDE runs. And since `amoro-metrics-prometheus` is bundled in the distribution and registers `PrometheusMetricsReporter` via SPI, `foundedPlugins` is non-empty, so the new branch auto-installs it with an empty config. `PrometheusMetricsReporter.open()` then throws `IllegalArgumentException("Lack required property: port")` ([PrometheusMetricsReporter.java#L39](https://github.com/apache/amoro/blob/master/amoro-metrics/amoro-metrics-prometheus/src/main/java/org/apache/amoro/metrics/promethues/PrometheusMetricsReporter.java#L39)), and since nothing on the path catches it (`install()` → `initialize()` → `MetricManager.getInstance()` → `startRestServices()`), AMS would fail to boot on a fresh install.

### The root of it: "empty" conflates two different states

`pluginConfigs.isEmpty()` currently means either:

1. the config **file is missing** (the IDE case this PR targets), or
2. the file **exists but its list is empty** — the deliberate way to disable a whole category, and the shipped default for `metric-reporters`.

For case 2, this change flips the semantics from "install nothing" to "install everything discovered". For example, an operator who disables the Iceberg REST catalog by deleting its entry from `rest-extensions.yaml` (rather than setting `enabled: false`) would get it silently re-enabled with default properties — that's surprising behavior we should avoid.

### Suggested direction

- Distinguish the two states, e.g. have `loadPluginConfigurations()` return `Optional<List<PluginConfiguration>>`: file missing → auto-install defaults (your new branch); file present but list empty → keep today's "install nothing" behavior. This resolves both points above in one go.
- In the auto-install path, wrap each `install()` in a try/catch that logs a WARN and skips the failing plugin, so a single plugin that requires mandatory properties (prometheus-exporter does today) can't abort AMS startup.
- Tests would be very welcome — `TestAbstractPluginManager` already has everything you need (injectable configs + `TestPluginImplA/B` registered in the test SPI):
  - empty configs + discovered plugins → all installed;
  - a discovered plugin whose `open()` throws → initialize logs a warning and continues.

Thanks again for working on this — the diagnosis was great, and with the trigger condition narrowed this will be a nice improvement for IDE development! 🙌

PR #3924, opened 2025-11-11
[Feature] Introduce a new framework that supports scheduling for Formats and Processes

LiangDai-Mars:
## Why are the changes needed?
 
To introduce a new framework that supports scheduling for a wider range of Formats and Processes

## Brief change log
 
- Introduce a new top-level service, ProcessService, responsible for managing the lifecycle of Processes.
- Introduce a new plugin type **ActionCoordinator** to extend how TableProcess is created.
- Introduce a new plugin type **ExecutionEngine** to extend how external **Process** is executed.

## How was this patch tested?

- [X] Add some test cases that check the changes thoroughly including negative and positive cases if possible

- [ ] Add screenshots for manual tests if appropriate

- [X] Run test locally before making a pull request

## Documentation

- Does this pull request introduce a new feature? (yes)
- If yes, how is the feature documented? (not documented)

PR #4349, opened 2026-08-31
[AMORO-4348] Fix infinite retry loop for failed table processes

### LINKED ISSUE #4348: [Bug] Failed table process is resubmitted in an infinite retry loop
### What happened?

When a table process (e.g. `EXPIRE-SNAPSHOTS`) fails once, AMS resubmits it in a tight infinite loop with no backoff. Each iteration logs the following trio, flooding the log — we observed ~270k occurrences and 1.3GB of logs within 70 minutes, with AMS constantly burning ~1.7 CPU cores:

```
WARN ... Failed to transit status for processEvent: SUBMIT_REQUESTED with initial status: FAILED, process: ... have been marked as terminal status: FAILED.
WARN ... Failed to transit status for processEvent: COMPLETE_FAILED with initial status: FAILED, ...
WARN ... Failed to transit status for processEvent: RETRY_REQUESTED with initial status: FAILED, ...
```

### Root cause

Since #4116, `ProcessService` constructs `DefaultTableProcessStore` passing `processMeta.getRetryNumber()` (the *current* retry count) into the constructor's `maxRetryTime` parameter (originally `PROCESS_MAX_RETRY_NUMBER` in #3924). For a new process this makes `maxRetryTime = 0`, so on the first failure `isTerminal(FAILED)` evaluates `retryNumber(0) >= maxRetryTime(0)` = true: the process becomes terminal immediately, every subsequent transition is rejected, and `retryNumber` never increases.

Meanwhile the retry branch in `executeOrTraceProcess` checks `retryNumber < PROCESS_MAX_RETRY_NUMBER` (0 < 3, always true), ignores the rejected `RETRY_REQUESTED` transition, and unconditionally resubmits — an infinite loop.

### Affects Versions

master

_No response_

### Code of Conduct

j1wonpark:

Fix #4348.

Since #4116, `ProcessService` constructs `DefaultTableProcessStore` passing the *current* retry count into the constructor's `maxRetryTime` parameter. A new process therefore gets `maxRetryTime = 0` and becomes terminal on its first failure, every subsequent transition is rejected, and `retryNumber` never increases — while the retry branch in `executeOrTraceProcess` keeps evaluating `retryNumber < PROCESS_MAX_RETRY_NUMBER` as true and resubmits the process forever, with no backoff.

## Brief change log

- Construct `DefaultTableProcessStore` with `PROCESS_MAX_RETRY_NUMBER` as `maxRetryTime`, in both the creation and the recovery path (restores the pre-#4116 semantics of #3924)
- Only resubmit a failed process when its `RETRY_REQUESTED` transition is actually accepted, so a rejected transition can never loop
- Add a regression test verifying a persistently failing process is retried exactly `PROCESS_MAX_RETRY_NUMBER` times and then dropped

## How was this patch tested?

`TestDefaultProcessService#testFailedProcessRetryIsBounded` reproduces the infinite loop before the fix (the process is never dropped and the wait times out) and passes after: the failing process is submitted `1 + PROCESS_MAX_RETRY_NUMBER` times in total and then untracked. Existing process service tests all pass.

## Documentation

### DIFF
--- a/amoro-ams/src/test/java/org/apache/amoro/server/TestDefaultProcessService.java
+++ b/amoro-ams/src/test/java/org/apache/amoro/server/TestDefaultProcessService.java
@@ -30,6 +30,7 @@ import org.apache.amoro.process.TableProcess;
 import org.apache.amoro.process.TableProcessStore;
 import org.apache.amoro.server.persistence.PersistentBase;
 import org.apache.amoro.server.persistence.mapper.TableProcessMapper;
+import org.apache.amoro.server.process.ActionCoordinatorScheduler;
 import org.apache.amoro.server.process.MockActionCoordinator;
 import org.apache.amoro.server.process.MockExecuteEngine;
 import org.apache.amoro.server.process.ProcessService;
@@ -47,6 +48,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


amoro-ams/src/test/java/org/apache/amoro/server/TestDefaultProcessService.java
```diff
@@ -392,6 +395,66 @@ public void testRecoverProcessFailSafe() {
...
+    @Override
+    public String submitTableProcess(org.apache.amoro.process.TableProcess tableProcess) {
+      String identifier = "failing-" + submitAttempts.incrementAndGet();
+      failedIdentifiers.add(identifier);
+      return identifier;
+    }
+
+    @Override
+    public ProcessStatus getStatus(String processIdentifier) {
+      if (failedIdentifiers.contains(processIdentifier)) {
```
xxubai: nit: Please make the fake engine fail directly from `submitTableProcess()`. 

Returning `FAILED` immediately takes the fast-terminal path and produces redundant `COMPLETE_FAILED` transition errors.


```suggestion
      throw new IllegalStateException("Submission failure");
```

zhoujinsong (approved):
LGTM.

amoro-ams/src/test/java/org/apache/amoro/server/TestDefaultProcessService.java
j1wonpark: Done.

xxubai (approved):

PR #4351, opened 2026-09-01
[AMORO-4350][catalog] Fix Hive view handling in table management

### LINKED ISSUE #4350: [Bug]: Opening a Hive view from the Tables page causes a NullPointerException
### What happened?

When an external Hive catalog contains a Hive view, the view may be returned by Hive Metastore table listing APIs and exposed on the Amoro Tables page as a manageable table.

### Actual behavior

Opening the view invokes the normal table detail endpoint. CommonUnifiedCatalog attempts to load it through the configured format catalogs, and MixedHiveCatalog handles the Hive view as a physical table. The view has no storage location, so MixedHiveCatalog.getMixedTableMeta calls `endsWith` on a null location and throws a NullPointerException. The detail request returns HTTP 500 and the page cannot be opened.

### Expected behavior

Hive virtual and materialized views should not be exposed as manageable Amoro tables. Loading a view through a physical table path should return a clear not-found or unsupported result instead of throwing a NullPointerException. Regular physical tables should remain unaffected.

### Affects Versions

master

Iceberg, Mixed-Hive

_No response_

### How to reproduce

1. Configure an external Hive catalog in Amoro.
2. Create a physical table and a Hive view in the same database.
3. Open the database from the Tables page.
4. Observe that the Hive view is listed as a manageable table.
5. Open the view entry.
6. Observe that the detail request returns HTTP 500 with a NullPointerException from MixedHiveCatalog.getMixedTableMeta.

### Relevant log output

```shell
java.lang.NullPointerException: null
    at org.apache.amoro.hive.catalog.MixedHiveCatalog.getMixedTableMeta(MixedHiveCatalog.java:195)
    at org.apache.amoro.hive.catalog.MixedHiveCatalog.loadTable(MixedHiveCatalog.java:347)
    at org.apache.amoro.formats.mixed.MixedCatalog.loadTable(MixedCatalog.java:72)
    at org.apache.amoro.CommonUnifiedCatalog.loadTable(CommonUnifiedCatalog.java:134)
    at org.apache.amoro.server.dashboard.ServerTableDescriptor.getTableDetail(ServerTableDescriptor.java:75)
    at org.apache.amoro.server.dashboard.controller.TableController.getTableDetail(TableController.java:151)
```

### Anything else

Hive views should remain available to query engines that support them, but they should not be treated as Iceberg, Mixed Hive, or native Hive physical tables by Amoro table management.

### Code of Conduct

csurong:

Hive views may be returned by Hive Metastore listing APIs and exposed as manageable tables. Opening one can make `MixedHiveCatalog` dereference a null storage location and return HTTP 500.

This PR excludes Hive virtual and materialized views from Amoro table management paths and rejects them before physical table metadata is accessed.

Closes #4350.

## Brief change log

- Add lightweight HMS table metadata access and centralized Hive view classification.
- Filter views from native Hive, Iceberg HiveCatalog, and Mixed Hive listings.
- Reject views and invalid physical metadata during table loading.
- Add unit and integration coverage for Hive, Iceberg, and Mixed Hive paths.

## How was this patch tested?

- [x] Add test cases that check the changes thoroughly including negative and positive cases if possible.
- [ ] Add screenshots for manual tests if appropriate.
- [x] Run tests locally before making a pull request.

The targeted unit and integration tests passed. AMS compilation, Hadoop 2/Hive 2.3 compilation, Spotless, Checkstyle, and diff checks also passed.

## Documentation


amoro-common/src/main/java/org/apache/amoro/hive/HiveTableTypeUtil.java
```diff
@@ -0,0 +1,112 @@
...
+  public static Set<String> listViewNames(
+      HMSClient client, String database, List<String> candidateTableNames) throws TException {
+    if (candidateTableNames == null || candidateTableNames.isEmpty()) {
+      return Collections.emptySet();
+    }
+
+    List<TableMeta> views;
+    try {
+      views = client.getTableMeta(database, "*", viewTypes());
+    } catch (UnsupportedOperationException e) {
```
czy006: **[Optional] The `UnsupportedOperationException` fallback is unreachable — the real failure mode is uncaught**

The fallback in `HiveTableTypeUtil.listViewNames` only triggers via the interface's default method, but `HMSClientImpl` (the only production implementation) always overrides `getTableMeta` — so it's dead code outside Mockito tests. The scenario that can actually happen is an old metastore server without the `get_table_meta` endpoint (it only entered the HMS thrift IDL in Hive 2.0.0): the client then throws `TApplicationException`, which is a `TException`, isn't caught here, and fails the whole listing call — a new hard dependency this PR introduces.

Non-blocking since Amoro supports HMS 2.x/3.x, but consider either dropping the fallback and letting `TException` propagate, or catching `TApplicationException` and falling back to `getTableObjectsByName` (which exists on Hive 1.x servers).

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/IcebergCatalog.java
```diff
@@ -89,11 +118,32 @@ public void dropDatabase(String database) {
...
-            icebergCatalog.listTables(Namespace.of(database)).stream()
-                .map(TableIdentifier::name)
-                .collect(Collectors.toList()));
+    List<String> tableNames =
+        metaStore.doAs(
+            () ->
+                icebergCatalog.listTables(Namespace.of(database)).stream()
+                    .map(TableIdentifier::name)
+                    .collect(Collectors.toList()));
+    if (hiveClientPool == null || tableNames.isEmpty()) {
```
czy006, line 132: **[Optional] Skip the extra `get_table_meta` RPC when iceberg already excludes views**

For hive-type iceberg catalogs, `IcebergCatalog.listTables` now always adds one `get_table_meta` call. With `list-all-tables=false` (the default; nothing in Amoro overrides it), iceberg's `HiveCatalog` already filters by `table_type=iceberg`, so views are absent and the new filter is a guaranteed no-op — pure overhead in the default config.

Consider gating on the property:

```java
boolean listAllTables =
    Boolean.parseBoolean(properties.getOrDefault("list-all-tables", "false"));
if (hiveClientPool == null || tableNames.isEmpty() || !listAllTables) {
  return tableNames;
}
```

amoro-common/src/main/java/org/apache/amoro/hive/HiveTableTypeUtil.java
csurong: Done, thanks for the suggestion

amoro-format-iceberg/src/main/java/org/apache/amoro/formats/iceberg/IcebergCatalog.java
csurong: Done, thanks for the suggestion

czy006 (approved):

PR #4352, opened 2026-09-02
[Hotfix][AMS] Fix optimizer registration failure under database HA mode

wangxianghu:
  When HA is enabled with ha.type=database, AmsUtil.getAMSThriftAddress still takes the ZooKeeper branch because it only checks ha.enabled. This makes it build a zookeeper://... address from ha.zookeeper-address, which is not configured in database-HA deployments. The bad URI is injected into the optimizer as ams.optimizer.uri, so the optimizer cannot reach AMS and fails to register.                                     
                                                                                                                                                                                                                    
  The fix narrows the ZooKeeper branch to ha.enabled && ha.type == zk. For database HA (and non-HA), it now returns the direct thrift://<expose-host>:<port> address, which is the same host/port the AMS node registers under its lease, so the optimizer registers correctly.

## Brief change log
  - AmsUtil.getAMSThriftAddress: only return the ZooKeeper address when ha.enabled is true and ha.type is zk; otherwise fall back to the direct thrift host:port address. This fixes optimizer registration under database-based HA. 

## How was this patch tested?
    - Added TestAmsUtil#getAMSThriftAddress covering: HA disabled, ZK-HA (returns zookeeper://...), and database-HA (falls back to thrift://...).                                                                   

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/dashboard/utils/AmsUtil.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/dashboard/utils/AmsUtil.java
@@ -124,7 +124,11 @@ public class AmsUtil {
   }
 
   public static String getAMSThriftAddress(Configurations conf, String serviceName) {
-    if (conf.getBoolean(AmoroManagementConf.HA_ENABLE)) {
+    boolean zkHaEnabled =
+        conf.getBoolean(AmoroManagementConf.HA_ENABLE)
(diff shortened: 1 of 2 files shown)


wangxianghu:
hi @zhoujinsong  could you please take a look when free ?   this is a simple bug fix

xxubai (approved):
LGTM. The HA branch is now correctly limited to ZooKeeper.

PR #4356, opened 2026-09-04
[ams] Fix optimizing status metrics never updated and reset on restart

czy006:

The table optimizing status metrics (table_optimizing_status_in_* and table_optimizing_status_*_duration_mills) have been broken since the TableRuntime storage refactor in AMORO-3747:

1. TableOptimizingMetrics#statusChanged is never invoked, so the status gauges stay at their initial values (idle) for the whole process lifetime and non-idle durations are always zero.
2. The runtime is not seeded with the persisted status code and update time when constructed, so durations restart from zero after an AMS restart.
3. Same-status status-code writes (e.g. setPendingInput on a non-idle table) refresh status_code_update_time, restarting the persisted duration clock used by the dashboard.

## Brief change log

- seed TableOptimizingMetrics from the persisted status code and status_code_update_time when constructing DefaultTableRuntime
- notify the metrics from the status-change handler after the update is committed to the store
- skip status-code updates that don't change the value, so no-op writes no longer refresh status_code_update_time or fire redundant status-change notifications

## How was this patch tested?

## Documentation


zhoujinsong (approved):
LGTM.

PR #4362, opened 2026-09-05
[AMORO-4361][ams] Fix Iceberg REST catalog table creation collision to return 409 Conflict

### LINKED ISSUE #4361: [Bug]: Iceberg REST catalog table creation collision returns HTTP 500 instead of 409 Conflict
### What happened?

**What happened?**
When an Iceberg REST client (PyIceberg, Spark, Trino, Flink) attempts to create a table that already exists or commit a staged table creation against an existing table, the AMS REST Catalog endpoint returns HTTP 500 InternalServerError instead of HTTP 409 Conflict.

**Root Cause**
1. In `RestCatalogService.IcebergRestErrorCode.exceptionToCode()`, only `org.apache.iceberg.exceptions.AlreadyExistsException` is mapped to `Conflict` (409). However, internal table managers throw `org.apache.amoro.exception.AlreadyExistsException`. Because the Amoro class does not extend the Iceberg exception, the check evaluates to `false` and falls through to HTTP 500.
2. In `createTable()` and `commitCreateTable()`, table existence is not verified upfront before creating physical metadata files on object storage (S3/HDFS), causing redundant file creation followed by rollback deletions on duplicate requests.

**Expected Behavior**
- Duplicate table creation requests should return HTTP 409 Conflict per the Iceberg REST OpenAPI specification.
- AMS should check table existence upfront to avoid redundant storage I/O and rollbacks.

### Affects Versions

master,0,.9.0

Iceberg

AMS, Spark, Trino

### How to reproduce

1. Start AMS with an Iceberg REST catalog enabled (e.g., lakehouse).

2. Send an initial table creation request to create table `test_db.test_table`:
curl -X POST http://localhost:1630/api/iceberg/rest/v1/catalogs/lakehouse/namespaces/test_db/tables \
  -H "Content-Type: application/json" \
  -d '{"name":"test_table","schema":{"type":"struct","schema-id":0,"fields":[{"id":1,"name":"id","required":true,"type":"int"}]}}'

Result: HTTP 200 OK (table created successfully).

3. Send the exact same table creation request a second time to attempt duplicate creation:
curl -X POST http://localhost:1630/api/iceberg/rest/v1/catalogs/lakehouse/namespaces/test_db/tables \
  -H "Content-Type: application/json" \
  -d '{"name":"test_table","schema":{"type":"struct","schema-id":0,"fields":[{"id":1,"name":"id","required":true,"type":"int"}]}}'

Expected: HTTP 409 Conflict with message indicating table already exists (compliant with Iceberg REST OpenAPI specification).

Actual: HTTP 500 InternalServerError with org.apache.amoro.exception.AlreadyExistsException.

### Relevant log output

```shell
2026-09-05 09:12:00,123 ERROR [JettyServerThreadPool-31] [org.apache.amoro.server.RestCatalogService] - Internal Server Error
org.apache.amoro.exception.AlreadyExistsException: Table test_table already exists
	at org.apache.amoro.server.table.DefaultTableManager.createTable(DefaultTableManager.java:142)
	at org.apache.amoro.server.RestCatalogService.createTable(RestCatalogService.java:338)
```

### Anything else

_No response_

### Code of Conduct

aakashofficial-k01:

Under concurrent table creation race conditions, `DefaultTableManager` throws `org.apache.amoro.exception.AlreadyExistsException`. Because `IcebergRestErrorCode.exceptionToCode()` previously only checked for `org.apache.iceberg.exceptions.AlreadyExistsException`, Amoro's exception fell through to HTTP 500 (InternalServerError), violating the Iceberg REST OpenAPI specification (which requires 409 Conflict) and breaking client idempotency for compute engines. Additionally, Amoro's `ForbiddenException` was unmapped and returned HTTP 500 instead of HTTP 403.

Close #4361.

## Brief change log

- Mapped `org.apache.amoro.exception.AlreadyExistsException` to `Conflict` (HTTP 409) in `RestCatalogService.IcebergRestErrorCode.exceptionToCode()`.
- Mapped `org.apache.amoro.exception.ForbiddenException` to `Forbidden` (HTTP 403) in `RestCatalogService.IcebergRestErrorCode.exceptionToCode()`.
- Added unit test assertions covering both Amoro and Iceberg exceptions in `TestRestCatalogService.java`.

## How was this patch tested?

- [x] Ran unit test verification:
  `./mvnw test -pl amoro-ams -Dtest=TestRestCatalogService`: BUILD SUCCESS (1/1 passed, 0 failures)
- [x] Ran style and checkstyle validation:
  `./mvnw checkstyle:check spotless:check -pl amoro-ams`: BUILD SUCCESS (0 violations)

## Documentation

### DIFF
--- a/amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
@@ -321,6 +321,7 @@ public class RestCatalogService extends PersistentBase implements RestExtension 
           CreateTableRequest request = bodyAsClass(ctx, CreateTableRequest.class);
           request.validate();
           String tableName = request.name();
+          checkAlreadyExists(!catalog.tableExists(database, tableName), "Table", tableName);
           TableFormat format =
               TablePropertyUtil.isBaseStore(request.properties(), TableFormat.MIXED_ICEBERG)
                   ? TableFormat.MIXED_ICEBERG
@@ -329,7 +330,6 @@ public class RestCatalogService extends PersistentBase implements RestExtension 
           try (InternalTableCreator creator =
               catalog.newTableCreator(database, tableName, format, request)) {
             if (request.stageCreate()) {
-              checkAlreadyExists(!catalog.tableExists(database, tableName), "Table", tableName);
               return LoadTableResponse.builder().withTableMetadata(creator.stage()).build();
             }
 
@@ -400,6 +400,7 @@ public class RestCatalogService extends PersistentBase implements RestExtension 
 
   private LoadTableResponse commitCreateTable(
       InternalCatalog catalog, String database, String tableName, UpdateTableRequest request) {
+    checkAlreadyExists(!catalog.tableExists(database, tableName), "Table", tableName);
     request.requirements().forEach(requirement -> requirement.validate((TableMetadata) null));
 
     Optional<Integer> formatVersion =
@@ -636,8 +637,12 @@ public class RestCatalogService extends PersistentBase implements RestExtension 
         return NotFound;
       } else if (e instanceof NoSuchNamespaceException) {
         return NotFound;
-      } else if (e instanceof AlreadyExistsException) {
+      } else if (e instanceof AlreadyExistsException
+          || e instanceof org.apache.amoro.exception.AlreadyExistsException
+          || e instanceof org.apache.amoro.exception.BlockerConflictException) {
         return Conflict;
+      } else if (e instanceof org.apache.amoro.exception.ForbiddenException) {
+        return Forbidden;
       }
       return InternalServerError;
     }
--- a/amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
+++ b/amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
@@ -321,7 +321,6 @@ public class RestCatalogService extends PersistentBase implements RestExtension 
           CreateTableRequest request = bodyAsClass(ctx, CreateTableRequest.class);
           request.validate();
           String tableName = request.name();
-          checkAlreadyExists(!catalog.tableExists(database, tableName), "Table", tableName);
           TableFormat format =
               TablePropertyUtil.isBaseStore(request.properties(), TableFormat.MIXED_ICEBERG)
                   ? TableFormat.MIXED_ICEBERG
@@ -330,6 +329,7 @@ public class RestCatalogService extends PersistentBase implements RestExtension 
           try (InternalTableCreator creator =
               catalog.newTableCreator(database, tableName, format, request)) {
             if (request.stageCreate()) {
+              checkAlreadyExists(!catalog.tableExists(database, tableName), "Table", tableName);
               return LoadTableResponse.builder().withTableMetadata(creator.stage()).build();
             }
 
@@ -400,7 +400,6 @@ public class RestCatalogService extends PersistentBase implements RestExtension 
 
   private LoadTableResponse commitCreateTable(
       InternalCatalog catalog, String database, String tableName, UpdateTableRequest request) {
-    checkAlreadyExists(!catalog.tableExists(database, tableName), "Table", tableName);
     request.requirements().forEach(requirement -> requirement.validate((TableMetadata) null));
 
     Optional<Integer> formatVersion =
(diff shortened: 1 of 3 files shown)


amoro-ams/src/main/java/org/apache/amoro/server/RestCatalogService.java
```diff
@@ -321,6 +321,7 @@ public void createTable(Context ctx) {
           CreateTableRequest request = bodyAsClass(ctx, CreateTableRequest.class);
           request.validate();
           String tableName = request.name();
+          checkAlreadyExists(!catalog.tableExists(database, tableName), "Table", tableName);
```
j1wonpark: (non-blocking) `InternalCatalogImpl.newTableCreator` already checks `tableExists` and throws Iceberg's `AlreadyExistsException` before any metadata is written, for both the stage and non-stage branches. This check and the one in `commitCreateTable` are redundant — could we drop them and adjust the description?

amoro-ams/src/test/java/org/apache/amoro/server/TestInternalIcebergCatalogService.java
```diff
@@ -437,6 +438,26 @@ public void testServerCatalogLoadTable() throws IOException {
...
+
+    @Test
+    public void testCommitCreateTableAlreadyExists(@TempDir Path tempDir) {
+      nsCatalog.createTable(identifier, schema);
+      Path tablePath = tempDir.resolve("staged-table-conflict");
+      Transaction transaction =
+          nsCatalog
+              .buildTable(identifier, schema)
+              .withLocation(tablePath.toUri().toString())
+              .createTransaction();
```
j1wonpark: (blocking) `createTransaction()` sends the stage-create request eagerly, so the 409 surfaces on this line, outside the `assertThrows`. More importantly, both new integration tests pass with `RestCatalogService.java` reverted to master (this one after reordering it to reach the commit), so they don't exercise the change. I'd remove both; the `TestRestCatalogService` assertions are the real regression test (they fail on master with `expected: <Conflict> but was: <InternalServerError>`).

j1wonpark (changes requested):
Thanks for the PR. I checked it out and ran the tests locally.

The 500 in #4361 doesn't reproduce on master for a plain duplicate create: `newTableCreator` checks `tableExists` first and throws Iceberg's `AlreadyExistsException`, which is already mapped to 409. The Amoro `AlreadyExistsException` from `DefaultTableManager.createTable` is only reachable if the table appears between those two checks. Removing the `newTableCreator` check and re-running `testCreateTableAlreadyExists` on master reproduces the issue exactly (`ServiceFailureException: Server error: AlreadyExistsException`), so that check is what produces the 409 today. If you saw the 500 on a real deployment, could you share the version and request path? (The issue's `DefaultTableManager.java:142` is inside `block()` on both master and 0.9.x.)

The mapping change itself is still worth having: it closes that race window and is consistent with `ObjectNotExistsException`, which is already mapped to 404. `ForbiddenException` → 403 is a real improvement too — `DashboardServer.preHandleRequest` can throw it on the REST path and it currently becomes a 500. `BlockerConflictException` is harmless but unreachable from `RestCatalogService`.

Suggestion: trim the PR to the `exceptionToCode` change plus the `TestRestCatalogService` assertions, drop the two upfront checks and the two integration tests, and update the description. With that I'm happy to approve. CI still needs a maintainer to approve the workflow once the branch is updated.

aakashofficial-k01:
Thanks for the thorough review and explanation @j1wonpark !

I've pushed a follow-up commit and updated the PR accordingly:
1. Reverted the upfront checks in `createTable` and `commitCreateTable`.
2. Removed the integration test changes in `TestInternalIcebergCatalogService`.
3. Dropped `BlockerConflictException` and retained mapping for Amoro's `AlreadyExistsException` (409) and `ForbiddenException` (403).
4. Updated unit test assertions in `TestRestCatalogService`.
5. Updated the PR description.

Please take another look when you get a chance!

j1wonpark (approved):
LGTM, thanks for the quick turnaround — CI is green.

czy006:
+1 from me as well. One minor note for the record (non-blocking): the ForbiddenException → 403 mapping is currently unreachable from the REST catalog routes — URL_WHITE_LIST in DashboardServer includes ICEBERG_REST_API_PREFIX + "/*", so preHandleRequest returns early there, and the only other throw site (DefaultOptimizingService) is on the Thrift path. It's harmless defensive mapping and semantically correct per the spec, so fine to keep as-is.

PR #4376, opened 2026-09-14
[AMORO-4375][Iceberg] Prevent automatic changes to unsupported V3 tables

### LINKED ISSUE #4375: [Bug]: Automatic maintenance can invalidate Iceberg V3 table metadata
## What happened?

Amoro’s Iceberg 1.8.1 writer can commit changes to native V3 tables created by Iceberg 1.9+, but drops the required `next-row-id` field. Subsequent loads with Iceberg 1.9+ fail:

```text
IllegalArgumentException: Cannot parse missing long: next-row-id
```

Expected: reject automatic changes to unsupported format versions and report the reason without modifying metadata.

## Affects Versions

Confirmed on `master` at `bdd5f5816`, using Iceberg 1.8.1.

Iceberg

AMS, Optimizer, Spark

## How to reproduce

1. Create a `format-version=3` table using Spark with Iceberg 1.11.0 in a catalog managed by Amoro.
2. Insert multiple batches and let Amoro commit a self-optimizing compaction.
3. The new metadata loses `next-row-id`, and Spark can no longer load the table.

Observed after one full compaction, reducing 12 data files to one. Also reproduced locally with an Iceberg 1.9.2 append followed by a 1.8.1 tag commit.

## Relevant log output

```text
IllegalArgumentException: Cannot parse missing long: next-row-id
```

## Anything else

Iceberg 1.8.1 interprets the absent `row-lineage` flag in newer V3 metadata as disabled and omits `next-row-id` when writing.

The existing V3 check only disables Parquet row-group merging; row-based rewriting still commits. Other metadata-changing maintenance operations are also affected.

Protection should cover execution and commit boundaries, including retries and recovered jobs. This fix would prevent unsafe changes, not add V3 support or repair affected tables.

j1wonpark:

Amoro’s bundled Iceberg 1.8.1 writer can omit `next-row-id` when committing metadata for V3 tables created by newer writers, making them unreadable by Iceberg 1.9+.

The existing V3 check only disables Parquet row-group merging and does not prevent other automatic changes.

Close #4375.

## Brief change log

- Check format compatibility before optimizing evaluation, planning, and recovery.
- Guard maintenance and metadata commits, including retries after an external V3 upgrade. Orphan cleanup checks compatibility before deleting files.
- Preserve recovery during catalog lookup failures, disabled maintenance behavior, and read-only summary collection.
- Record unsupported-version closure reasons and document the V1/V2 support boundary.

This patch prevents unsafe automatic changes; it does not add V3 support or repair affected metadata.

## How was this patch tested?

- [x] Add positive and negative tests.
- [x] Run tests locally.

All 50 new test cases, including parameterized cases, passed, covering metadata preservation, supported versions, concurrent upgrades, retries, and recovery.

Related regression tests: 218 passed, 62 existing skips. After narrowing the orphan-cleanup wrapper scope, 63 targeted cases passed.

Full Maven validation passed before that final adjustment; targeted reactor style checks passed afterward. Separate Iceberg 1.9.2 verification confirmed reads and writes remain possible after maintenance is rejected.

The full repository test suite was not run. Screenshots are not applicable.

## Documentation

- Does this pull request introduce a new feature? **No.**
- Updated `docs/formats/iceberg.md` with the compatibility boundary.


amoro-ams/src/main/java/org/apache/amoro/server/optimizing/OptimizingProcess.java
```diff
@@ -30,6 +30,9 @@ public interface OptimizingProcess {
 
   void close(boolean needCommit);
 
+  /** Close without committing and persist the reason in the process history. */
+  void close(String reason);
```
zhoujinsong: Suggest exposing the full overload as the sole abstract method, with defaults for the two convenience forms:

```java
void close(boolean needCommit, String reason);

default void close(boolean needCommit) { close(needCommit, null); }
default void close(String reason) { close(false, reason); }
```

zhoujinsong (commented):
Thanks for the work.

I left a minor suggestion.

zhoujinsong (approved):
LGTM.

PR #4378, opened 2026-09-15
[AMORO-4377][AMS] Decouple the maintenance Iceberg version from mixed-format connectors and move AMS to Iceberg 1.10.2

### LINKED ISSUE #4377: [Improvement]: Decouple the Iceberg version used by AMS/optimizer from mixed-format connectors
### Search before asking

- [x] I have searched in the issues and found no similar issues.

AMS, the optimizers and the mixed-format connectors all share the single root `iceberg.version` (1.8.1). Every Iceberg bump therefore has to be paid twice: once for the maintenance path (planner, rewrite executors, snapshot expiration, orphan clean) and once for every mixed-format engine connector. The connector side is where the cost actually lands, because each Iceberg release retires engine bindings (1.9 drops Spark 3.3, 1.10 drops Flink 1.18, 1.11 requires JDK 17), so upgrading Iceberg for maintenance forces the same engine drops on users of the mixed-format connectors.

The maintenance service and the mixed-format connectors run in different processes (AMS / optimizer JVMs vs. Spark, Flink and Trino jobs), so they do not need the same Iceberg version.

### How should we improve?

Let the maintenance path (AMS, optimizers, dist) select its own Iceberg version while `amoro-format-iceberg`, `amoro-mixed-hive` and the engine connectors stay on the current baseline.

The attached PR keeps every module in place and only

- adds a `maintenance-iceberg.version` property that AMS, optimizer and dist use for their Iceberg (and Parquet) dependency management,
- switches the two Parquet reader calls removed in Iceberg 1.9 to the equivalent form that exists in both 1.8.1 and 1.10.x,
- runs the shared mixed-format I/O tests once more from the AMS module, so they are executed against the Iceberg version AMS actually ships.

One decision is needed alongside the PR: the newer Parquet required by Iceberg 1.10 does not run on Hadoop 2, so AMS would require Hadoop 3 while the connectors keep their Hadoop 2 build. Where new V3-only maintenance code should live, given that the shared modules stay on the lower API, is a follow-up.

### Code of Conduct

j1wonpark:

Close #4377.

This follows up on the Iceberg upgrade discussion: let the Iceberg version used by the self-optimizing / maintenance path evolve independently from the mixed-format connectors, so that a maintenance-side upgrade no longer forces engine drops on connector users.

It keeps every module and class where it is. AMS, the optimizers and `dist` select Iceberg 1.10.2; `amoro-format-iceberg`, `amoro-mixed-hive` and all engine connectors stay on 1.8.1.

AMS and the optimizers require Hadoop 3 from this change on (see below). Whether the mixed-format connectors keep their Hadoop 2 build is being discussed on the dev list: https://lists.apache.org/thread/v7zhhzzgn14wk1vwfqygy1mswwwgb9fy. This PR does not touch the connector side either way.

## Brief change log

- Root `pom.xml`: new `maintenance-iceberg.version` (1.10.2) and `maintenance-parquet.version` (1.16.0) properties. `iceberg.version` stays 1.8.1.
- Root `pom.xml`: the connector-side Parquet version is now explicit as well. `parquet-avro.version` becomes `parquet.version` (still 1.15.2) and the root manages all six Parquet artifacts with it, so the two sides read as `iceberg.version` / `parquet.version` vs. `maintenance-iceberg.version` / `maintenance-parquet.version`. On the connector side `parquet-jackson` keeps its separate 1.13.1 pin; on the maintenance side it is managed at 1.16.0 like the rest of the family. Verified with `dependency:list` before and after: shipped artifacts (connector runtime jars, Trino) resolve exactly as before; the only change is that `amoro-format-paimon` and `amoro-mixed-spark-3-common` previously resolved a mix of 1.13.1 / 1.15.0 / 1.15.2 and now resolve 1.15.2 consistently.
- `amoro-ams`, `amoro-optimizer`, `dist`: set `iceberg.version` to the maintenance version and manage the Iceberg artifacts the root does not pin (`iceberg-api`, `iceberg-bundled-guava`, Spark binding and extensions) plus the Parquet family. The Parquet alignment is required: Iceberg 1.10.x needs a newer Parquet than the one resolved through 1.8.1, and this shows up as `NoSuchMethodError` at runtime, not at build time.
- Two Parquet reader calls that Iceberg removed in 1.9 (`ParquetValueReaders.StructReader(types, readers)` and `ParquetValueReader.setPageSource(pages, rowPosition)`) are switched to the form that exists in both 1.8.1 and 1.10.x, in `amoro-mixed-hive` and the Spark 3.4 / 3.5 readers that AMS loads for its local terminal. The shared modules are still compiled against 1.8.1; the same bytecode runs on both versions.
- New `TestMaintenanceIcebergCompatibility` in `amoro-ams` writes a Parquet file and reads it back through the mixed-Hive generic reader (`AdaptHiveParquet` / `AdaptHiveGenericParquetReaders`) and the Spark 3.5 reader (`SparkParquetReaders`) on the AMS classpath, i.e. against Iceberg 1.10.2. Without the reader change above both fail with the `NoSuchMethodError`; with it, both pass. It needs no metastore, so it does not interact with the shared test HMS of the other AMS tests.
- `TestIcebergV3OptimizingCommit`: one V2 fixture wrote three records to replace two. Iceberg 1.10 validates that a REPLACE commit does not add more records than it replaces (`Invalid REPLACE operation: 3 added records > 2 replaced records`), so the fixture now keeps the input row count. Amoro's rewrite outputs never add records, so no production code is affected.
- CI: the Spark 3.3 matrix entry and the Hadoop 2 job now build with `-pl '!amoro-ams,!dist'`, see the Hadoop 2 section below.
- The Hadoop 2 distribution is no longer produced: the hadoop2 binary release is dropped from `tools/releasing/create_binary_release.sh`, the `hadoop: v2` matrix entry and the `-hadoop2` image tags from `docker-images.yml`, and the Hadoop 2 dist dependency list from `dev/dependencies.sh` / `dev/deps`. The mixed-format connectors, their Hadoop 2 CI job and the optimizer images are not changed.
- `dev/deps`: dependency lists regenerated with `dev/dependencies.sh --replace`. Besides Iceberg 1.10.2 and Parquet 1.16.0, the maintenance classpath picks up `parquet-variant`, `jts-core`, and transitive bumps of aircompressor (0.27 -> 2.0.3), avro (1.12.0 -> 1.12.1) and httpclient5 / httpcore5 (5.4.1 / 5.3.1 -> 5.5 / 5.3.4). The Hadoop 2 list is removed with the Hadoop 2 dist build.
- `LICENSE-binary`: `jts-core` is a new transitive dependency of Parquet 1.16 (EDL 1.0). The other new or bumped artifacts are already covered.

## How was this patch tested?

- Same shared bytecode (`amoro-common`, `amoro-format-iceberg`, `amoro-mixed-hive`, compiled with JDK 11 against Iceberg 1.8.1) run with the existing I/O tests on both an Iceberg 1.8.1 and an Iceberg 1.10.2 classpath: 85 passed / 15 assumption-skipped / 0 failed on each side. Class hashes verified identical between the two runs.
- On the AMS classpath (Iceberg 1.10.2, JDK 11): `TestKeyedContinuousOptimizing` and `TestUnKeyedContinuousOptimizing` (12/12, native Iceberg, mixed Iceberg and mixed Hive), `TestSnapshotExpire*` and `TestOrphanFileClean*` (116 passed / 28 assumption-skipped / 0 failed).
- Local Spark 3.5 with the mixed, Iceberg and Paimon extensions registered together (the terminal's combination): create native Iceberg table, insert, aggregate.
- Packaged artifacts inspected with a small probe on JDK 11: `amoro-ams-dependency/lib`, Flink and Spark optimizer jars load Iceberg 1.10.2 (no duplicate `org/apache/iceberg` classes); mixed Flink 1.18 / 1.19 / 1.20 and Spark 3.5 runtime jars load relocated Iceberg 1.8.1.
- Reactor builds: default (Spark 3.5), `-Pspark-3.4`, `-Pspark-3.3` and `-Phadoop2` (both with `-pl '!amoro-ams,!dist'`).
- Optimizer classpath on Hadoop 2 (`-Phadoop2,spark-3.5 -pl amoro-optimizer/amoro-optimizer-common -am`, Hadoop 2.10.2 / Iceberg 1.10.2 / Parquet 1.16.0): reading a Parquet file through `HadoopFileIO`, both via `Parquet.read` (the rewrite executor's data reader) and via `ParquetFileReader.open` (the row-group merge path), fails with `NoSuchMethodError: FileSystem.openFile(Path)` at `HadoopInputFile.newStream`. The same test passes on the default Hadoop 3.4.3 classpath. Writing is not affected.

Not covered: end-to-end run with an external optimizer process, real terminal API / mixed catalog SQL, Trino, full dist tarball run.

## Hadoop 2

Parquet 1.16 calls `FileSystem.openFile(Path)`, which does not exist in Hadoop 2.10, so AMS and the optimizers cannot read Parquet on a Hadoop 2 classpath (see the test section). There is no workaround on our side short of keeping them on the old Parquet, which defeats the upgrade.

AMS and the optimizers therefore require Hadoop 3 from the next release, and this PR stops producing the Hadoop 2 distribution (binary tarball, docker image, dependency list). The Hadoop 2 build of the mixed-format connectors is kept as is; whether to keep it long term is a separate question and is being discussed on the dev list (link above). The Flink optimizer image still bundles `flink-shaded-hadoop-2-uber` 2.8.3 and needs to move to a Hadoop 3 bundle; that is a follow-up.

## Notes and limitations

1. **Spark 3.3.** `iceberg-spark-3.3` was last published for Iceberg 1.8.1, so AMS can no longer be built with `-Pspark-3.3`. The default AMS build is Spark 3.5 and is unaffected; the Spark 3.3 mixed-format connector is still built and published, just not in the same reactor as AMS.
2. **AMS and optimizers must share one Iceberg version.** Task descriptors cross the AMS / optimizer boundary through Java serialization, and `DataFile` serialization is not compatible between 1.8.1 and 1.9+ (`serialVersionUID`). This PR sets both from the same property.
3. **Compile boundary.** The shared modules are compiled against 1.8.1, so code that needs 1.9+ API (row lineage, `MetadataColumns.ROW_ID`, ...) cannot live in `amoro-format-iceberg` as it is today. New V3-specific maintenance code needs a location that compiles against the maintenance version. This PR does not decide that; it only shows that the version split itself does not require moving modules first.
4. Parquet 1.16.0 is the version Iceberg 1.10.2 declares. I have not tried to find the minimum Parquet version that works.

## Documentation

- README build section updated: Hadoop 2 and Spark 3.3 connector build commands.


zhoujinsong:
@j1wonpark Thanks for the work.

Regarding whether Hadoop 2 should be removed or only supported within the mixed format, I suggest initiating a discussion thread on the dev mailing list to raise awareness of this issue and gather more opinions.

sivakumar-mahalingam:
PR notes that the Hadoop 2 distribution still builds, but AMS cannot read Parquet with the upgraded dependency. Could we remove or disable the Hadoop 2 distribution build before merging, so users aren't provided with an artifact that cannot perform this operation?

j1wonpark:
@zhoujinsong Thanks, opened the thread: https://lists.apache.org/thread/v7zhhzzgn14wk1vwfqygy1mswwwgb9fy
Independent of the outcome, AMS and the optimizers cannot read Parquet on Hadoop 2 either way, so 6533fdd removes the Hadoop 2 distribution build; the connector side is untouched until the discussion settles.

j1wonpark:
@sivakumar-mahalingam Agreed, done in 6533fdd: the Hadoop 2 binary release, docker image and dist dependency list are removed.
The optimizer fails the same way on Hadoop 2.10.2, so the README now describes the Hadoop 2 build as connectors only.

sivakumar-mahalingam (approved):
The change looks good 👍

sivakumar-mahalingam:
Since the Hadoop 2 distribution is being removed, would it be useful to create a follow-up issue to clean up the remaining Hadoop 2 dependencies/configuration in AMS and optimizers, such as `flink-shaded-hadoop-2-uber`?

I can take this up if it is aligned with the direction of the project.

